PR #62, opened 2023-11-28
Updated code-example to work out-of-the-box

cwognum:
## Changelogs

- Update the code-example to work out-of-the-box.
- Update the documentation of the main `load_dataset` and `load_benchmark` endpoints.

---

Changes that were suggested during user interviews.


polaris/hub/client.py
```diff
@@ -414,7 +429,11 @@ def upload_results(self, results: BenchmarkResults, access: AccessType = "privat
         return response
 
     def upload_dataset(
-        self, dataset: Dataset, access: AccessType = "private", timeout: TimeoutTypes = (10, 200)
+        self,
+        dataset: Dataset,
+        access: AccessType = "private",
+        timeout: TimeoutTypes = (10, 200),
```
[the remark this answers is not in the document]
cwognum: Didn't make this change myself, but it seems to come from `httpx` ([ref](https://www.python-httpx.org/advanced/#timeout-configuration)). Will document it!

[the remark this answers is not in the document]
cwognum: I believe @zhu0619 added this. Any preference on the above, Lu?

polaris/hub/client.py
```diff
@@ -432,7 +432,7 @@ def upload_dataset(
         self,
         dataset: Dataset,
         access: AccessType = "private",
-        timeout: TimeoutTypes = (10, 200),
+        timeout: TimeoutTypes = TimeoutTypes(connect=10, write=200),
```
hadim, line 435: I don't think it works that way - looks like you need to either use a `httpx.Timeout` object or stick with a tuple and simply explain what the tuple is about in the docstring.

hadim: I think a tuple is best (one less import for our users)

cwognum: Yeah, my bad! Should be good now!

hadim: Thanks! LGTM for me.

PR #66, opened 2023-12-07
remove nans

zhu0619:
This is to solve the issue when validate and compute `type_of_target` for datasets containing NaNs (for the  sparse datasets with multi targets).

## Changelogs
- Remove nans from dataset when check the target type.

### DIFF
--- a/polaris/benchmark/_base.py
+++ b/polaris/benchmark/_base.py
@@ -209,12 +209,16 @@ class BenchmarkSpecification(BaseArtifactModel):
 
         dataset = info.data.get("dataset")
         target_cols = info.data.get("target_cols")
+        splits = np.concatenate(info.data["split"])
+
         if dataset is None or target_cols is None:
             return v
 
         for target in target_cols:
             if target not in v:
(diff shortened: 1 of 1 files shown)


tests/test_benchmark.py
```diff
@@ -156,7 +156,7 @@ def test_benchmark_checksum(is_single_task, test_single_task_benchmark, test_mul
 
     # --- Test that the checksum is NOT the same ---
     def _check_for_failure(_kwargs):
-        with pytest.raises(ValidationError) as error:
+        with pytest.raises((ValidationError, TypeError)) as error:
```
cwognum, line 159: Why is this needed? Seems unrelated!

cwognum (approved):
Thanks for spotting this Lu! Just one question on the test case.

tests/test_benchmark.py
zhu0619: In the failure tests, a column of strings was tested. That string array fails at the `np.isnan` operator due to the type. 
Both TypeError and ValidationError are raised in that case.So the TypeError has to be added here.

PR #72, opened 2024-02-23
feature: polaris fsspec basic read only implementation

lmtroper:
## Changelogs

Integrated a read-only version of [Hadrien's POC](https://github.com/polaris-hub/polaris-fs/blob/main/polarisfs/client/fs.py) for the Polaris FSSpec. Some of the major changes include:
- Added a new endpoint in the client to be able to read the zarr file of a dataset.
- Added a new submodule in the hub folder, `polarisfs.py` which contains the FSSpec.
- For read-only implementation, the FSSpec currently supports the `ls` and `cat_file` operations.

#### More details of the FSSpec Implementation:
- The `ls` function makes a GET request to the Hub's api storage ls endpoint which returns an array of dictionaries, containing the files and directories within a given path in the Cloudflare R2 bucket.
- In the `ls` function, the prefix is required to be removed from each entry as the path provided to the Zarr store contains only the `{path}/...`, and not the `dataset/{owner}/{name}/{path}...`
- The `cat_file` makes a GET request to the Hub's api storage endpoint to obtain a signed URL for a given path and follows the redirect to obtain the contents of the file.
- The `cat_file` needed to have an specified timeout to avoid ending prematurely for larger sized zarr files

---

_Checklist:_

Linked to: https://github.com/polaris-hub/polaris-hub/pull/276

### DIFF
--- a/docs/api/hub.polarisfs.md
+++ b/docs/api/hub.polarisfs.md
@@ -1,4 +1,4 @@
-::: polaris.hub.polarisfs.PolarisFS
+::: polaris.hub.polarisfs.PolarisFileSystem
     options: 
         merge_init_into_class: true
         filters: ["!^_"]
--- a/docs/api/hub.polarisfs.md
+++ b/docs/api/hub.polarisfs.md
@@ -1,5 +1,5 @@
-::: polaris.hub.polarisfs.PolarisFSFileSystem
+::: polaris.hub.polarisfs.PolarisFS
     options: 
         merge_init_into_class: true
-        filters: ["!^_", "!create_authorization_url", "!fetch_token", "!request", "!token"]
+        filters: ["!^_"]
 ---
--- a/docs/api/hub.polarisfs.md
+++ b/docs/api/hub.polarisfs.md
@@ -1,0 +1,5 @@
+::: polaris.hub.polarisfs.PolarisFSFileSystem
+    options: 
+        merge_init_into_class: true
+        filters: ["!^_", "!create_authorization_url", "!fetch_token", "!request", "!token"]
+---
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -354,7 +354,9 @@ class PolarisHubClient(OAuth2Client):
             store = zarr.storage.FSStore(path, fs=polaris_fs)
             return zarr.open(store, mode="r")
         except TimeoutError as timeout_error:
-            raise PolarisHubError(f"Timeout error: {timeout_error}. Consider increasing the expiration_seconds parameter and try again.")
+            raise PolarisHubError(
+                f"Timeout error: {timeout_error}. Consider increasing the expiration_seconds parameter and try again."
+            )
         except Exception as e:
             raise PolarisHubError(f"Error opening Zarr store: {e}")
 
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -333,7 +333,7 @@ class PolarisHubClient(OAuth2Client):
 
         return Dataset(**response)
 
-    def read_zarr_file(self, owner: str, name: str, path: str, **kwargs):
+    def read_zarr_file(self, owner: Union[str, HubOwner], name: str, path: str) -> zarr.hierarchy.Group:
         """Read a Zarr file from a Polaris dataset
 
         Args:
@@ -342,12 +342,10 @@ class PolarisHubClient(OAuth2Client):
             path: Path to the Zarr file within the dataset.
 
         Returns:
-            zarr.hierarchy.Group: The Zarr object representing the dataset.
+            The Zarr object representing the dataset.
         """
         polaris_fs = PolarisFSFileSystem(
             polaris_client=self,
-            polarisfs_url=str(self.base_url),
-            default_expirations_seconds=10 * 60,
             dataset_owner=owner,
             dataset_name=name,
         )
@@ -355,6 +353,8 @@ class PolarisHubClient(OAuth2Client):
         try:
             store = zarr.storage.FSStore(path, fs=polaris_fs)
             return zarr.open(store, mode="r")
+        except TimeoutError as timeout_error:
+            raise PolarisHubError(f"Timeout error: {timeout_error}. Consider increasing the expiration_seconds parameter and try again.")
         except Exception as e:
             raise PolarisHubError(f"Error opening Zarr store: {e}")
 
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -28,7 +28,7 @@ from polaris.benchmark import (
 from polaris.dataset import Dataset
 from polaris.evaluate import BenchmarkResults
 from polaris.hub.settings import PolarisHubSettings
-from polaris.hub.polarisfs import PolarisFSFileSystem
+from polaris.hub.polarisfs import PolarisFS
 from polaris.utils import fs
 from polaris.utils.constants import DEFAULT_CACHE_DIR
 from polaris.utils.errors import PolarisHubError, PolarisUnauthorizedError
@@ -344,7 +344,7 @@ class PolarisHubClient(OAuth2Client):
         Returns:
             The Zarr object representing the dataset.
         """
-        polaris_fs = PolarisFSFileSystem(
+        polaris_fs = PolarisFS(
             polaris_client=self,
             dataset_owner=owner,
             dataset_name=name,
@@ -354,11 +354,9 @@ class PolarisHubClient(OAuth2Client):
             store = zarr.storage.FSStore(path, fs=polaris_fs)
             return zarr.open(store, mode="r")
         except TimeoutError as timeout_error:
-            raise PolarisHubError(
-                f"Timeout error: {timeout_error}. Consider increasing the expiration_seconds parameter and try again."
-            )
+            raise PolarisHubError(f"Timeout error.") from timeout_error
         except Exception as e:
-            raise PolarisHubError(f"Error opening Zarr store: {e}")
+            raise PolarisHubError(f"Error opening Zarr store") from e
 
     def list_benchmarks(self, limit: int = 100, offset: int = 0) -> list[str]:
         """List all available benchmarks on the Polaris Hub.
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -18,6 +18,7 @@ from authlib.oauth2.client import OAuth2Client as _OAuth2Client
 from httpx import HTTPStatusError
 from httpx._types import HeaderTypes, TimeoutTypes, URLTypes
 from loguru import logger
+import zarr
 
 from polaris.benchmark import (
     BenchmarkSpecification,
@@ -27,6 +28,7 @@ from polaris.benchmark import (
 from polaris.dataset import Dataset
 from polaris.evaluate import BenchmarkResults
 from polaris.hub.settings import PolarisHubSettings
+from polaris.hub.polarisfs import PolarisFSFileSystem
 from polaris.utils import fs
 from polaris.utils.constants import DEFAULT_CACHE_DIR
 from polaris.utils.errors import PolarisHubError, PolarisUnauthorizedError
@@ -101,6 +103,10 @@ class PolarisHubClient(OAuth2Client):
 
         self.code_verifier = generate_token(48)
 
+        self.polarisfs = PolarisFSFileSystem(
+            polarisfs_url=settings.api_url,
+            default_expirations_seconds=10 * 60)
+
         super().__init__(
             # OAuth2Client
             client_id=settings.client_id,
@@ -142,7 +148,6 @@ class PolarisHubClient(OAuth2Client):
             raise PolarisHubError(
                 f"The request to the Polaris Hub failed. See the error message below for more details:\n{response}"
             ) from error
-
         # Convert the response to json format if the response contains a 'text' body
         try:
             response = response.json()
@@ -302,6 +307,42 @@ class PolarisHubClient(OAuth2Client):
         )
         dataset_list = [bm["artifactId"] for bm in response["data"]]
         return dataset_list
+    
+    # TEMP
+    # Just using this to verify the path works from directly from the client
+    def list_objects_in_path(self, 
+        owner="polaris-test", # owner: Union[str, HubOwner],
+        name="__test", # name: str
+        path="zarr_test_archives/1GB_many_arrays.zarr/data.zarr/", 
+        **kwargs
+    ):
+
+        return self._base_request_to_hub(url=f"storage/dataset/{owner}/{name}/ls/{path}", 
+            method="GET")
+
+
+    def get_zarr_root_file(self, 
+    owner="polaris-test", 
+    name="__test", 
+    parts="zarr_test_archives/1GB_many_arrays.zarr/data.zarr/", 
+    **kwargs
+    ):
+        import traceback
+
+        path = f"{self.polarisfs.protocol}://storage/dataset/polaris-test/__test/ls/zarr_test_archives/1GB_many_arrays.zarr/data.zarr/"
+
+        try:
+            store = zarr.storage.FSStore(path, fs=self.polarisfs)
+            root = zarr.open(store, mode="a")
+            print(root.info)
+
+        except Exception as e:
+            print(f"Error opening Zarr store: {e}")
+            traceback.print_exc()
+
+        return None
+
+
 
     def get_dataset(self, owner: Union[str, HubOwner], name: str) -> Dataset:
         """Load a dataset from the Polaris Hub.
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -11,14 +11,14 @@ import certifi
 import fsspec
 import httpx
 import pandas as pd
+import zarr
 from authlib.common.security import generate_token
 from authlib.integrations.base_client.errors import InvalidTokenError, MissingTokenError
 from authlib.integrations.httpx_client import OAuth2Client, OAuthError
 from authlib.oauth2.client import OAuth2Client as _OAuth2Client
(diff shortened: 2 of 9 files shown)


polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # GET request to Polaris Hub for signed URL of file
+        response = self.polaris_client.get(cat_path, params={})
+
+        # This should be a 307 redirect with the signed URL
+        if response.status_code != 307:
+            try:
+                response.raise_for_status()
+            except httpx.HTTPStatusError as error:
+                raise PolarisHubError("Could not get signed URL from Polaris Hub.") from error
```
cwognum, line 78: An exception should always be thrown within this `if` statement! Now it can happen that the Hub returns a `200`, but we really needed that `307`.

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,82 @@
...
+                {"name": p["name"].replace(self.prefix, ""), "size": p["size"], "type": p["type"]}
+                for p in response.json()
+            ]
+            return detailed_entries
+
+    def cat_file(self, path: str, **kwargs):
+        cat_path = f"{self.base_path}/{path}"
+
+        # GET request to Polaris Hub for signed URL of file
+        response = self.polaris_client.get(cat_path, params={})
```
cwognum, line 71: Can we remove the params? 

```suggestion
        response = self.polaris_client.get(cat_path)
```

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,82 @@
...
+
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+from polaris.utils.errors import PolarisHubError
+
+
+class PolarisfsSettings(BaseSettings):
+    url: str = "http://localhost:3000/"
+
+    model_config = SettingsConfigDict(env_prefix="POLARIS_")
```
cwognum, line 14: I don't think we need this? Hadrien needed this as he was using a newly created `httpx` client in his request. We can rely on the Polaris client, which already has the relevant settings.

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,82 @@
...
+        # This should be a 307 redirect with the signed URL
+        if response.status_code != 307:
+            try:
+                response.raise_for_status()
+            except httpx.HTTPStatusError as error:
+                raise PolarisHubError("Could not get signed URL from Polaris Hub.") from error
+
+        storage_response = self.http_client.get(response.json()["url"], timeout=(10, 200))
+
+        return storage_response.content
```
cwognum, line 82: Related to my earlier comments, we should use the `start` and `end` parameter to limit which bytes we return!

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # This should be a 307 redirect with the signed URL
+        if response.status_code != 307:
+            try:
+                response.raise_for_status()
+            except httpx.HTTPStatusError as error:
+                raise PolarisHubError("Could not get signed URL from Polaris Hub.") from error
+
+        storage_response = self.http_client.get(response.json()["url"], timeout=(10, 200))
+
```
cwognum, line 81: Something might have gone wrong with the `storage_response`. We should add a `raise_for_status()`!

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,82 @@
...
+        self.base_path = f"/storage/dataset/{dataset_owner}/{dataset_name}"
+
+        # Prefix to remove from ls entries
+        self.prefix = f"dataset/{dataset_owner}/{dataset_name}/"
+
+    def ls(self, path: str, detail=False, **kwargs):
+        ls_path = f"{self.base_path}/ls/{path}"
+
+        # GET request to Polaris Hub to list objects in path
+        response = self.polaris_client.get(ls_path.rstrip("/"), params={})
```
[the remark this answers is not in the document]
cwognum: Can we also remove the `params={}`?

cwognum (commented):
Don't be discouraged with the many comments. I am super excited how this is coming together and you're doing great work! :rocket: 

Because it is all-in-all quite an intricate and complex system, I was quite nitpicky in my review to make sure it will be easier to maintain moving forward!

polaris/hub/client.py
```diff
@@ -332,6 +333,31 @@ def get_dataset(self, owner: Union[str, HubOwner], name: str) -> Dataset:
...
+            name: Name of the dataset.
+            path: Path to the Zarr file within the dataset.
+
+        Returns:
+            zarr.hierarchy.Group: The Zarr object representing the dataset.
+        """
+        polaris_fs = PolarisFSFileSystem(
+            polaris_client=self,
+            polarisfs_url=str(self.base_url), 
+            default_expirations_seconds=10 * 60,
```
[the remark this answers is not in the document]
lmtroper: @cwognum Should `default_expiration_seconds` be the parameter passed for timeout? 

e.g., `storage_response = self.polaris_client.get(response.json()["url"], auth=None, timeout=self.default_expiration_seconds)`

polaris/hub/client.py
```diff
@@ -332,6 +333,33 @@ def get_dataset(self, owner: Union[str, HubOwner], name: str) -> Dataset:
...
+        polaris_fs = PolarisFSFileSystem(
+            polaris_client=self,
+            dataset_owner=owner,
+            dataset_name=name,
+        )
+
+        try:
+            store = zarr.storage.FSStore(path, fs=polaris_fs)
+            return zarr.open(store, mode="r")
+        except TimeoutError as timeout_error:
```
cwognum, line 356: Looking at the HTTPX documentation, it appears `httpx` will throw a custom `TimeoutException`. See: https://www.python-httpx.org/exceptions/#the-exception-hierarchy

polaris/hub/client.py
```diff
@@ -332,6 +333,33 @@ def get_dataset(self, owner: Union[str, HubOwner], name: str) -> Dataset:
...
+            raise PolarisHubError(
+                f"Timeout error: {timeout_error}. Consider increasing the expiration_seconds parameter and try again."
+            )
```
cwognum, line 359: `expiration_seconds` is not a parameter the user can actually influence as of now! 

I would do the following: 
1. Add a `default_timeout` parameter to `PolarisFSFileSystem`
2. Keep the `timeout` parameter for `cat_file` and `ls`, but its default value is `None`. 
3. If `None` in `cat_file` or `ls`, the timeout will be set to the `default_timeout`
4. Add a `timeout` parameter to the `read_zarr` method that is passed to `default_timeout` parameter when creating the PolarisFS object.

Then you can change this line to: 

```suggestion
            raise PolarisHubError(
                f"Timeout error. Consider increasing the `timeout` parameter and try again."
            ) from timeout_error
```

polaris/hub/client.py
```diff
@@ -332,6 +333,33 @@ def get_dataset(self, owner: Union[str, HubOwner], name: str) -> Dataset:
...
+        except Exception as e:
+            raise PolarisHubError(f"Error opening Zarr store: {e}")
```
cwognum, line 361:
```suggestion
            raise PolarisHubError(f"Error opening Zarr store") from e
```

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,104 @@
+import fsspec
+
+from typing import Union, List, Dict, Any
+
+from polaris.utils.errors import PolarisHubError
+
+from httpx._types import TimeoutTypes
+
+
+class PolarisFSFileSystem(fsspec.AbstractFileSystem):
```
cwognum, line 10: Should we rename the class to `PolarisFileSystem` ? `FS` = `FileSystem`, right?

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,104 @@
...
+        **kwargs: Additional keyword arguments.
+    """
+
+    sep = "/"
+    protocol = "polarisfs"
+    async_impl = False
+
+    def __init__(
+        self,
+        polaris_client: Any,
```
cwognum, line 34:
```suggestion
        polaris_client: PolarisHubClient,
```

Although hat probably leads to cyclic imports? You could do a conditional import. See https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING. 

Something like:

```python
from typing import TYPE_CHECKING
if TYPE_CHECKING: 
    from polaris.hub import PolarisHubClient
```

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,104 @@
...
+        super().__init__(**kwargs)
+
+        self.polaris_client = polaris_client
+
+        # Prefix to remove from ls entries
+        self.prefix = f"dataset/{dataset_owner}/{dataset_name}/"
+        self.base_path = f"/storage/{self.prefix}"
+
+    def ls(
+        self, path: str, detail: bool = False, timeout: TimeoutTypes = (10, 200), **kwargs: dict
```
cwognum, line 49: In line with earlier comments: 

```suggestion
        self, path: str, detail: bool = False, timeout: Optional[TimeoutTypes] = None, **kwargs: dict
```

and then: 

```python
if timeout is None: 
    timeout = self._default_timeout
```

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,104 @@
...
+            return entries
+        else:
+            detailed_entries = [
+                {"name": p["name"][len(self.prefix) :], "size": p["size"], "type": p["type"]}
+                for p in response.json()
+            ]
+            return detailed_entries
+
+    def cat_file(
+        self, path: str, start: int = None, end: int = None, timeout: TimeoutTypes = (10, 200), **kwargs: dict
```
cwognum, line 78: In line with earlier comments: 

```suggestion
        self, path: str, start: int = None, end: int = None, timeout: Optional[TimeoutTypes] = None, **kwargs: dict
```

and then: 

```python
if timeout is None: 
    timeout = self._default_timeout
```

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,104 @@
...
+    sep = "/"
+    protocol = "polarisfs"
+    async_impl = False
+
+    def __init__(
+        self,
+        polaris_client: Any,
+        dataset_owner: str,
+        dataset_name: str,
+        default_expirations_seconds: int = 10 * 60,
```
cwognum, line 37: Not being used, right? Maybe in line with earlier comments, we can change this to `default_timeout`? 

```suggestion
        default_timeout: TimeoutTypes = (10, 200),
```

docs/api/hub.polarisfs.md
```diff
@@ -0,0 +1,5 @@
+::: polaris.hub.polarisfs.PolarisFSFileSystem
+    options: 
+        merge_init_into_class: true
+        filters: ["!^_", "!create_authorization_url", "!fetch_token", "!request", "!token"]
```
cwognum, line 4: Minor, but the other filters are note needed!

```suggestion
        filters: ["!^_"]
```

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def ls(
+        self, path: str, detail: bool = False, timeout: TimeoutTypes = (10, 200), **kwargs: dict
+    ) -> Union[List[str], List[Dict[str, Any]]]:
+        """List objects in the specified path within the Polaris dataset.
+
+        Args:
+            path: The path within the dataset to list objects.
+            detail: If True, returns detailed information about each object.
+            **kwargs: Additional keyword arguments.
```
cwognum, line 56: Since we do not use the `kwargs`, I would not document them!

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,104 @@
...
+
+class PolarisFSFileSystem(fsspec.AbstractFileSystem):
+    """
+    A file system interface for accessing datasets on the Polaris platform.
+
+    This class extends `fsspec.AbstractFileSystem` and provides methods to list objects within a Polaris dataset
+    and fetch the content of a file from the dataset.
+
+    Note: Zarr Integration
+        This file system can be used with Zarr for working with multidimensional array data stored in the Polaris dataset.
```
cwognum, line 18: This is great! 🚀 

Maybe we can extend this note even more to explain - at a high-level - why this is needed. We can also add a simple code example, I'm thinking: 

```python
fs = PolarisFileSystem(...)
store = zarr.storage.FSStore(..., fs=polaris_fs)
return zarr.open(store, mode="r")
```

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,104 @@
...
+        self, path: str, start: int = None, end: int = None, timeout: TimeoutTypes = (10, 200), **kwargs: dict
+    ) -> bytes:
+        """Fetches and returns the content of a file from the Polaris dataset.
+
+        Args:
+            path: The path to the file within the dataset.
+            start: The starting index of the content to retrieve.
+            end: The ending index of the content to retrieve.
+            timeout: Maximum time (in seconds) to wait for the request to complete.
+            **kwargs: Additional keyword arguments.
```
cwognum, line 87: If we're not using the `kwargs`, we should remove them from the docs! 

Looking at the [docs](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.cat_file) though, it seems we should use `open` and pass the `kwargs` there. 

Maybe: 

```python
with fsspec.open(signed_url, "rb", **kwargs) as f: 
    data = f.read()
return data[start:end]
```

Is that possible with signed URLs?

cwognum (commented):
Coming together nicely! We're almost there! 🔥

CICD seems to be failing due to a mismatch between `ruff` and `black` due to line length. I will look into this some more!

polaris/hub/client.py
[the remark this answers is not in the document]
cwognum: See my comment in the new review!

polaris/hub/client.py
jstlaurent, line 359: **Suggestion**
In Python, when catching an exception and raising another one, it's often a good idea to use the `from`clause to [chain the exceptions](https://realpython.com/python-raise-exception/#chaining-exceptions-with-the-from-clause) together. That makes debugging error easier.

```suggestion
            raise PolarisHubError(
                f"Timeout error: {timeout_error}. Consider increasing the expiration_seconds parameter and try again."
            ) from timeout_error
```

polaris/hub/client.py
jstlaurent, line 361: **Suggestion**
Here too, chaining the exception:
```suggestion
            raise PolarisHubError(f"Error opening Zarr store: {e}") from e
```

polaris/hub/polarisfs.py
jstlaurent, line 37: **Question**
Whenever I see numbers like these in the code, I have to ask whether putting with the other settings (in `polaris.hub.settings.PolarisHubSettings`) would be a better place to do so.

Even if the user doesn't customize it through an env variable, all the settings are in one place, and that makes it easier to manage.

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,104 @@
...
+            A list of dictionaries if detail is True; otherwise, a list of object names.
+        """
+        ls_path = f"{self.base_path}ls/{path}"
+
+        # GET request to Polaris Hub to list objects in path
+        response = self.polaris_client.get(ls_path.rstrip("/"), timeout=timeout)
+        response.raise_for_status()
+
+        if not detail:
+            entries = [p["name"][len(self.prefix) :] for p in response.json()]
```
jstlaurent, line 68: **Suggestion**
If your goal is to remove the prefix from the string, then a more reliable way is the [`removeprefix`](https://docs.python.org/3/library/stdtypes.html#str.removeprefix) method from the string class:

```suggestion
            entries = [p["name"].removeprefix(self.prefix) for p in response.json()]
```

polaris/hub/polarisfs.py
```diff
@@ -0,0 +1,104 @@
...
+            return entries
+        else:
+            detailed_entries = [
+                {"name": p["name"][len(self.prefix) :], "size": p["size"], "type": p["type"]}
```
jstlaurent, line 72: **Suggestion**
Same here:
```suggestion
                {"name": p["name"].removeprefix(self.prefix), "size": p["size"], "type": p["type"]}
```

polaris/hub/polarisfs.py
jstlaurent, line 78: **Suggestion**
I think the typing here should be:
```suggestion
        self, path: str, start: Union[int, None] = None, end: Union[int, None] = None, timeout: TimeoutTypes = (10, 200), **kwargs: dict
```

jstlaurent (approved):
I'll defer to Cas for most comments, since he knows the client better than I do. I did have some suggestions here and there, but nothing major.

polaris/hub/polarisfs.py
cwognum: Learned something new here! Didn't know this existed! 🎉

polaris/hub/polarisfs.py
lmtroper: @cwognum Sure, that makes sense! Would this then remove the need to pass `default_timeout` as a parameter to the FS? (e.g., as a parameter in `read_zarr_file`)

`def read_zarr_file( self, owner: Union[str, HubOwner], name: str, path: str, timeout: Optional[TimeoutTypes] = None) -> zarr.hierarchy.Group:`

cwognum: Let's remove it for now. 

Once we start thinking about a more integrated API that loads a `Dataset` and the underlying zarr archive, we will revisit how to let the user easily change the timeout.

cwognum:
Great work, @lmtroper !

PR #73, opened 2024-03-04
feature: polaris fsspec write implementation

lmtroper:
## Changelogs

Integrated a basic write implementation for Zarr files with the Polaris FSSpec. Some of the major changes include:
- Added a new endpoint in the client to be able to write a zarr file of a Polaris dataset.
- Added two new methods in the FSSpec: `rm` and `pipe_file`

#### More details of the FSSpec Implementation:
- The `rm` method is currently added as placeholder since it's required for the Zarr writing integration. 
- The `pipe_file` method calls the hub to get a signed URL for a PUT request, then writes the data to the bucket. The S3 bucket uses SHA-256 for data security but it is currently set to None.

---

_Checklist:_

---

### DIFF
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -118,7 +118,9 @@ class PolarisHubClient(OAuth2Client):
             **kwargs,
         )
 
-    def _load_from_signed_url(self, url: URLTypes, load_fn: Callable, headers: Optional[HeaderTypes] = None):
+    def _load_from_signed_url(
+        self, url: URLTypes, load_fn: Callable, headers: Optional[HeaderTypes] = None
+    ):
         """Utility function to load a file from a signed URL"""
         response = self.get(url, auth=None, headers=headers)  # type: ignore
         response.raise_for_status()
@@ -184,10 +186,14 @@ class PolarisHubClient(OAuth2Client):
     def fetch_token(self, **kwargs):
         """Light wrapper to automatically pass in the right URL"""
         return super().fetch_token(
-            url=self.settings.token_fetch_url, code_verifier=self.code_verifier, **kwargs
+            url=self.settings.token_fetch_url,
+            code_verifier=self.code_verifier,
+            **kwargs,
         )
 
-    def request(self, method, url, withhold_token=False, auth=httpx.USE_CLIENT_DEFAULT, **kwargs):
+    def request(
+        self, method, url, withhold_token=False, auth=httpx.USE_CLIENT_DEFAULT, **kwargs
+    ):
         """Wraps the base request method to handle errors"""
         try:
             response = super().request(method, url, withhold_token, auth, **kwargs)
@@ -203,8 +209,16 @@ class PolarisHubClient(OAuth2Client):
                     "SSL verification by setting the POLARIS_CA_BUNDLE environment variable to `false`."
                 ) from error
             raise error
-        except (MissingTokenError, InvalidTokenError, httpx.HTTPStatusError, OAuthError) as error:
-            if isinstance(error, httpx.HTTPStatusError) and error.response.status_code != 401:
+        except (
+            MissingTokenError,
+            InvalidTokenError,
+            httpx.HTTPStatusError,
+            OAuthError,
+        ) as error:
+            if (
+                isinstance(error, httpx.HTTPStatusError)
+                and error.response.status_code != 401
+            ):
                 raise
             raise PolarisUnauthorizedError from error
         return response
@@ -242,7 +256,9 @@ class PolarisHubClient(OAuth2Client):
     @property
     def user_as_owner(self) -> HubOwner:
         """Easily get the currently logged-in user a `HubOwner` instance."""
-        return HubOwner(user_id=self.user_info["user_id"], slug=self.user_info["username"])
+        return HubOwner(
+            user_id=self.user_info["user_id"], slug=self.user_info["username"]
+        )
 
     def login(self, overwrite: bool = False, auto_open_browser: bool = True):
         """Login to the Polaris Hub using the OAuth2 protocol.
@@ -272,7 +288,9 @@ class PolarisHubClient(OAuth2Client):
         authorization_url, _ = self.create_authorization_url()
 
         if auto_open_browser:
-            logger.info(f"Your browser has been opened to visit:\n{authorization_url}\n")
+            logger.info(
+                f"Your browser has been opened to visit:\n{authorization_url}\n"
+            )
             webbrowser.open_new_tab(authorization_url)
         else:
             logger.info(f"Please visit the following URL:\n{authorization_url}\n")
@@ -315,7 +333,9 @@ class PolarisHubClient(OAuth2Client):
             A `Dataset` instance, if it exists.
         """
 
-        response = self._base_request_to_hub(url=f"/dataset/{owner}/{name}", method="GET")
+        response = self._base_request_to_hub(
+            url=f"/dataset/{owner}/{name}", method="GET"
+        )
         storage_response = self.get(response["tableContent"]["url"])
 
         # This should be a 307 redirect with the signed URL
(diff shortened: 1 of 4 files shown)


polaris/hub/polarisfs.py
```diff
@@ -120,3 +128,71 @@ def cat_file(
         with fsspec.open(signed_url, "rb", **kwargs) as f:
             data = f.read()
         return data[start:end]
+
+    def rm(self, path: str, recursive: bool = False, maxdepth: Optional[int] = None):
```
cwognum, line 132: For the documentation GH CICD, you will need to specify a return type. 

```suggestion
    def rm(self, path: str, recursive: bool = False, maxdepth: Optional[int] = None) -> None:
```

cwognum (commented):
This look great already! Thank you for taking the time to document everything! :pray: 

For the failing code checks, it seems there is a mismatch between the line length we expect and the line length used. This should be automatic, as it is configured [in the `pyproject.toml`](https://github.com/polaris-hub/polaris/blob/main/pyproject.toml#L103-L114). After navigating to the root directory, you should be able to just run: 
```
ruff format .
```

It is a little weird though, because the error message states: 
```
Would reformat: polaris/hub/client.py
Would reformat: polaris/utils/types.py
2 files would be reformatted, 43 files already formatted
Error: Process completed with exit code 1.
```

But `types.py` is fine for me locally. Let's try again and I'll look into it some more if it's not fixed.

polaris/hub/polarisfs.py
```diff
@@ -120,3 +128,71 @@ def cat_file(
...
+
+        if response.status_code != 307:
+            raise PolarisHubError("Could not get signed URL from Polaris Hub.")
+
+        hub_response_body = response.json()
+        signed_url = hub_response_body["url"]
+
+        headers = {
+            "Content-Type": "application/octet-stream",
+            "x-amz-content-sha256": "UNSIGNED-PAYLOAD",
```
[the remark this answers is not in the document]
cwognum: Might be too slow?

polaris/hub/polarisfs.py
```diff
@@ -120,3 +128,71 @@ def cat_file(
...
+            maxdepth: The maximum depth to recurse when removing directories.
+
+        Returns:
+            None
+
+        Note:
+            This method currently it does not perform any removal operations and is included
+            as a placeholder that aligns with the Zarr interface's expectations.
+        """
+        return
```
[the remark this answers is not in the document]
lmtroper: @cwognum From what I noticed, if you want to write in a directory that already exists, it will try to remove it first. 

For example, if test.zarr directory is found in the dataset and I try to write to that existing path, it calls the `rm` file operation.
Screenshot 2024-03-04 at 3 13 44 PM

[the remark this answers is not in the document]
cwognum: Thank you for taking a look! If that's the only case in which it fails, I would prefer to still have it raise an exception to prevent a silent bug later on.

polaris/hub/polarisfs.py
```diff
@@ -120,3 +128,71 @@ def cat_file(
...
+            content: The content to be piped into the file.
+            timeout: Maximum time (in seconds) to wait for the request to complete.
+
+        Returns:
+            None
+        """
+        if timeout is None:
+            timeout = self.default_timeout
+
+        pipe_path = f"{self.base_path}put/{path}"
```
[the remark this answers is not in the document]
lmtroper: Since the second half of base_path comes from `prefix`, this would require removing the trailing slash from `base_path` using `r.strip("/")` since `prefix` needs it.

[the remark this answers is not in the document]
cwognum: Ah I see. That's a good point. I do think I prefer that.

Maybe you could then do: 
```python
self.base_path = f"/storage/{self.prefix}".rstrip("/")
```
in the constructor?

polaris/hub/polarisfs.py
```diff
@@ -1,6 +1,9 @@
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
-
+import datetime
 import fsspec
+import hashlib
+import time
```
cwognum, line 5: This is an unused import, because of which `ruff` fails. 

You can check the linting locally with: 
```
ruff check .
```

You can also see the CICD logs in the GH actions. For this PR, you can find these [here](https://github.com/polaris-hub/polaris/actions/runs/8177921121/job/22360677551?pr=73).

polaris/hub/client.py
```diff
@@ -333,13 +333,16 @@ def get_dataset(self, owner: Union[str, HubOwner], name: str) -> Dataset:
 
         return Dataset(**response)
 
-    def read_zarr_file(self, owner: Union[str, HubOwner], name: str, path: str) -> zarr.hierarchy.Group:
-        """Read a Zarr file from a Polaris dataset
+    def open_zarr_file(
+        self, owner: Union[str, HubOwner], name: str, path: str, mode: str
```
cwognum, line 337: Instead of the `str` type annotation, could we come up with a more specific type hint? From the [Zarr source code](https://github.com/zarr-developers/zarr-python/blob/a1fbedb18c1fc70f026c423fafca6d84ad88ce53/zarr/convenience.py#L49-L53), it seems Zarr supports a bit more than just `r` and `w`.

Maybe we can introduce a custom type in [`polaris/utils/types.py`](https://github.com/polaris-hub/polaris/blob/main/polaris/utils/types.py) For example: 
```python
IOMode: TypeAlias = Literal["r", "r+", "a", "w", "w-"]
```

polaris/hub/polarisfs.py
```diff
@@ -120,3 +131,76 @@ def cat_file(
...
+            None
+
+        Note:
+            This method currently it does not perform any removal operations and is included
+            as a placeholder that aligns with the Zarr interface's expectations.
+        """
+        try:
+            raise NotImplementedError("PolarisFS does not currently support the file removal operation.")
+        except NotImplementedError as e:
+            print(f"NotImplementedError: {e}")
```
cwognum, line 156: Do we need the `try-except` here because your local tests otherwise fail?

cwognum (commented):
Thanks @lmtroper ! Almost there!

polaris/hub/polarisfs.py
cwognum: I can remember this could fail if you're writing to a same location that already has a Zarr directory. Does it fail otherwise as well?

lmtroper: @cwognum Yes, the case I mentioned earlier only occurs when you open the zarr file (and when the path you provide already exists). But when I was writing data to the path, at some point it runs the `rm` function and with this way, we can see when the `rm` function is being called but without stopping the writing operations.

cwognum: If I understand correctly, that implies that Zarr reorganizes the directory as it is writing it (deleting files it had created earlier). That's interesting, because it might mean that it would be more efficient to create the Zarr archive locally and then [copying](https://zarr.readthedocs.io/en/stable/tutorial.html#copying-migrating-data) the data to Cloudflare only when it's finished (specifically, see [`zarr.convenience.copy_store()`](https://zarr.readthedocs.io/en/stable/api/convenience.html#zarr.convenience.copy_store)). 

For now, it feels a little silly to raise and catch the exception ourselves. What do you think of: 
```suggestion
        logger.critical(f"Trying to remove {path}, but the rm() functionality is not implemented yet")
```

with `logger` coming from `import logger from loguru`.

lmtroper: Just tried it out this morning and seems like it is no longer an issue. Will remove the try-except!

polaris/hub/polarisfs.py
```diff
@@ -120,3 +129,73 @@ def cat_file(
...
+
+        hub_response_body = response.json()
+        signed_url = hub_response_body["url"]
+
+        sha256_hash = hashlib.sha256(content).hexdigest()
+
+        headers = {
+            "Content-Type": "application/octet-stream",
+            "x-amz-content-sha256": sha256_hash,
+            "x-amz-date": datetime.datetime.utcnow().strftime("%Y%m%dT%H%M%SZ"),
```
cwognum, line 189: `utcnow()` is [being deprecated](https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated), so this is causing warnings! We should instead use:

```python
from datetime import datetime, timezone
datetime.now(timezone.utc)
```

cwognum (approved):
Very polished! 💅 

One minor comment still, but this is great work! Thank you @lmtroper !

cwognum:
Feel free to merge this PR whenever you're ready! 🙏

PR #77, opened 2024-03-14
Add additional features to support SDF to Zarr conversion

cwognum:
## Changelogs

- Made it possible to use _slices_ in pointers (e.g. `path/to/data.zarr/col#0:10`).
- Added the `SDFConverter` class to convert a SDF object to a Polaris Dataset.
- Started the implementation of a `DatasetFactory` as a unified system for creating datasets.
- Added the `create_dataset_from_file` utility method.
- Added `Adapter`s to convert datapoints during data loading. These can be saved alongside a dataset.

---

_Checklist:_

---

This PR started to add support for the [MARCEL benchmark](https://github.com/SXKDZ/MARCEL/tree/main).

There were three critical changes to support MARCEL:
- We needed to convert a SDF file to Zarr, which we ended up doing using RDKit's bytestrings. See also the discussion https://github.com/rdkit/rdkit/discussions/7235
- We needed to allow for a variable number of conformers per datapoint, which we did by adding support for pointers with slice indexing (e.g. `path/to/data.zarr/col#0:10`)
- Allow users to specify an `Adapter`, which during data loading adapts the content saved in the Zarr file to a desired format (in this case, from a bytestring to a RDKit Mol object).

In addition, I long wanted to start on a system to make it easier to create dataset. I'm quite happy with this as a first version. It's a little crude still, but think it will be easy to extend as we work to make adding datasets more and more easy.

I added a tutorial and docstrings, but here is a short snippet: 
```python
from polaris.dataset import DatasetFactory
from polaris.dataset.converters import SDFConverter

# Create a new factory object
save_dst = "/path/to/destination.zarr
factory = DatasetFactory(zarr_root_path=save_dst)

# Register a converter for the SDF file format
factory.register_converter("sdf", SDFConverter())

# Process your SDF file
factory.add_from_file("/path/to/data.sdf")

# Build the dataset
dataset = factory.build()
```


polaris/dataset/_factory.py
```diff
@@ -0,0 +1,229 @@
...
+        dataset = create_dataset_from_file("path/to/my_dataset.sdf")
+        ```
+
+    Question: How to make adding meta-data easier?
+        The `DatasetFactory` is designed to more easily pull together data from different sources.
+        However, adding meta-data remains a laborous process. How could we make this simpler through
+        the Python API?
+    """
+
+    def __init__(self, zarr_root_path: Optional[str] = None) -> None:
```
jstlaurent, line 56: **Suggestion**
Perhaps it would make this API more flexible if the user could specify the converters they want directly here, when initializing the factory?

jstlaurent (approved):
I'm a bit of my depth here, but that's some nice test coverage and I like concepts like pluggable converters.

polaris/dataset/_adapters.py
```diff
@@ -0,0 +1,22 @@
...
+
+class Adapter(Enum):
+    """
+    Adapters are predefined callables that change the format of the data.
+    Adapters are serializable and can thus be saved alongside datasets.
+
+    Attributes:
+        SMILES_TO_MOL: Convert a SMILES string to a RDKit molecule.
+        BYTES_TO_MOL: Convert a RDKit binary string to a RDKit molecule.
+    """
```
zhu0619, line 14: I can imagine extending the adaptors with `SMARTS_TO_MOL` etc.

zhu0619 (approved):

polaris/dataset/_adapters.py
cwognum: For sure! More generally, I think we could add all sorts of adapters. To keep things simple, however, I think we should add these as they are needed / requested.

PR #79, opened 2024-03-18
chore: Support uv

jstlaurent:
## Changelogs

- Synch the dependencies in pyproject.toml with those in env.yml
- Add a some commands to run when using uv for local development 

---

_Checklist:_

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,18 +93,26 @@ pip install polaris-lib
 
 ### Setup dev environment
 
-```bash
+```shell
 conda env create -n polaris -f env.yml
 conda activate polaris
 
 pip install --no-deps -e .
 ```
 
+Alternatively, using [uv](https://github.com/astral-sh/uv):
+```shell
+uv venv -p 3.12 polaris
+source .venv/polaris/bin/activate
+uv pip compile pyproject.toml -o requirements.txt --all-extras
+uv pip install -r requirements.txt 
+```
+
 ### Tests
 
 You can run tests locally with:
 
-```bash
+```shell
 pytest
 ```
 
--- a/.gitignore
+++ b/.gitignore
@@ -135,3 +135,6 @@ rever/
 
 # VS Code
 .vscode/
+
+# Generated requirements.txt
+requirements.txt
--- a/env.yml
+++ b/env.yml
@@ -10,6 +10,7 @@ dependencies:
   - pyyaml
   - pydantic >=2
   - pydantic-settings >=2
+  - fsspec
 
   # Hub client
   - authlib
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,9 +49,31 @@ dependencies = [
     "scipy",
     "scikit-learn",
     "seaborn",
-    "datamol >=0.12",
+    "datamol >=0.12.1",
     "zarr",
     "pyarrow",
+    "fsspec[http]"
+]
+
+[project.optional-dependencies]
+dev = [
+    "pytest",
+    "pytest-xdist",
+    "pytest-cov",
+    "ruff",
+    "jupyterlab",
+    "ipywidgets"
+]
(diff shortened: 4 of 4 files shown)


README.md
```diff
@@ -93,18 +93,26 @@ pip install polaris-lib
...
 pip install --no-deps -e .
 ```
 
+Alternatively, using [uv](https://github.com/astral-sh/uv):
+```shell
+uv venv -p 3.12 polaris
+source .venv/polaris/bin/activate
+uv pip compile pyproject.toml -o requirements.txt --all-extras
+uv pip install -r requirements.txt 
+```
```
cwognum, line 109: By default, I would like to keep the README concise to not overwhelm with too much information. I would thus recommend a single way of installing it, but offer alternative installation options through a expandable component (similar to e.g. [Pytorch Lightning](https://github.com/Lightning-AI/pytorch-lightning) does it!). 

Something like: 

```suggestion
<details>
  <summary>Other installation options</summary>
  
    Alternatively, using [uv](https://github.com/astral-sh/uv):
    ```shell
    uv venv -p 3.12 polaris
    source .venv/polaris/bin/activate
    uv pip compile pyproject.toml -o requirements.txt --all-extras
    uv pip install -r requirements.txt 
    ```   
</details>

```

cwognum (approved):
Looks great! 

Do you know if there's an automated way to keep things in sync? I assume not and I expect it's not that big an issue because we won't frequently update it, but just out of curiosity.

Finally, if you're curious, I would also recommend to give [`pixi`](https://github.com/prefix-dev/pixi) a look. From the creators of `mamba`. It's built on top of Conda (and thus cross-platform and multi-language), fast (written in Rust) and feels like a more complete package manager (e.g. it has lock files, [task manager](https://pixi.sh/latest/advanced/advanced_tasks/)). `pixi` actually wants to integrate PyPi and Conda workflows and has written its own PyPi resolver called [`rip`](https://github.com/prefix-dev/rip). It seems though they are now switching to `uv` instead! See https://prefix.dev/blog/uv_in_pixi

cwognum:
The failing test cases seem to be an upstream issue. Maybe we need to set a limit on `pandas`, `datamol` or `rdkit`?

cwognum (approved):

jstlaurent:
> The failing test cases seem to be an upstream issue. Maybe we need to set a limit on `pandas`, `datamol` or `rdkit`?

I put an upper limit on `pandas`. It looks like rdkit tries to patch a function that's moved around in 2.2.0.

jstlaurent:
> Do you know if there's an automated way to keep things in sync? I assume not and I expect it's not that big an issue because we won't frequently update it, but just out of curiosity.

I don't believe so, unfortunately. We'd need to set something up ourselves in CI, if we wanted that.
 
> Finally, if you're curious, I would also recommend to give [`pixi`](https://github.com/prefix-dev/pixi) a look. From the creators of `mamba`. It's built on top of Conda (and thus cross-platform and multi-language), fast (written in Rust) and feels like a more complete package manager (e.g. it has lock files, [task manager](https://pixi.sh/latest/advanced/advanced_tasks/)). `pixi` actually wants to integrate PyPi and Conda workflows and has written its own PyPi resolver called [`rip`](https://github.com/prefix-dev/rip). It seems though they are now switching to `uv` instead! See https://prefix.dev/blog/uv_in_pixi

Hadrien had mentioned it. I had a look and it seems like the goals are pretty much the same as with uv (becoming Cargo for Python). I personally prefer uv's leveraging the Python standards through `pyproject.toml`, rather than using a custom format, like pixi's. It doesn't force potential users to use one specific tool.

cwognum:
I agree! That's why I haven't switched, but they have some PRs on importing / exporting requirements to other formats.

PR #82, opened 2024-03-26
feat: consolidating metadata for opening zarr files in read mode

lmtroper:
## Changelogs

Incorporated Zarr function that [consolidates metadata](https://zarr.readthedocs.io/en/stable/tutorial.html#consolidating-metadata) when opening a Zarr group in read mode. This results in reduced number of `ls` calls and increases the speed in reading from the Zarr group.


polaris/hub/client.py
```diff
@@ -359,6 +359,9 @@ def open_zarr_file(
 
         try:
             store = zarr.storage.FSStore(path, fs=polaris_fs)
+            if mode in ["r", "r+"]:
+                zarr.consolidate_metadata(store)
+                return zarr.open_consolidated(store, mode=mode)
```
cwognum, line 364: No need to call `open_consolidated` again, since the return value of `zarr.consolidate_metadata` is the root opened in consolidated mode. See [the source code](https://zarr.readthedocs.io/en/stable/_modules/zarr/convenience.html#consolidate_metadata).

cwognum (changes requested):
To optimize performance, we don't want to consolidate the meta-data everytime. 

From [the Zarr docs](https://zarr.readthedocs.io/en/stable/tutorial.html#consolidating-metadata): 

> ```
> >>> zarr.consolidate_metadata(store)  
> ```
> This creates a special key with a copy of all of the metadata from all of the metadata objects in the store.

The _key_ here refers to a single file for us (i.e. by default named `.zmetadata`, although this can be combined with the `metadata_key` parameter in the consolidate methods).

I think what we would want to do is: 
1. Consolidate the archive locally. 
2. Then copy over the _consolidated_ archive to the Hub with [`zarr.convenience.copy_all`](https://zarr.readthedocs.io/en/stable/api/convenience.html#zarr.convenience.copy_all).
3. If this doesn't work (i.e. I assume it would copy over the `.zmetadata` file, but maybe not?), then we should find out a way to copy over this single file manually. 

Could you look into the above? I would be curious to know if this is possible!

Not having to consolidate everything on the Hub would make things a lot faster!

cwognum:
Now that https://github.com/polaris-hub/polaris/pull/83 is merged, could you actually look into adding the consolidation in the flow for using Zarr datasets from the Hub:

- We want to consolidate the Zarr archive just before uploading it to the Hub ([here](https://github.com/polaris-hub/polaris/blob/main/polaris/hub/client.py#L589-L598))
- We assume that any archive uploaded to the Hub has been consolidated, so when we open it [here](https://github.com/polaris-hub/polaris/blob/main/polaris/dataset/_dataset.py#L223) it should use `open_consolidated`. Maybe we add an `as_consolidated` parameter to `client.open_zarr_file()`?

polaris/hub/client.py
```diff
@@ -587,13 +590,17 @@ def upload_dataset(
...
                 dest = self.open_zarr_file(
                     owner=dataset.owner,
                     name=dataset.name,
                     path=dataset_json["zarrRootPath"],
                     mode="w",
                 )
+                # Locally consolidate the archive metadata
+                zarr.consolidate_metadata(dataset.zarr_root.store)
+                zmetadata_content = dataset.zarr_root.store[".zmetadata"]
+                dest.store[".zmetadata"] = zmetadata_content
```
cwognum, line 602: Let's reference https://github.com/zarr-developers/zarr-python/issues/1731 in the docs here! This is a simple enough solution. My worry is that one day they might change how a consolidated archive is saved and we wouldn't notice it until things start breaking. Once there is an answer on that issue, we can update this code to reflect the recommendation from the Zarr developers.

cwognum (commented):
Great work! 

Let's assume that any Zarr archive that is loaded for a dataset has been consolidated. This means that we should also change [these lines of code](https://github.com/polaris-hub/polaris/blob/main/polaris/dataset/_dataset.py#L221-L225) to load in consolidated mode!

polaris/hub/client.py
```diff
@@ -587,13 +593,20 @@ def upload_dataset(
         if dataset.zarr_root is not None:
             with tmp_attribute_change(self.settings, "default_timeout", timeout):
                 # Copy the Zarr archive to the hub
-                # This does not copy the consolidated data
                 dest = self.open_zarr_file(
                     owner=dataset.owner,
                     name=dataset.name,
                     path=dataset_json["zarrRootPath"],
                     mode="w",
```
cwognum, line 600: This will fail, no? Because `is_consolidated` equals `True` and `mode == "w"`?

polaris/dataset/_dataset.py
```diff
@@ -222,7 +222,7 @@ def zarr_root(self):
             if saved_on_hub:
                 self._zarr_root = self.client.open_zarr_file(self.owner, self.name, self.zarr_root_path, "r+")
             else:
-                self._zarr_root = zarr.open(self.zarr_root_path, "r+")
+                self._zarr_root = zarr.open_consolidated(self.zarr_root_path, "r+")
```
cwognum, line 225: It would be great if we can somehow check whether an archive has been consolidated or not, so we can throw more informative errors. Even for people familiar with Zarr, this might not be intuitive. 

We could either: 
1. Use `try-except` and raise a custom error (e.g. `PolarisZarrError`)
2. Check if the `.zmetadata` file exists before opening the archive. 

I think I would prefer the latter option. What do you think?

cwognum (commented):
Almost there! 

The test cases are failing though because the test Zarr archive is not consolidated! The formatting also fails right now.

polaris/dataset/_dataset.py
lmtroper: Yes, I don't think we should throw an error if for some reason it is there but not consolidated so I agree with the second option as well.

polaris/dataset/_dataset.py
```diff
@@ -222,7 +222,12 @@ def zarr_root(self):
             if saved_on_hub:
                 self._zarr_root = self.client.open_zarr_file(self.owner, self.name, self.zarr_root_path, "r+")
             else:
-                self._zarr_root = zarr.open_consolidated(self.zarr_root_path, "r+")
+                try:
+                    # Attempt to access .zmetadata to open consolidated file
+                    _ = self.zarr_root.store[".zmetadata"]
+                    self._zarr_root = zarr.open_consolidated(self.zarr_root.store, mode="r")
+                except KeyError:
+                    self._zarr_root = zarr.open(self.zarr_root.store, mode="r+")
```
cwognum, line 230: I don't think this will work, because it recurses indefinitely. In `zarr_root()` you call `zarr_root()`. 

I also think I explained myself poorly! :sweat_smile: 

I suggest we _do_ throw an error. That way, if people create a custom Zarr archive themselves and it doesn't work, they get an informative error on what went wrong. I suggest: 

```py
try:
    if saved_on_hub:
        self._zarr_root = self.client.open_zarr_file(self.owner, self.name, self.zarr_root_path, "r+")
    else:
        self._zarr_root = zarr.open_consolidated(self.zarr_root.store, mode="r")
except KeyError as error:
    raise InvalidDatasetError("A Zarr archive associated with a Polaris dataset has to be consolidated.")
```

lmtroper: Would explain why all the tests are failing 😅 I will update it with these changes!

cwognum: You will still have to update the test cases I believe because that [Zarr archive](https://github.com/polaris-hub/polaris/blob/main/tests/conftest.py#L78-L84) we use is not consolidated! :slightly_smiling_face:

cwognum (approved):

PR #84, opened 2024-03-26
Added the version attribute to the artifact data model

cwognum:
## Changelogs

- Adds the `version` attribute to the `BaseArtifactModel`. 

---

_Checklist:_

---

If we ever want to change a function that would break backwards compatibility, we now have the opportunity to maintain backwards compatibility by adding conditional logic based on the version. Not saying this should be our preferred approach, but I think it's nevertheless good to have the opportunity.


jstlaurent (approved):
Good idea. We'll probably like having that info at some point in the future.

PR #86, opened 2024-04-02
Updating column adapter logic and usage

Andrewq11:
## Changelogs

- Updated the attribute name in the BaseArtifactModel which indicates the current Polaris client version from `version` to `polaris_version`
- Fixed column adapter logic by introducing a conversion map and updating how the conversion is applied on dataset retrieval

---

_Checklist:_

---

### DIFF
--- a/polaris/dataset/_adapters.py
+++ b/polaris/dataset/_adapters.py
@@ -2,10 +2,8 @@ from enum import Enum
 import datamol as dm
 
 # Map of conversion operations which can be applied to dataset columns
-conversion_map = {
-    'SMILES_TO_MOL': dm.to_mol,
-    'BYTES_TO_MOL': dm.Mol
-}
+conversion_map = {"SMILES_TO_MOL": dm.to_mol, "BYTES_TO_MOL": dm.Mol}
+
 
 class Adapter(Enum):
     """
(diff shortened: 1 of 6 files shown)


polaris/dataset/_adapters.py
```diff
@@ -13,10 +15,10 @@ class Adapter(Enum):
         BYTES_TO_MOL: Convert a RDKit binary string to a RDKit molecule.
     """
 
-    SMILES_TO_MOL = dm.to_mol
-    BYTES_TO_MOL = dm.Mol
+    SMILES_TO_MOL = "SMILES_TO_MOL"
+    BYTES_TO_MOL = "BYTES_TO_MOL"
```
cwognum, line 19: Minor, but since the `name` and the `value` of the Enum are the same, maybe we should just use [`auto()`](https://docs.python.org/3/library/enum.html#enum.auto) as the value to prevent confusion on whether to use the `name` or the `value`.

polaris/dataset/_dataset.py
```diff
@@ -154,9 +154,9 @@ def _validate_model(cls, m: "Dataset"):
 
         return m
 
-    @field_validator("default_adapters")
+    @field_validator("default_adapters", mode="before")
```
cwognum, line 157: Nice usage of `before` here! We should probably use that in other places too so we can simplify the type definitions of our Pydantic models.

polaris/dataset/_adapters.py
cwognum: To be really safe, maybe we combine it with the [`@unique`](https://docs.python.org/3/library/enum.html#enum.unique) decorator.

cwognum (approved):
Looks great. Just a minor suggestion! Thanks, @Andrewq11 !

PR #89, opened 2024-04-03
Preventing certain license types on dataset upload to hub

Andrewq11:
## Changelogs

- Added an enum for the Creative Commons licenses which we will accept for datasets being uploaded to the hub
- Added logic to throw an error on dataset upload attempts which contain an unsupported license type

---

_Checklist:_

---

- This PR is related to this issue - https://github.com/polaris-hub/polaris-hub/pull/281

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,7 +10,7 @@ from polaris.benchmark import (
     SingleTaskBenchmarkSpecification,
 )
 from polaris.dataset import ColumnAnnotation, Dataset
-from polaris.utils.types import HubOwner, License
+from polaris.utils.types import HubOwner
 
 
 def check_version(artifact):
@@ -68,7 +68,7 @@ def test_dataset(test_data, test_org_owner):
         tags=["tagA", "tagB"],
         user_attributes={"attributeA": "valueA", "attributeB": "valueB"},
         owner=test_org_owner,
-        license=License(id="CC-BY-4.0"),
+        license="CC-BY-4.0",
         curation_reference="https://www.example.com",
     )
     check_version(dataset)
--- a/tests/test_type_checks.py
+++ b/tests/test_type_checks.py
@@ -3,7 +3,7 @@ from pydantic import ValidationError
 
 import polaris as po
 from polaris._artifact import BaseArtifactModel
-from polaris.utils.types import HubOwner, License
+from polaris.utils.types import HubOwner
 
 
 def test_slug_string_type():
@@ -55,14 +55,6 @@ def test_slug_compatible_string_type():
         "x" * 4,
     ]:
         BaseArtifactModel(name=name)
-
-
-def test_license():
-    # If not a valid CC license, you must specify a valid reference
-    with pytest.raises(ValidationError):
-        License(id="MIT")
-    with pytest.raises(ValidationError):
-        License(id="0BSD")
 
 
 def test_version():


polaris/dataset/_dataset.py
```diff
@@ -58,7 +58,7 @@ class Dataset(BaseArtifactModel):
         annotations: Each column _can be_ annotated with a [`ColumnAnnotation`][polaris.dataset.ColumnAnnotation] object.
             Importantly, this is used to annotate whether a column is a pointer column.
         source: The data source, e.g. a DOI, Github repo or URI.
-        license: The dataset license
+        license: The dataset license. Polaris only supports some Creative Commons licenses. See the SupportedLicenseType enum for accepted ID values.
```
cwognum, line 61: You can reference other code in your docstring by using double square brackets, e.g. 

```
[`SupportedLicensesTyp`][polaris.utils.types.SupportedLicenseType]
```

polaris/utils/types.py
```diff
@@ -128,7 +138,7 @@ class License(BaseModel):
         "https://raw.githubusercontent.com/spdx/license-list-data/main/json/licenses.json"
     )
 
-    id: str
+    id: SupportedLicenseType | str
     reference: Optional[HttpUrlString] = None
 
     @model_validator(mode="after")  # type: ignore
```
cwognum, line 144: We no longer need this validator!

We should, however, add a validator to convert the `id` field into the enum if it is a string and to not have to convert between str and enum.

polaris/utils/types.py
```diff
@@ -128,7 +138,7 @@ class License(BaseModel):
         "https://raw.githubusercontent.com/spdx/license-list-data/main/json/licenses.json"
     )
 
-    id: str
+    id: SupportedLicenseType | str
     reference: Optional[HttpUrlString] = None
```
cwognum, line 142: We should automatically compute the `reference` attribute based on the `id` if the `id` is not `None`. Because we don't want the user to be able to specify a custom link anymore, we should turn the `reference` into a [Pydantic `@computed_field`](https://docs.pydantic.dev/2.0/usage/computed_fields/).

polaris/utils/types.py
```diff
@@ -115,11 +115,21 @@ def __str__(self):
         return self.slug
 
 
+class SupportedLicenseType(Enum):
```
cwognum, line 118: I'm not sure if we need an Enum here. I suggest we just define a new `TypeAlias` for a `Literal` to prevent confusion on whether the use the name or key of the enum.

polaris/utils/types.py
cwognum: Not needed if you decide to replace the enum by a literal type!

polaris/hub/client.py
```diff
@@ -520,6 +520,13 @@ def upload_dataset(
                 See also: https://www.python-httpx.org/advanced/#timeout-configuration
             owner: Which Hub user or organization owns the artifact. Takes precedence over `dataset.owner`.
         """
+
+        # Check if the specified license for the dataset is supported
+        if dataset.license and dataset.license.id not in SupportedLicenseType:
+            raise InvalidDatasetError(
+                f"\nThe license type you have specified for this dataset is not supported by the Polaris Hub.\nOnly some Creative Commons licenses are supported - {[member.value for member in SupportedLicenseType]}. For more information, see https://creativecommons.org/share-your-work/cclicenses/"
+            )
```
cwognum, line 528: This is probably out of scope for this PR, but I'm thinking how we should go about doing the checks before uploading to the Hub (we'll need more) more robustly.

Two ideas:
1. We make the data model for the Dataset more stringent, such that it matches what is expected by the Hub.
2. We create a subclass of the Dataset (and Benchmark and Result) that are more stringent and we validate against this Pydantic model on upload.

Thoughts?

cwognum (commented):
Thanks for picking this up @Andrewq11 ! 

I had some suggestions. Let me know what you think!

polaris/hub/client.py
```diff
@@ -520,6 +520,13 @@ def upload_dataset(
                 See also: https://www.python-httpx.org/advanced/#timeout-configuration
             owner: Which Hub user or organization owns the artifact. Takes precedence over `dataset.owner`.
         """
+
+        # Check if the specified license for the dataset is supported
+        if dataset.license and dataset.license.id not in SupportedLicenseType:
```
cwognum, line 525: Missed something: The license should always defined when uploading to the hub! Can no longer be None!

polaris/hub/client.py
cwognum: Orrrr we keep relying on the errors returned by the Hub to prevent code duplication. This is what we currently do, but the errors have been confusing for some users so we would have to better format then

cwognum: Yeah, agreed. I believe the errors we throw in case there is mismatch between expected and provided data structure, have a JSON field (see [here](https://github.com/polaris-hub/polaris-hub/blob/main/src/lib/errors.ts#L77-L81)) so we should be able to parse and format these more nicely.

cwognum: @Andrewq11 Did you create an issue for error formatting already? We can reference this thread in that issue.

cwognum (approved):

PR #90, opened 2024-04-04
Fix: Passing wrong Zarr store that caused metadata consolidation error in upload_dataset() 

lmtroper:
## Changelogs

When consolidating the metadata locally, we were passing different stores which was throwing the error `memoryview: a bytes-like object is required, not 'dict'`. The fix was to append another `.store` when consolidating the metadata.

E.g., `dataset.zarr_root.store`  =>  <zarr.storage.ConsolidatedMetadataStore object at 0x13ddfaba0>
`dataset.zarr_root.store.store`  => <zarr.storage.DirectoryStore object at 0x13d900980>


cwognum (approved):
Good find! Thanks @lmtroper !

PR #93, opened 2024-04-09
feat: Listings Caching implementation using fsspec dircache

lmtroper:
## Changelogs

Implementing a listings caching system using fsspec's dircache to reduce the number of `ls` calls made during zarr upload/download. 

The fsspec `info()` method was the method that was resulting in all of the `ls` calls being made during upload so I tried targeting this method by using a listings caching. Given that listings caching is activated, each call to `ls` checks the path for cached listings and if none are found, it will check the listings for the parent path. This was an idea to solve the problem of checking child paths even though we learn that the parent path is empty:

```ls_path:  /storage/dataset/lmtroper/93a16e4534324def975d5a97285722c5/ls/
[]
ls_path:  /storage/dataset/lmtroper/93a16e4534324def975d5a97285722c5/ls/data.zarr
[]
ls_path:  /storage/dataset/lmtroper/93a16e4534324def975d5a97285722c5/ls/data.zarr/.zgroup
[]
pipe_path:  /storage/dataset/lmtroper/93a16e4534324def975d5a97285722c5/data.zarr/.zgroup
ls_path:  /storage/dataset/lmtroper/93a16e4534324def975d5a97285722c5/ls/data.zarr
[{'name': 'dataset/lmtroper/93a16e4534324def975d5a97285722c5/data.zarr/.zgroup', 'size': 24, 'type': 'file'}]
```

To update the listings caching, it can happen during `ls` if the path and parent path have no listings, or during `pipe_file` to update the list of cachings for a given path. 

This all reduces it down to one actual `ls` call during upload (right at the beginning to verify an empty folder) but we could signal refresh to `True` when we upload a file, which would still reduce the number of total ls calls (and maybe a more reliable way?).

### DIFF
--- a/polaris/hub/polarisfs.py
+++ b/polaris/hub/polarisfs.py
@@ -3,6 +3,7 @@ from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 import fsspec
+from fsspec.dircache import DirCache
 
 from polaris.utils.errors import PolarisHubError
 from polaris.utils.types import TimeoutTypes
@@ -44,9 +45,18 @@ class PolarisFileSystem(fsspec.AbstractFileSystem):
         polaris_client: "PolarisHubClient",
         dataset_owner: str,
         dataset_name: str,
+        use_listings_cache: bool = True,
+        listings_expiry_time=None,
+        max_paths=None,
         **kwargs: dict,
     ):
         super().__init__(**kwargs)
+
+        self.use_listings_cache = use_listings_cache
+        if self.use_listings_cache:
+            self.dircache = DirCache(listings_expiry_time=listings_expiry_time, max_paths=max_paths)
+        else:
+            self.dircache = None
 
         self.polaris_client = polaris_client
         self.default_timeout = self.polaris_client.settings.default_timeout
@@ -54,6 +64,8 @@ class PolarisFileSystem(fsspec.AbstractFileSystem):
         # Prefix to remove from ls entries
         self.prefix = f"dataset/{dataset_owner}/{dataset_name}/"
         self.base_path = f"/storage/{self.prefix.rstrip('/')}"
+
+        self.refresh = False
 
     @staticmethod
     def is_polarisfs_path(path: str) -> bool:
@@ -89,14 +101,21 @@ class PolarisFileSystem(fsspec.AbstractFileSystem):
 
         ls_path = self.sep.join([self.base_path, "ls", path])
 
+        if self.use_listings_cache and not self.refresh:
+            cached_listings = self.dircache.get(path, None)
+            if cached_listings is not None:
+                return cached_listings if detail else [d["name"] for d in cached_listings]
+
+            cached_listings = self.dircache.get(self._parent(path), None)
+            if cached_listings is not None:
+                self.dircache[path] = []
+                return cached_listings if detail else [d["name"] for d in cached_listings]
+
         # GET request to Polaris Hub to list objects in path
         response = self.polaris_client.get(ls_path.rstrip("/"), timeout=timeout)
         response.raise_for_status()
 
-        if not detail:
-            return [p["name"].removeprefix(self.prefix) for p in response.json()]
-
-        return [
+        detailed_listings = [
             {
                 "name": p["name"].removeprefix(self.prefix),
                 "size": p["size"],
@@ -104,6 +123,14 @@ class PolarisFileSystem(fsspec.AbstractFileSystem):
             }
             for p in response.json()
         ]
+
+        if self.use_listings_cache:
+            self.dircache[path] = detailed_listings
+
+        if not detail:
+            return [p["name"].removeprefix(self.prefix) for p in response.json()]
+
+        return detailed_listings
 
     def cat_file(
         self,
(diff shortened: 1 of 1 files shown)


polaris/hub/polarisfs.py
```diff
@@ -44,17 +45,28 @@ def __init__(
...
         **kwargs: dict,
     ):
         super().__init__(**kwargs)
 
+        self.use_listings_cache = use_listings_cache
+        if self.use_listings_cache:
+            self.dircache = DirCache(listings_expiry_time=listings_expiry_time, max_paths=max_paths)
+        else:
+            self.dircache = None
+
```
cwognum, line 60: We can create the dircache through the parent class, see e.g. [here](https://github.com/fsspec/filesystem_spec/blob/518984d494e48f007d8aa48de882ed7014da2574/fsspec/spec.py#L149).

I also wouldn't let people configure it. Everyone should use this - It's faster!
```suggestion
    ):
        super().__init__(
            use_listing_cache=True, 
            listings_expiry_time: None,
            max_paths: None,
       )
```

polaris/hub/polarisfs.py
```diff
@@ -89,14 +101,21 @@ def ls(
...
 
+        if self.use_listings_cache and not self.refresh:
+            cached_listings = self.dircache.get(path, None)
+            if cached_listings is not None:
+                return cached_listings if detail else [d["name"] for d in cached_listings]
+
+            cached_listings = self.dircache.get(self._parent(path), None)
+            if cached_listings is not None:
+                self.dircache[path] = []
+                return cached_listings if detail else [d["name"] for d in cached_listings]
```
cwognum, line 112: There is the [`_ls_from_cache`](https://github.com/fsspec/filesystem_spec/blob/518984d494e48f007d8aa48de882ed7014da2574/fsspec/spec.py#L354-L377) method in the parent class.

polaris/hub/polarisfs.py
```diff
@@ -105,6 +124,14 @@ def ls(
             for p in response.json()
         ]
 
+        if self.use_listings_cache:
+            self.dircache[path] = detailed_listings
```
cwognum, line 128: `self.use_listings_cache` should always exist for us, so we can simplify this to:

```suggestion
        self.dircache[path] = detailed_listings
```

polaris/hub/polarisfs.py
```diff
@@ -105,6 +124,14 @@ def ls(
...
+
+        if not detail:
+            return [p["name"].removeprefix(self.prefix) for p in response.json()]
```
cwognum, line 131: `removeprefix` is not needed anymore right? Already done when creating `detailed_listings`?

polaris/hub/polarisfs.py
```diff
@@ -212,3 +239,30 @@ def pipe_file(
...
+        path = path.rstrip("/")
+        out1 = [o for o in out if o["name"].rstrip("/") == path]
+        if len(out1) == 1:
+            if "size" not in out1[0]:
+                out1[0]["size"] = None
+            return out1[0]
+        elif len(out1) > 1:
+            return {"name": path, "size": 0, "type": "directory"}
+        else:
+            raise FileNotFoundError(path)
```
cwognum, line 268: This is exactly the same as it is in the parent class, right? No need to overwrite it?

polaris/hub/polarisfs.py
```diff
@@ -212,3 +239,30 @@ def pipe_file(
...
         response.raise_for_status()
+
+        if self.use_listings_cache:
+            new_listing = [{"name": path, "type": "file", "size": len(content)}]
+            current_listings = self.dircache.get(self._parent(path), None)
+            if current_listings is not None:
+                new_listing.extend(current_listings)
+                self.dircache[self._parent(path)] = new_listing
+            else:
+                self.dircache[self._parent(path)] = new_listing
```
cwognum, line 250: Good thinking to add this here!

Can be simplified a bit if we assume `self.use_listings_cache` to always be True. 

```suggestion
        new_listing = [{"name": path, "type": "file", "size": len(content)}]
        parent_path = self._parent(path)
        listings = self.dircache.get(parent_path, [])
        listings.append(new_listing)
        self.dircache[parent_path] = listings
```

cwognum (changes requested):
Nice optimization @lmtroper !

There is a more _official_ implementation that simplifies the code further. It's not very well documented, but what I did was to go through an [existing implementation](https://github.com/fsspec/filesystem_spec/blob/518984d494e48f007d8aa48de882ed7014da2574/fsspec/implementations/dbfs.py#L59-L96) to understand how this can be used. 

The good news - You reached a very similar solution! Great minds think alike! :wink:

polaris/hub/polarisfs.py
lmtroper: Yes, I had it slightly changed (originally `elif len(out1) > 1 or out`) as it was resulting in an error, but appears to be good without it now! :+1:

polaris/hub/polarisfs.py
```diff
@@ -55,6 +60,8 @@ def __init__(
         self.prefix = f"dataset/{dataset_owner}/{dataset_name}/"
         self.base_path = f"/storage/{self.prefix.rstrip('/')}"
 
+        self.refresh = False
```
cwognum, line 63: Why is this needed?

polaris/hub/polarisfs.py
```diff
@@ -87,16 +94,22 @@ def ls(
...
             timeout = self.default_timeout
 
+        cached_listings = self._ls_from_cache(path)
+        if cached_listings is not None:
+            return cached_listings if detail else [d["name"] for d in cached_listings]
+
+        cached_listings = self._ls_from_cache(self._parent(path))
+        if cached_listings is not None:
+            self.dircache[path] = []
+            return cached_listings if detail else [d["name"] for d in cached_listings]
```
cwognum, line 104: Why do we check ls for both the parent and the path? I feel like we should only do so for the path, as is e.g. done in [this existing implementation](https://github.com/fsspec/filesystem_spec/blob/518984d494e48f007d8aa48de882ed7014da2574/fsspec/implementations/dbfs.py#L72).

polaris/hub/polarisfs.py
```diff
@@ -212,3 +232,15 @@ def pipe_file(
...
+
+        new_listing = [{"name": path, "type": "file", "size": len(content)}]
+        try:
+            current_listings = self._ls_from_cache(self._parent(path))
+            new_listing.extend(current_listings)
+        except FileNotFoundError:
+            # self._ls_from_cache() raises FileNotFoundError when no listings
+            # are found within a path. We can then assume it's a new directory
+            # and new_listing can be used as is.
+            pass
```
cwognum, line 244: `extend` is not in-place, so this actually doesn't have an effect.
```suggestion
        new_listing = [{"name": path, "type": "file", "size": len(content)}]
        try:
            current_listings = self._ls_from_cache(self._parent(path))
            new_listing = current_listings.extend(new_listing)
        except FileNotFoundError:
            # self._ls_from_cache() raises FileNotFoundError when no listings
            # are found within a path. We can then assume it's a new directory
            # and new_listing can be used as is.
            pass
```

lmtroper: I believe `extend` does modify in place. Doing this way returns `None` for `new_listing`.

cwognum: You're absolute right. My bad!

cwognum (approved):
Implementation looks great! 

However, in your experience it has been slower, right?

I'm not sure why it would be... Maybe leave the PR open for now while we investigate?

PR #98, opened 2024-04-15
Added new visualizations and updated logic to be conditioned on data type.

zhu0619:
## Changelogs

- Adjust the curation logic for curation on categorical data (on data merging and activity shift detection )
- Additional distribution plot [detailed_distributions_plots](https://github.com/polaris-hub/polaris/blob/185d101482b5f827950b3f721082ff94c6be6f46/polaris/curation/viz_utils.py#L190C5-L190C13) 
- Updated the unittests.
---

_Checklist:_

---

### DIFF
--- a/polaris/curation/_data_curator.py
+++ b/polaris/curation/_data_curator.py
@@ -51,8 +51,7 @@ def _identify_stereoisomers_with_activity_cliff(
         if group_df.shape[0] > 1:
             if data_type in CONTINUOUS:
                 has_cliff = (
-                    group_df[f"{data_col}_zscore"].max()
-                    - group_df[f"{data_col}_zscore"].min()
+                    group_df[f"{data_col}_zscore"].max() - group_df[f"{data_col}_zscore"].min()
                 ) > threshold
             else:
                 has_cliff = len(group_df[data_col].dropna().unique()) > 1
@@ -70,7 +69,7 @@ def _process_stereoisomer_with_activity_cliff(
     mask_stereo_undefined_mols: bool = True,
     **kwargs,
 ):
-    """Identify and discard the stereoisomers which the stereocenter are not well-defined but show activity cliff.
+    """Identify the stereoisomers which the stereocenter are not well-defined but show activity cliff.
 
     Args:
         data: Parsed dataset
@@ -81,28 +80,22 @@ def _process_stereoisomer_with_activity_cliff(
     """
     for data_col, data_type in zip(data_cols, data_types):
         data_col_mask = (
-            [data_col, data_col[len(CLASS_PREFIX) :]]
-            if data_col.startswith(CLASS_PREFIX)
-            else data_col
+            [data_col, data_col[len(CLASS_PREFIX) :]] if data_col.startswith(CLASS_PREFIX) else data_col
         )
 
         mol_with_cliff = _identify_stereoisomers_with_activity_cliff(
             data=data, data_col=data_col, data_type=data_type, **kwargs
         )
-        data.loc[
-            data[NO_STEREO_UNIQUE_ID].isin(mol_with_cliff), f"{data_col}_stereo_cliff"
-        ] = True
+        data.loc[data[NO_STEREO_UNIQUE_ID].isin(mol_with_cliff), f"{data_col}_stereo_cliff"] = True
         if len(mol_with_cliff) > 0 and mask_stereo_undefined_mols:
-            mol_ids = data.query(
-                f"`{data_col}_stereo_cliff`==True & (`{UNDEF_ED}` | `{UNDEF_EZ}`) "
-            )[UNIQUE_ID].values
+            mol_ids = data.query(f"`{data_col}_stereo_cliff`==True & (`{UNDEF_ED}` | `{UNDEF_EZ}`) ")[
+                UNIQUE_ID
+            ].values
             data.loc[data[UNIQUE_ID].isin(mol_ids), data_col_mask] = None
     return data
 
 
-def _merge_duplicates(
-    data: pd.DataFrame, data_cols: List[str], merge_on=None
-) -> pd.DataFrame:
+def _merge_duplicates(data: pd.DataFrame, data_cols: List[str], merge_on=None) -> pd.DataFrame:
     """Merge molecules with multiple measurements.
        To be robust to the outliers, `median` is used to compute the average value cross all measurements.
 
@@ -123,16 +116,14 @@ def _merge_duplicates(
         df.loc[:, data_cols] = data_vals
         df_list.append(df)
     merged_df = pd.concat(df_list).sort_values(by=merge_on)
-    merged_df = merged_df.drop_duplicates(subset=merge_on, keep="first").reset_index(
-        drop=True
-    )
+    merged_df = merged_df.drop_duplicates(subset=merge_on, keep="first").reset_index(drop=True)
     return merged_df
 
 
 def _class_conversion(
     data: pd.DataFrame,
     data_cols: List[str],
-    conversion_params: List[dict],
+    conversion_params: Dict[str, dict],
     prefix: str = CLASS_PREFIX,
 ) -> pd.DataFrame:
     """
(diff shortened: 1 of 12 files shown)


polaris/curation/_data_curator.py
```diff
@@ -32,10 +36,11 @@ def _identify_stereoisomers_with_activity_cliff(
         mol_hash_ids_with_cliff: A list mol hash ids for the stereoisomer molecules which show activity cliff.
 
     """
-    data_type = type_of_target(data[data_col].dropna().values, input_name=data_col)
 
     if data_type is None:
         raise ValueError(f"The column {data_col} contains less than 2 unique values.")
+    if CLASS_PREFIX in data_col:
```
cwognum, line 42: For such global variables, I think we should add a Pydantic Settings class. Similar to [how we did this for the client](https://github.com/polaris-hub/polaris/blob/main/polaris/hub/settings.py). That way, a user could configure them through e.g. environment variables!

polaris/curation/_data_curator.py
```diff
@@ -173,7 +183,10 @@ def check_outliers(data, data_cols, method: str = "zscore", prefix="OUTLIER", **
         <polaris.curation.util.ZscoreOutlier>
 
     """
-    for data_col in data_cols:
+    for data_col, data_type in zip(data_cols, data_types):
+        if data_type not in CONTINUOUS:
+            logger.info(f"Outlier detection is omit for categorical data {data_col}.")
```
cwognum, line 188:
```suggestion
            logger.info(f"Outlier detection is not supported for categorical data {data_col}.")
```

polaris/curation/_data_curator.py
```diff
@@ -195,49 +208,71 @@ def check_outliers(data, data_cols, method: str = "zscore", prefix="OUTLIER", **
...
+    )
 
     # User should deal with scaling and normalization on their end
-    data = _merge_duplicates(
-        data=data,
-        data_cols=data_cols,
-        merge_on=[NO_STEREO_UNIQUE_ID] if ignore_stereo else [UNIQUE_ID],
-    )
-    # class conversion
+    # The merging should only be performed for continuous data
```
cwognum, line 250: Why only for continuous? Is it because we don't know how to merge a data point with conflicting classes?

polaris/curation/viz_utils.py
```diff
@@ -23,27 +33,32 @@ def visualize_distribution(data_cols: List[str], dataset: PandasDataFrame):
     for col in data_cols:
         outlier_col = f"OUTLIER_{col}"
         if outlier_col not in dataset.columns:
-            raise ValueError(
-                f"{outlier_col} is not found in the given dataset.\
-                             Please run <polaris.curation.utils.outlier_detection> first."
+            logger.warning(
+                f"{outlier_col} is not found in the given dataset. Please run <polaris.curation.utils.outlier_detection> first."
             )
```
cwognum, line 38: Raising an error here makes more sense to me, because the method otherwise does nothing. Why did you change it?

polaris/curation/viz_utils.py
```diff
@@ -52,23 +67,46 @@ def verify_stereoisomers(data_cols: List[str], dataset: PandasDataFrame):
...
+        legend_cols = []
     for col in data_cols:
         logger.info(f"Verify the stereo ismomers for readout `{col}`")
         cliff_col = f"CLASS_{col}_stereo_cliff"
         if cliff_col not in dataset:
-            raise ValueError(
-                "The stereo chemistry information is unavailable. \
-                             Please run the <polaris.curation.run_chemistry_curation> first."
-            )
+            cliff_col = f"{col}_stereo_cliff"
```
cwognum, line 76: Turn into a setting, e.g. `ACTIVITY_CLIFF_COLL_SUFFIX`? Should be configurable through a Pydantic `BaseSettings` class.

polaris/curation/viz_utils.py
cwognum: You also throw an error for the `cliff_coll`. I think we should be consistent!

polaris/curation/viz_utils.py
```diff
@@ -102,3 +140,200 @@ def check_undefined_stereocenters(dataset: PandasDataFrame):
...
+        size_col: used for setting the style of the resulting scatterplot
+        seed: int, random seed to use for reproducibility
+        umap_metric: similarity metric to use for constructing the UMAP representation
+        hue_names: List of grouping variable that will produce points with different colors for the same UMAP plotting.
+
+    Returns:
+        figs, figures for each kind of split specified in split_names
+    """
+    figs = plt.figure(num=3)
+    features = [dm.to_fp(mol) for mol in data[mol_col]]  # Convert molecules to fingerprint
```
cwognum, line 169: I would make `features` a parameter, such that people could also use something else than ECFP4. Maybe we can have one parameter `X` and one parameter `smiles` and if `X` is not set, `smiles` needs to be and we will use ECFP4.

polaris/curation/viz_utils.py
```diff
@@ -102,3 +140,200 @@ def check_undefined_stereocenters(dataset: PandasDataFrame):
...
+        mol_col: string column where we will look for a molecule to featurize
+        size_col: used for setting the style of the resulting scatterplot
+        seed: int, random seed to use for reproducibility
+        umap_metric: similarity metric to use for constructing the UMAP representation
+        hue_names: List of grouping variable that will produce points with different colors for the same UMAP plotting.
+
+    Returns:
+        figs, figures for each kind of split specified in split_names
+    """
+    figs = plt.figure(num=3)
```
cwognum, line 168: Why `num=3`? I think we should do something like:

```python
hue_cols = [None] if hue_cols is None else hue_cols

num_figs = len(hue_cols)
fig, axs = plt.subplots(nrows=1, ncols=num_figs, figsize=(fig_size * num_figs, fig_size))
axs = np.atleast_1d(axs)

for ax, hue_col in zip(axs, hue_cols): 
    sns.plot(..., ax=ax, hue=hue_col)
```

Where `fig_size` is also a parameter.

polaris/curation/viz_utils.py
```diff
@@ -102,3 +140,200 @@ def check_undefined_stereocenters(dataset: PandasDataFrame):
...
     fig = dm.to_image(to_plot.mol.tolist(), legends=legends)
     return fig
+
+
+def visualize_chemspace(
+    data: pd.DataFrame,
+    mol_col: str = "smiles",
+    size_col=None,
+    seed=1428,
+    umap_metric="jaccard",
```
cwognum, line 150: I would replace by `**umap_kwargs`. Other parameters are also very influential and user will want to play with them! See e.g. https://umap-learn.readthedocs.io/en/latest/parameters.html

polaris/curation/viz_utils.py
```diff
@@ -102,3 +140,200 @@ def check_undefined_stereocenters(dataset: PandasDataFrame):
...
+        fig = sns.scatterplot(
+            data=data,
+            x="UMAP_0",
+            y="UMAP_1",
+            style=size_col,
+            hue=hue,
+            alpha=0.7,
+            palette="colorblind",
+        )
+        fig.set_title(f"UMAP embedding {hue}")
```
cwognum, line 187: Would make this more informative:
```suggestion
        fig.set_title(f"UMAP embedding (colored by {hue})")
```

polaris/curation/viz_utils.py
```diff
@@ -102,3 +140,200 @@ def check_undefined_stereocenters(dataset: PandasDataFrame):
...
+    return figs
+
+
+def detailed_distributions_plots(
```
cwognum, line 191: This is a very nice function! Some bits of it seem reusable across multiple methods (e.g. for computing figure size). Not needed now, but seems we could extract some reusable code from this method to be used by other methods as well.

cwognum (changes requested):
Thanks, @zhu0619! I did a first pass. 

There seems to be a mismatch between the line length you have configured locally for Ruff and the (custom) line length the [Polaris `pyproject.toml` specifies](https://github.com/polaris-hub/polaris/blob/main/pyproject.toml#L129). Because of this, a lot of changes actually aren't changes and it's a bit hard to review. Could you make sure these changes are reverted? Once that's done, I can do another pass! 

As this is getting more complex, I feel strongly that we should move it to a separate package. Once we go live and more people start using the Polaris client, it would be harder to do so (e.g. we would have to start with deprecation warnings etc.), so I would prefer to do it soon. Shouldn't be too hard, given it's already quite decoupled from the rest of the code.

cwognum:
@zhu0619 Could this PR be closed? Or are you still comparing it with https://github.com/polaris-hub/auroris/pull/1 ?

PR #101, opened 2024-04-22
Metrics absolute average fold error

zhu0619:
## Changelogs

- Add metric Balanced average precision from sklearn 
- Add metric Absolute Average Fold Error implementation from https://pubs.acs.org/doi/10.1021/acs.chemrestox.3c00305

---

_Checklist:_

---

### DIFF
--- a/polaris/benchmark/_base.py
+++ b/polaris/benchmark/_base.py
@@ -446,17 +446,19 @@ class BenchmarkSpecification(BaseArtifactModel):
         for test_label, y_true_subset in y_true.items():
             # For every metric...
             for metric in self.metrics:
-                y_pred_eval = y_prob if metric.needs_probs else y_pred
-
                 if metric.is_multitask:
                     # Multi-task but with a metric across targets
-                    score = metric(y_true=y_true_subset, y_pred=y_pred_eval[test_label])
+                    score = metric(
+                        y_true=y_true_subset, y_pred=y_pred[test_label], y_prob=y_prob.get(test_label, None)
+                    )
                     scores.loc[len(scores)] = (test_label, "aggregated", metric, score)
                     continue
 
                 if not isinstance(y_true_subset, dict):
                     # Single task
-                    score = metric(y_true=y_true_subset, y_pred=y_pred_eval[test_label])
+                    score = metric(
+                        y_true=y_true_subset, y_pred=y_pred[test_label], y_prob=y_prob.get(test_label, None)
+                    )
                     scores.loc[len(scores)] = (
                         test_label,
                         self.target_cols[0],
@@ -472,7 +474,10 @@ class BenchmarkSpecification(BaseArtifactModel):
                     mask = ~np.isnan(y_true_target)
                     score = metric(
                         y_true=y_true_target[mask],
-                        y_pred=y_pred_eval[test_label][target_label][mask],
+                        y_pred=y_pred[test_label][target_label][mask],
+                        y_prob=y_prob[test_label][target_label][mask]
+                        if y_prob[test_label] is not None
+                        else None,
                     )
                     scores.loc[len(scores)] = (test_label, target_label, metric, score)
 
--- a/polaris/benchmark/_base.py
+++ b/polaris/benchmark/_base.py
@@ -433,7 +433,7 @@ class BenchmarkSpecification(BaseArtifactModel):
         if not isinstance(y_prob, dict) or all(k in self.target_cols for k in y_prob):
             y_prob = {"test": y_prob}
 
-        if any(k not in y_pred for k in test.keys()):
+        if any(k not in y_pred for k in test.keys()) and any(k not in y_prob for k in test.keys()):
             raise KeyError(
                 f"Missing keys for at least one of the test sets. Expecting: {sorted(test.keys())}"
             )
@@ -448,7 +448,7 @@ class BenchmarkSpecification(BaseArtifactModel):
                 if metric.is_multitask:
                     # Multi-task but with a metric across targets
                     score = metric(
-                        y_true=y_true_subset, y_pred=y_pred[test_label], y_prob=y_prob.get(test_label)
+                        y_true=y_true_subset, y_pred=y_pred.get(test_label), y_prob=y_prob.get(test_label)
                     )
                     scores.loc[len(scores)] = (test_label, "aggregated", metric, score)
                     continue
@@ -456,7 +456,7 @@ class BenchmarkSpecification(BaseArtifactModel):
                 if not isinstance(y_true_subset, dict):
                     # Single task
                     score = metric(
-                        y_true=y_true_subset, y_pred=y_pred[test_label], y_prob=y_prob.get(test_label, None)
+                        y_true=y_true_subset, y_pred=y_pred.get(test_label), y_prob=y_prob.get(test_label)
                     )
                     scores.loc[len(scores)] = (
                         test_label,
@@ -473,7 +473,9 @@ class BenchmarkSpecification(BaseArtifactModel):
                     mask = ~np.isnan(y_true_target)
                     score = metric(
                         y_true=y_true_target[mask],
-                        y_pred=y_pred[test_label][target_label][mask],
+                        y_pred=y_pred[test_label][target_label][mask]
+                        if y_pred[test_label] is not None
+                        else None,
                         y_prob=y_prob[test_label][target_label][mask]
                         if y_prob[test_label] is not None
                         else None,
--- a/polaris/benchmark/_base.py
+++ b/polaris/benchmark/_base.py
@@ -409,7 +409,6 @@ class BenchmarkSpecification(BaseArtifactModel):
                 If there are multiple test sets, the predictions should be further wrapped in a dictionary
                     with the test subset labels as keys.
             y_prob: The predicted probabilities for the test set, as NumPy arrays.
-                Currently only multiclass in singletask setting is supported.
 
         Returns:
             A `BenchmarkResults` object. This object can be directly submitted to the Polaris Hub.
--- a/polaris/benchmark/_base.py
+++ b/polaris/benchmark/_base.py
@@ -388,7 +388,7 @@ class BenchmarkSpecification(BaseArtifactModel):
 
         return train, test
 
-    def evaluate(self, y_pred: PredictionsType) -> BenchmarkResults:
+    def evaluate(self, y_pred: PredictionsType, y_prob: PredictionsType = None) -> BenchmarkResults:
         """Execute the evaluation protocol for the benchmark, given a set of predictions.
 
         info: What about `y_true`?
@@ -408,6 +408,8 @@ class BenchmarkSpecification(BaseArtifactModel):
                 If there are multiple targets, the predictions should be wrapped in a dictionary with the target labels as keys.
                 If there are multiple test sets, the predictions should be further wrapped in a dictionary
                     with the test subset labels as keys.
+            y_prob: The predicted probabilities for the test set, as NumPy arrays.
+                Currently only multiclass in singletask setting is supported.
 
         Returns:
             A `BenchmarkResults` object. This object can be directly submitted to the Polaris Hub.
@@ -429,6 +431,9 @@ class BenchmarkSpecification(BaseArtifactModel):
         if not isinstance(y_pred, dict) or all(k in self.target_cols for k in y_pred):
             y_pred = {"test": y_pred}
 
+        if not isinstance(y_prob, dict) or all(k in self.target_cols for k in y_prob):
+            y_prob = {"test": y_prob}
+
         if any(k not in y_pred for k in test.keys()):
             raise KeyError(
                 f"Missing keys for at least one of the test sets. Expecting: {sorted(test.keys())}"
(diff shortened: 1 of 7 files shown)


polaris/evaluate/_metric.py
```diff
@@ -30,6 +31,28 @@ def spearman(y_true: np.ndarray, y_pred: np.ndarray):
...
+
+    Parameters:
+    y_true : array-like of shape (n_samples,)
+        The true target values.
+    y_pred : array-like of shape (n_samples,)
+        The predicted target values.
+
+    Returns:
+    aafe : float
+        The Absolute Average Fold Error.
```
cwognum, line 46: Should be consistent with the docstring style in the rest of the repo, so `Args` instead of `Parameters` and no type annotations in the docstring.

polaris/evaluate/_metric.py
```diff
@@ -30,6 +31,28 @@ def spearman(y_true: np.ndarray, y_pred: np.ndarray):
     return stats.spearmanr(y_true, y_pred).statistic
 
 
+def absolute_average_fold_error(y_true: np.ndarray, y_pred: np.ndarray):
+    """
+    Calculate the Absolute Average Fold Error (AAFE) metric.
```
cwognum, line 36: Could you explain what this metric measures? A brief description + a citation would be enough!

polaris/evaluate/_metric.py
```diff
@@ -75,7 +99,9 @@ class Metric(Enum):
     pr_auc = MetricInfo(fn=average_precision_score, direction="max")
     mcc = MetricInfo(fn=matthews_corrcoef, direction="max")
     cohen_kappa = MetricInfo(fn=cohen_kappa_score, direction="max")
+
     # TODO: adding metrics for multiclass tasks
+    ba = MetricInfo(fn=balanced_accuracy_score, direction="max")
```
cwognum, line 104: For clarity, I would use:
```suggestion
    balanced_accuracy = MetricInfo(fn=balanced_accuracy_score, direction="max")
```

polaris/evaluate/_metric.py
```diff
@@ -65,6 +88,7 @@ class Metric(Enum):
     pearsonr = MetricInfo(fn=pearsonr, direction="max")
     spearmanr = MetricInfo(fn=spearman, direction="max")
     explained_var = MetricInfo(fn=explained_variance_score, direction="max")
+    aafe = MetricInfo(fn=absolute_average_fold_error, direction="max")
```
cwognum, line 91: For clarity, I would use:
```suggestion
    absolute_average_fold_error = MetricInfo(fn=absolute_average_fold_error, direction="max")
```

polaris/evaluate/_metric.py
cwognum: Also, by placing this under the `TODO` comment, it seems to me this is a multi-task metric? Is that the case? If so, the `MetricInfo.is_multitask` flag should be set and I would update the comment. If not, I would move it above the comment.

cwognum (changes requested):
Thanks, @zhu0619 

For this to be merged, there needs to be a similar PR on the Hub side as well that adds support for these new metrics. Is this something you would feel comfortable doing, @zhu0619 ? If not, I can look into it!

cwognum:
Ah I see you created https://github.com/polaris-hub/polaris-hub/issues/317.

polaris/evaluate/_metric.py
zhu0619: It's multiclass not multitask.

cwognum (changes requested):
Could we add some test cases? Especially for the multi-class one, this sounds like something that could fail.

zhu0619:
> Thanks, @zhu0619
> 
> For this to be merged, there needs to be a similar PR on the Hub side as well that adds support for these new metrics. Is this something you would feel comfortable doing, @zhu0619 ? If not, I can look into it!

I will do it.

polaris/evaluate/_metric.py
```diff
@@ -92,18 +92,21 @@ class Metric(Enum):
...
     accuracy = MetricInfo(fn=accuracy_score, direction="max")
+    balanced_accuracy = MetricInfo(fn=balanced_accuracy_score, direction="max")
+    mcc = MetricInfo(fn=matthews_corrcoef, direction="max")
+    cohen_kappa = MetricInfo(fn=cohen_kappa_score, direction="max")
+
+    # binary only
     f1 = MetricInfo(fn=f1_score, kwargs={"average": "binary"}, direction="max")
-    f1_macro = MetricInfo(fn=f1_score, kwargs={"average": "macro"}, direction="max")
-    f1_micro = MetricInfo(fn=f1_score, kwargs={"average": "micro"}, direction="max")
+    # note: At the moment, multi-dimension inputs for classification are not supported
```
zhu0619, line 103: @cwognum We can extend input attributes with `y_prob` so user can provide both predicted classes and probabilities for different multiclass metrics.

zhu0619: It doesn't have to be addressed in this PR though.

polaris/evaluate/_metric.py
```diff
@@ -30,6 +31,30 @@ def spearman(y_true: np.ndarray, y_pred: np.ndarray):
...
+
+    Args:
+        y_true : array-like of shape (n_samples,)
+            The true target values.
+        y_pred : array-like of shape (n_samples,)
+            The predicted target values.
+
+    Returns:
+        aafe : float
+            The Absolute Average Fold Error.
```
cwognum, line 48: We don't annotate the types in the docstring because we use type-hints already!  

```suggestion
    Args:
        y_true: The true target values of shape (n_samples,)
        y_pred: The predicted target values of shape (n_samples,).

    Returns:
        aafe: The Absolute Average Fold Error.
```

polaris/evaluate/_metric.py
```diff
@@ -30,6 +31,30 @@ def spearman(y_true: np.ndarray, y_pred: np.ndarray):
     return stats.spearmanr(y_true, y_pred).statistic
 
 
+def absolute_average_fold_error(y_true: np.ndarray, y_pred: np.ndarray):
```
cwognum, line 34: This is actually needed, because the docs will fail otherwise. 

```suggestion
def absolute_average_fold_error(y_true: np.ndarray, y_pred: np.ndarray) -> float:
```

polaris/evaluate/_metric.py
```diff
@@ -30,6 +31,30 @@ def spearman(y_true: np.ndarray, y_pred: np.ndarray):
...
+    """
+    Calculate the Absolute Average Fold Error (AAFE) metric.
+    It measures the fold change between predicted values and observed values.
+    The implementation is based on https://pubs.acs.org/doi/10.1021/acs.chemrestox.3c00305.
```
cwognum, line 38: Looks nice in the rendered docs! 

```suggestion
    The implementation is based on [this paper](https://pubs.acs.org/doi/10.1021/acs.chemrestox.3c00305).
```

polaris/evaluate/_metric.py
cwognum: That's a good point. We probably should and add a `needs_probs` attribute to the `MetricInfo` class?

tests/test_evaluate.py
cwognum, line 1: Could we also add a unit test to simply check if calling `absolute_average_fold_error` works as expected, since it's a custom implementation? Can be simple! Just some hard-coded values to compare against.

polaris/evaluate/_metric.py
cwognum, line 34: Let's add this to our API docs. You can add the following to `docs/api/evaluation.md`:

```

::: polaris.evaluate._metric.absolute_average_fold_error

---
```

cwognum (changes requested):
Hey @zhu0619 , this is probably a bigger change than you anticipated by now! Sorry. Think we're almost there! Let me know if you need help. Happy to jump in.

polaris/evaluate/_metric.py
```diff
@@ -65,17 +90,23 @@ class Metric(Enum):
     pearsonr = MetricInfo(fn=pearsonr, direction="max")
     spearmanr = MetricInfo(fn=spearman, direction="max")
     explained_var = MetricInfo(fn=explained_variance_score, direction="max")
+    absolute_average_fold_error = MetricInfo(fn=absolute_average_fold_error, direction="max")
```
cwognum, line 93: Based on our discussion in person, just adding this comment so we don't forget: We should check the `direction`!

cwognum: I believe the target is `1.0`, with a range of `[0, ∞)`.

cwognum: Actually, if targets are 0.0 there could be a division by zero. We should probably add a small delta to prevent that? Or check for it?

polaris/evaluate/_metric.py
```diff
@@ -45,6 +70,7 @@ class MetricInfo(BaseModel):
     is_multitask: bool = False
     kwargs: dict = Field(default_factory=dict)
     direction: DirectionType
+    needs_probs: bool = False
```
cwognum, line 73: In line with scikit-learn, I would like to use named arguments that are specific to what's expected, e.g. `def metric_1(y_true, y_pred)` vs `def metric_2(y_true, y_prob)`. I find that to be more clear and like to be consistent with Scikit-learn.

It seems scikit-learn uses `y_pred`, `y_prob` and `y_score`. What would you think of: 
```suggestion
    y_type: Literal["y_pred", "y_prob", "y_score"] = "y_pred"
```
?

 Then in `Metric.score()` we can do something like: 
```python
def score(
    self, 
    y_true: np.ndarray, 
    y_pred: Optional[np.ndarray] = None, 
    y_score: Optional[np.ndarray] = None, 
    y_prob: Optional[np.ndarray] = None
) -> float:
        
    if y_pred is None and y_score is None and y_prob is None: 
        raise ValueError(...)

    if self.y_type == "y_pred":
        pred = y_pred
    elif self.y_type == "y_score":
        pred = y_score
    else:
        pred = y_prob

    kwargs = {"y_true": y_true, self.y_type: pred}
    return self.fn(**kwargs)
```

I think it's more robust, but night complicate the API for the user, because you would probably have to propagate such changes to the API of the benchmark too.

What do you think?

tests/test_evaluate.py
```diff
@@ -73,17 +73,33 @@ def test_metrics_singletask_clf(
...
+    result = test_single_task_benchmark_multi_clf.evaluate(predictions, probablities)
+    for metric in test_single_task_benchmark_multi_clf.metrics:
+        assert metric in result.results.Metric.tolist()
+
+
 def test_metrics_multitask_clf(tmpdir: str, test_multi_task_benchmark_clf: MultiTaskBenchmarkSpecification):
     train, test = test_multi_task_benchmark_clf.get_train_test_split()
     predictions = {
         target_col: np.random.randint(2, size=test.inputs.shape[0]) for target_col in train.target_cols
     }
```
cwognum, line 98: This can be removed, right?

cwognum (changes requested):
Hey @zhu0619, I had one big question I would be curious to hear your thoughts on. Other than that, just some small comments.

polaris/benchmark/_base.py
```diff
@@ -408,6 +408,8 @@ def evaluate(self, y_pred: PredictionsType) -> BenchmarkResults:
                 If there are multiple targets, the predictions should be wrapped in a dictionary with the target labels as keys.
                 If there are multiple test sets, the predictions should be further wrapped in a dictionary
                     with the test subset labels as keys.
+            y_prob: The predicted probabilities for the test set, as NumPy arrays.
+                Currently only multiclass in singletask setting is supported.
```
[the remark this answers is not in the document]
zhu0619: I don't think we have metrics which are able to handle multitask multiclass with mutli dimentsional probablities ?

polaris/evaluate/_metric.py
zhu0619: It would work. But isn't it additional work that user who defines metrics have to figure out the "y_type" for each metric?

zhu0619: Maybe use `metric_args = inspect.signature(self.fn).parameters` to get the `y_type` instead.

zhu0619: > It would work. But isn't it additional work that user who defines metrics have to figure out the "y_type" for each metric?

oh. I see. It's defined in the MetricInfo.

zhu0619: The sklearn metrics used various argument names for `y_pred`. https://scikit-learn.org/stable/modules/model_evaluation.html#classification-metrics
Which is fine because they will be defined internally with MetricInfo. 
I am not sure whether to differenciate "y_score" and "y_prob". To me, one of them would be enough. 
In Benchmark.evaluate(), I think we can simplify them to either "y_score" or "y_prob". Maybe "y_score" is more universal.

polaris/benchmark/_base.py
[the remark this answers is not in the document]
cwognum: Ah okay, so it's a limitation of the metrics? If so, I would remove the comment or add a `TODO` to the `Metric` class. I thought it was a limitation of the `BenchmarkSpecification.evaluate()` method!

polaris/evaluate/_metric.py
cwognum: > In Benchmark.evaluate(), I think we can simplify them to either "y_score" or "y_prob".

Yes, agreed, but let's go with `y_prob` as you have it now.

From the scikit-learn docs on `y_score` (see also [here](https://scikit-learn.org/stable/glossary.html#term-decision_function)): 
> These can be either probability estimates or non-thresholded decision values

It seems that if we expect `y_score`, we can always provide `y_prob`, but not the other way around. I thus think `y_prob` will be the better choice and I also think its naming is better understood.

polaris/benchmark/_base.py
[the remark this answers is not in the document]
zhu0619: Yeah. It's the limitation of the metrics. I will add TODO.

tests/test_evaluate.py
zhu0619: The evaluation will take both 'predictions' and 'probabilities' as input for all the defined metrics.

polaris/benchmark/_base.py
```diff
@@ -443,13 +447,17 @@ def evaluate(self, y_pred: PredictionsType) -> BenchmarkResults:
             for metric in self.metrics:
                 if metric.is_multitask:
                     # Multi-task but with a metric across targets
-                    score = metric(y_true=y_true_subset, y_pred=y_pred[test_label])
+                    score = metric(
+                        y_true=y_true_subset, y_pred=y_pred[test_label], y_prob=y_prob.get(test_label, None)
```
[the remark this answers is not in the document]
cwognum: I am not sure if we're saying the same thing. I meant to say that `{}.get()` will return `None`. No need to specify the second (i.e. `default`) parameter.

cwognum (approved):
Thanks, @zhu0619 ! Looks great!

PR #102, opened 2024-05-08
Include a checksum for Zarr archives in the Dataset checksum

cwognum:
## Changelogs

- Integrated a checksum for Zarr archives in the Polaris Dataset.
- Made it possible to lazily compute the checksum for an entire dataset or benchmark. 
- Verify the checksum of each Zarr file on download.
- Removed broken `ls` caching from the `PolarisFileSystem`. 
  - This slows down up- and downloads, but we are working on a new JWT-based system anyways.
- Added debug logs and set logging level by default to `INFO`. 
- In accordance with the Apache 2 license, added a separate NOTICE and LICENSE file.

---

_Checklist:_

---

### How it works

The Polaris implementation is heavily based on the [`zarr-checksum` package](https://github.com/dandi/zarr_checksum).  This algorithm computes a checksum per file in the Zarr archive and then works its way up the file system hierarchy in a deterministic order to compute a checksum for the entire archive.

> [!NOTE]
> This implies the checksum can change, even if the content doesn't. For example because you rechunk the dataset. This fits the data integrity use case of Polaris. 

The checksum per file is saved in a _manifest_, which is saved in the Hub DB and used by the Hub to verify the integrity of the archive on upload. 

Data integrity is verified in two use-cases: 
- When downloading a _single chunk_ from the Zarr archive, the md5sum of that single chunk is retrieved from the headers and verified. 
- When downloading the _entire Zarr archive_, the checksum for the entire dataset is computed and verified.

# Links
- Related PR for Hub: https://github.com/polaris-hub/polaris-hub/pull/377


polaris/dataset/_dataset.py
```diff
@@ -143,7 +143,7 @@ def _validate_model(cls, m: "Dataset"):
         # Verify the checksum
         # NOTE (cwognum): Is it still reasonable to always verify this as the dataset size grows?
         actual = m.md5sum
-        expected = cls._compute_checksum(m.table)
+        expected = cls._compute_checksum(m.table, m.zarr_root_path)
```
cwognum, line 146: We now always compute the checksum. I don't think that makes sense. 

We should probably only do this on up- and download? 

For example, a TB-sized datasets could be stored on the Hub. Do you really want to compute the checksum when you're loading from the hub?

cwognum: I meant to say that the mechanism I'm adding with this PR is for an _entire_ Zarr archive. However, if you don't download the dataset in its entirety but rather keep it stored on the Hub and only load the Zarr chunks you need, it is a waste to compute the checksum for the entire archive.

cwognum: > Do we have any estimation of how many chunks a user would typically need to call in at once?

This depends on how the dataset is chunked and the data access pattern. I don't think we need to worry about this too, too much though. This is abstracted away through Zarr and it seems to me that we just need to make sure that as we download a `chunk` from the Hub through the [`PolarisFileSystem`](https://github.com/polaris-hub/polaris/blob/main/polaris/hub/polarisfs.py), that we verify the hash.

polaris/dataset/zarr/_checksum.py
```diff
@@ -0,0 +1,314 @@
...
+    # Make sure the path exists and is a Zarr archive
+    zarr.open_group(zarr_root_path, mode="r")
+
+    # Generate the checksum
+    tree = ZarrChecksumTree()
+
+    # Find all files in the root
+    leaves = fs.find(zarr_root_path, detail=True)
+
+    for file in tqdm(leaves.values(), desc="Finding all files in the Zarr archive"):
```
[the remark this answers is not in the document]
cwognum: With up- and downloads, it is a little harder because this is currently handled by Zarr (see [`copy_all`](https://zarr.readthedocs.io/en/stable/api/convenience.html#zarr.convenience.copy_all) and [`copy_store`](https://zarr.readthedocs.io/en/stable/api/convenience.html#zarr.convenience.copy_store)). These functions expose the `log` argument to log updates on the upload, but I am not a big fan of the formatting. Might nevertheless be a good idea to start using that for a better UX.

cwognum:
Thanks, @Andrewq11. 

To answer your questions: 

> Might be missing something here but why would we need to compute the checksum on the hub? 

> why not just then shift to individual chunk checksums only

Because we want to know if the entire dataset was uploaded / downloaded successfully. If not, a Dataset should for example not be made public on the Hub. This means we don't only need to verify the data integrity of every individual chunk, but we also need to check if those chunks are organized as expected within the Zarr archive.

> a faster hashing function might be needed depending on if the number of chunks to download becomes exceedingly large.

One issue with changing hashing functions, is that we need to make sure to Hub supports the same hashing function. One argument in favor of md5 is that I believe the md5 hash to already be returned "for free" in the headers when we fetch a file from the R2 bucket.

Andrewq11:
> Because we want to know if the entire dataset was uploaded / downloaded successfully. If not, a Dataset should for example not be made public on the Hub. 

So we would need to download the zarr archive to the hub directly? Is this the correct flow?

Client computes the checksum for the zarr archive -> uploads it to R2 ->  the Hub downloads the dataset -> the Hub calculates and confirms checksums match.

If so, I could see this being an issue on the Hub when datasets get sufficiently large. I think we would likely need a dedicated service with enough memory and disk space available to complete the download and perform the check. 

> This means we don't only need to verify the data integrity of every individual chunk, but we also need to check if those chunks are organized as expected within the Zarr archive.

Makes sense, thank you. 

> One issue with changing hashing functions, is that we need to make sure to Hub supports the same hashing function. One argument in favor of md5 is that I believe the md5 hash to already be returned "for free" in the headers when we fetch a file from the R2 bucket.

That would be very handy if so. This seems like something that could be useful to benchmark in the future and see if we could get any significant gains using different hashing functions.

polaris/dataset/_dataset.py
```diff
@@ -192,6 +193,11 @@ def _compute_checksum(table):
         table_hash = pd.util.hash_pandas_object(df, index=False).sum()
         hash_fn.update(table_hash)
 
+        # If the Zarr archive exists, we hash its contents too.
+        if zarr_root_path is not None:
+            zarr_hash = compute_zarr_checksum(zarr_root_path)
+            hash_fn.update(zarr_hash.encode())
```
jstlaurent, line 199: **Comment**
@cwognum and I had a live discussion about this yesterday. See my general review comments.

jstlaurent (commented):
Aside from a few suggestions here, Cas and I did some brainstorming about this feature. 

We have two use cases for this feature:

1. Checking Zarr chunk integrity on download in the client
2. Checking upload integrity and completion in the Hub

In order to support both, we came to the conclusion we need two things:

**Persist the whole Zarr checksum tree in the Hub**

We can do this as a JSON structure in the Dataset table. This will allow us to check, on the Hub side, that every part of the Zarr archive has been uploaded, and that every part is correct.

There might be some concern about how big that JSON structure would become for the largest dataset (RXRX3), but we'll have to see how that evolves.

**Use checksums on individual files, stored as metadata**

Unfortunately, R2 does not support the automated checksum generation for downloaded files that S3 does. So we'll need to work around that by computing each file's checksum prior to upload, setting that as metadata, and checking it back on download.

cwognum:
Just came across this: Instead of manually saving the checksum to R2, we could also a codec like this one: https://numcodecs.readthedocs.io/en/latest/checksum32.html

Although this lets us verify the integrity of individual chunks, I don't think it covers our use case. We would also need to be able to lookup the checksums on the Hub to verify the entire dataset was uploaded correctly.

jstlaurent:
@cwognum : I'm seeing new activity here. Should this be part of the all-Zarr dataset we had talked about?

cwognum:
> @cwognum : I'm seeing new activity here. Should this be part of the all-Zarr dataset we had talked about?

I'm not sure I understand what you mean. We already "support" Zarr datasets, so we will need this checksum anyways, independent of when we plan to pursue the all-Zarr dataset. Or am I missing something?

polaris/dataset/_dataset.py
```diff
@@ -241,14 +293,17 @@ def zarr_root(self):
...
-        saved_remote = saved_on_hub or not fs.is_local_path(self.zarr_root_path)
 
-        if saved_remote:
-            logger.warning(
-                f"You're loading data from a remote location. "
-                f"To speed up this process, consider caching the dataset first "
-                f"using {self.__class__.__name__}.cache()"
-            )
+        if self._warn_about_remote_zarr:
+            saved_remote = saved_on_hub or not dmfs.is_local_path(self.zarr_root_path)
```
jstlaurent, line 298: **Suggestion**
Going to also suggest `pathlib`here too:
```suggestion
            saved_remote = saved_on_hub or not Path(self.zarr_root_path).exists()
```

polaris/dataset/_dataset.py
```diff
@@ -369,7 +424,11 @@ def from_json(cls, path: str):
         data.pop("cache_dir", None)
         return cls.model_validate(data)
 
-    def to_json(self, destination: str) -> str:
+    def to_json(
+        self,
+        destination: str,
+        if_exists: ZarrConflictResolution = "replace",
```
jstlaurent, line 430: **Comment**
Nice typing. 😄

polaris/dataset/_dataset.py
```diff
@@ -384,44 +443,48 @@ def to_json(self, destination: str) -> str:
...
         # Lu: Avoid serilizing and sending None to hub app.
         serialized = self.model_dump(exclude={"cache_dir"}, exclude_none=True)
         serialized["table"] = table_path
 
         # Copy over Zarr data to the destination
-        if self.zarr_root is not None:
-            dest = zarr.open(zarr_archive, "w")
-            zarr.copy_all(source=self.zarr_root, dest=dest)
+        if self.uses_zarr:
+            with tmp_attribute_change(self, "_warn_about_remote_zarr", False):
```
jstlaurent, line 463: **Question**
Is that context manager needed, if `_warn_about_remote_zarr` will be set to `False`after the first warning? Seems like more complexity for little gain.

polaris/dataset/_dataset.py
```diff
@@ -432,10 +495,13 @@ def cache(self, cache_dir: Optional[str] = None) -> str:
...
         self.to_json(self.cache_dir)
 
-        if self.zarr_root_path is not None:
-            self.zarr_root_path = fs.join(self.cache_dir, "data.zarr")
+        if self.uses_zarr:
+            self.zarr_root_path = dmfs.join(self.cache_dir, "data.zarr")
             self._zarr_root = None
 
+        if verify_checksum and self._md5sum is not None:
+            self.verify_checksum(md5sum=self._md5sum)
```
jstlaurent, line 503: **Question**
This feels a little weird. Wouldn't this do the same thing?
```suggestion
        if verify_checksum:
            self.verify_checksum()
```

polaris/hub/client.py
```diff
@@ -174,6 +178,20 @@ def _normalize_owner(
...
         return artifact_owner if isinstance(artifact_owner, HubOwner) else HubOwner(slug=artifact_owner)
 
+    @staticmethod
+    def _normalize_verify_checksum(
+        verify_checksum: Optional[bool],
+        dataset: Dataset,
+    ):
+        if verify_checksum is not None:
+            return verify_checksum
+        return dataset._md5sum is not None and not dataset.uses_zarr
```
jstlaurent, line 188: **Comment**
I'm not liking direct access of private properties on the dataset object. If the intent is to check that the md5sum exists, then I'd suggest adding another property to the dataset class to expose this.

polaris/hub/client.py
```diff
@@ -326,13 +344,16 @@ def list_datasets(self, limit: int = 100, offset: int = 0) -> list[str]:
...
+        self, owner: Union[str, HubOwner], name: str, verify_checksum: Optional[bool] = None
+    ) -> Dataset:
         """Load a dataset from the Polaris Hub.
 
         Args:
             owner: The owner of the dataset. Can be either a user or organization from the Polaris Hub.
             name: The name of the dataset.
-            verify_checksum: Whether to use the checksum to verify the integrity of the dataset.
+            verify_checksum: Whether to use the checksum to verify the integrity of the dataset. If None,
+                will infer a practical default based on the dataset's storage location.
```
jstlaurent, line 356: **Comment**
This feels like overloading a boolean arg to have more than two possible values. Also, it makes it unclear to me what "a practical default" would be. If I don't specify anything, will the checksum be verified?

Maybe something like an `Enum` or a `Literal` type with clearer options on what will happen?

polaris/loader/load.py
```diff
@@ -37,12 +38,20 @@ def load_dataset(path: str, verify_checksum: bool = True) -> Dataset:
...
+    # Load from local file
     if extension == "json":
-        return Dataset.from_json(path)
-    return create_dataset_from_file(path)
+        dataset = Dataset.from_json(path)
+    else:
+        dataset = create_dataset_from_file(path)
 
+    # Verify checksum if requested
+    if PolarisHubClient._normalize_verify_checksum(verify_checksum, dataset):
```
jstlaurent, line 48: **Comment**
Accessing private static methods always feels weird to me.

polaris/hub/polarisfs.py
```diff
@@ -202,14 +223,7 @@ def pipe_file(
...
 
-        sha256_hash = hashlib.sha256(content).hexdigest()
-
-        headers = {
-            "Content-Type": "application/octet-stream",
-            "x-amz-content-sha256": sha256_hash,
-            "x-amz-date": datetime.now(timezone.utc).strftime("%Y%m%dT%H%M%SZ"),
-            **hub_response_body["headers"],
-        }
+        headers = {"Content-Type": "application/octet-stream", **hub_response_body["headers"]}
```
jstlaurent, line 226: **Suggestion**
Cloudflare R2 [supports](https://developers.cloudflare.com/r2/api/s3/api/#implemented-object-level-operations) the `Content-MD5` HTTP header on PUT operations. You could add it here to ensure integrity on the fragment being uploaded here.

jstlaurent (approved):
Added some questions and comments, but looks good overall to me. Nice work. 😄

polaris/benchmark/_base.py
```diff
@@ -310,6 +294,36 @@ def _compute_checksum(dataset, target_cols, input_cols, split, metrics):
...
+            raise PolarisChecksumError(
+                f"The specified checksum {md5sum} does not match the computed checksum {self.md5sum}"
+            )
+
+    @computed_field
+    @property
+    def md5sum(self) -> Optional[str]:
+        """Lazily compute the checksum once needed."""
+        if self._md5sum is None:
+            self._md5sum = self._compute_checksum(
```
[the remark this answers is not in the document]
cwognum: Good point! This was a static method before because it was called from within a class method (i.e., a Pydantic `@model_validator`). That's no longer the case!

polaris/dataset/_dataset.py
```diff
@@ -99,7 +109,7 @@ def _validate_table(cls, v):
         """
         # Load from path if not a dataframe
         if not isinstance(v, pd.DataFrame):
-            if not fs.is_file(v) or fs.get_extension(v) not in _SUPPORTED_TABLE_EXTENSIONS:
+            if not dmfs.is_file(v) or dmfs.get_extension(v) not in _SUPPORTED_TABLE_EXTENSIONS:
```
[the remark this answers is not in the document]
cwognum: Actually, for this specific case: Will that work? People may specify a table that is stored in e.g. GCP or AWS, in which case `filepath.is_file()` would return `False` and an error would be thrown, right?

polaris/dataset/_dataset.py
cwognum: It's not exactly the same thing, because it would throw an exception if `self._md5sum` would be `None`. On second thought, however, I think that is better behavior? If you specify `verify_checksum`, it should fail if there is no checksum to verify.

polaris/dataset/_dataset.py
```diff
@@ -191,8 +186,8 @@ def _compute_checksum(
 
         # If the Zarr archive exists, we hash its contents too.
         zarr_md5sum_manifest = None
-        if zarr_root_path is not None:
-            zarr_hash, zarr_md5sum_manifest = compute_zarr_checksum(zarr_root_path)
+        if self.zarr_root_path is not None:
```
kiramclean, line 189: Can this use the `uses_zarr` property instead?

polaris/benchmark/_base.py
```diff
@@ -310,6 +294,47 @@ def _compute_checksum(dataset, target_cols, input_cols, split, metrics):
...
+            md5sum = self._md5sum
+        if md5sum is None:
+            logger.warning(
+                "No checksum to verify against. Specify either the md5sum parameter or "
+                "store the checksum in the benchmark.md5sum attribute. Skipping!"
+            )
+            return
+
+        # Temporarily reset
+        # Calling self.md5sum will recompute the checksum and set it again
```
kiramclean, line 311: It seems like this might the kind of thing that becomes confusing in a few months once this isn't so fresh anymore, it's a little hard on the head to follow what `md5sum` is set to and what's being computed or set -- could we use `self._compute_checksum()` here explicitly to make it obvious what's going on?

kiramclean: I think this could also potentially leave the benchmark with no `md5sum` set? Is that intentional? I.e. in the case where the `_md5sum` gets nulled out but the check fails, now the old `_md5sum` value will be lost and the benchmark's `_md5sum` will remain `None`, so in the case where the check fails `verify_checksum` is effectively resetting the `_md5sum` to `None`, which seems like it might be a bit surprising.

polaris/benchmark/_base.py
```diff
@@ -310,6 +294,47 @@ def _compute_checksum(dataset, target_cols, input_cols, split, metrics):
...
+        self._md5sum = None
+        if self.md5sum != md5sum:
+            raise PolarisChecksumError(
+                f"The specified checksum {md5sum} does not match the computed checksum {self.md5sum}"
+            )
+
+    @computed_field
+    @property
+    def md5sum(self) -> Optional[str]:
+        """Lazily compute the checksum once needed."""
```
kiramclean, line 321: Is it possible that this might be very slow for large datasets? Should we log something to indicate that we're computing (and caching) the checksum to any confusion around why it might take a surprisingly long time to access the `md5sum`?

polaris/dataset/_dataset.py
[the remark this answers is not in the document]
jstlaurent: Ah, I guess you're right. In that case, we need `fsspec`.

polaris/dataset/_dataset.py
```diff
@@ -140,24 +147,12 @@ def _validate_model(cls, m: "Dataset"):
...
-            raise PolarisChecksumError(
-                "The dataset md5sum does not match what was specified in the meta-data. "
-                f"{actual} != {expected}"
-            )
-
         # Set the default cache dir if none and make sure it exists
         if m.cache_dir is None:
-            m.cache_dir = fs.join(DEFAULT_CACHE_DIR, _CACHE_SUBDIR, m.name, m.md5sum)
-        fs.mkdir(m.cache_dir, exist_ok=True)
+            dataset_id = m._md5sum if m.has_md5sum else str(uuid.uuid4())
```
kiramclean, line 152: Is there a public getter for `_md5sum`?

polaris/dataset/_dataset.py
```diff
@@ -170,30 +165,90 @@ def _serialize_adapters(self, value: List[Adapter]):
...
+    @computed_field
+    @property
+    def md5sum(self) -> str:
+        """Lazily compute the checksum once needed."""
+        if not self.has_md5sum:
+            self._md5sum, self._zarr_md5sum_manifest = self._compute_checksum()
+        return self._md5sum
+
+    @md5sum.setter
+    def md5sum(self, value: str):
```
kiramclean, line 234: Can any of these implementations be consolidated into some higher-up class or util module and shared? Just the ones that are identical across artifacts, to avoid them getting out of sync in the future.

polaris/dataset/_subset.py
```diff
@@ -76,6 +76,10 @@ def __init__(
 
         self._adapters = adapters
         self._featurization_fn = featurization_fn
+
+        # NOTE (cwognum): Note to future self. As we're starting to think about competition-style benchmarks,
+        #  we will likely split up datasets. In that case, this default iloc_to_loc mapping won't work.
+        #  By that time, we should probably be able to overwrite this mapping.
```
kiramclean, line 82: This is a good point, also relevant is the ongoing discussion about making all datasets Zarr-only. There have been a lot of discussions over the years about supporting a dataframe-like interface for Zarr files (e.g. [here](https://github.com/zarr-developers/numcodecs/issues/452), [here](https://github.com/zarr-developers/community/issues/31), and [here](https://github.com/zarr-developers/zarr-python/pull/84)), but as it stands the Zarr team has no intention of supporting dataframe-like access to Zarr files. It will remain array-like only for the time being, which may require re-thinking this kind of indexed access (or maybe not, but just something to investigate more and keep in mind!)

polaris/dataset/zarr/_checksum.py
```diff
@@ -0,0 +1,410 @@
+"""
```
kiramclean, line 1: Is the intention to keep this up to date with the `zarr_checksum` lib as it evolves and receives updates? Or do we really want to fork it and freeze it at this point in time? If we want to keep up with any updates in the original source lib this will make it very difficult (or impossible). But maybe that doesn't matter. If it does, though, another option could be to fork the `zarr_checksum` library and use our fork as dependency, then try to organize changes we make in a way that will minimize merge conflicts with any upstream changes, so we can pull them in as desired.

polaris/hub/client.py
```diff
@@ -347,10 +371,14 @@ def get_benchmark(
...
 
-        if not verify_checksum:
-            response.pop("md5Sum", None)
+        benchmark = benchmark_cls(**response)
 
-        return benchmark_cls(**response)
+        if should_verify_checksum(verify_checksum, benchmark.dataset):
+            benchmark.verify_checksum()
+        else:
+            benchmark.md5sum = response["md5Sum"]
```
kiramclean, line 379: Is there ever a case where we get a `md5Sum` in the response but don't want to use it as the artifact's checksum? Just wondering if this could possibly be simplified to assume `benchmark.md5sum` is either `response["md5Sum"]` or the result of calling `verify_checksum()`

polaris/utils/misc.py
```diff
@@ -16,3 +19,15 @@ def sluggify(sluggable: SlugCompatibleStringType):
...
     """
     return sluggable.lower().replace("_", "-")
+
+
+def should_verify_checksum(strategy: ChecksumStrategy, dataset: "Dataset") -> bool:
+    """
+    Determines whether a checksum should be verified.
+    """
+    if strategy == "ignore":
+        return False
```
kiramclean, line 29: Just out of curiosity: What is the scenario where one wants to ignore checksum verification?

polaris/dataset/_dataset.py
cwognum: Absolutely! Good catch!

polaris/dataset/_dataset.py
cwognum: Given that we're accessing the private property from within the same class, I am okay with this access!

polaris/utils/misc.py
cwognum: Good question! It was added because I once - in my infinite wisdom - decided to change how the checksum was computed in the code and I migrated the DB accordingly. This led to a scenario in which any older client could not load any of the datasets or benchmarks anymore, despite the data model being the same...

I will keep it in there for now, mostly as a fail-safe, but maybe I'm being paranoid here!

polaris/dataset/zarr/_checksum.py
cwognum: We have already deviated from the original implementation and if anything, I expect us to deviate more in the future. It also doesn't seem like the maintainers of the upstream package are working towards an exciting set of new features we would want to pull in. The last commit was from 6 months ago. 

I thus don't think we would want to receive updates from the upstream package. I could see this be its own package at some point, but for now I consider it small enough that I would prefer to keep it all in a single repo to simplify logistics (e.g. no need to coordinate PRs between repositories).

polaris/hub/client.py
cwognum: > could possibly be simplified to assume benchmark.md5sum is either response["md5Sum"] or the result of calling verify_checksum()

I may be missing something here, but I think that's the assumption right now! How would you simplify it?

polaris/dataset/_subset.py
cwognum: Fun fact: `hadim` is an old colleague! His accounts pops up everywhere (at least in most open-source projects relevant to us at Valence). I can actually remember discussing this issue with him! Feels like forever ago :smile:

polaris/hub/client.py
kiramclean: I was thinking checking for the presence of the response `"md5Sum"` could replace the `should_verify_checksum` call in this case, something like:
```
if response["md5sum"]:
    benchmark.md5sum = response["md5Sum"]
else:
    benchmark.verify_checksum()
```

(but haven't tested that, maybe it won't work.. was just wondering if we really need `should_verify_checksum`)..

polaris/dataset/_dataset.py
cwognum: Good idea! I introduced a Mixin class!

polaris/benchmark/_base.py
cwognum: Good idea!

polaris/hub/client.py
cwognum: Interesting idea, but I don't that will work. Even if we download the right checksum, verifying it is still needed to make sure we pulled in the rest of the data correctly as well.

polaris/dataset/_subset.py
kiramclean: Haha amazing! Small world.

polaris/utils/misc.py
kiramclean: ah makes sense.. backwards compatibility always leads to things like this.

kiramclean (approved):
This looks great to me. Nice work.

polaris/dataset/_dataset.py
```diff
@@ -185,69 +183,24 @@ def _compute_checksum(self):
...
-        """Set the checksum."""
-        if len(value) != 32 or not all(c in "0123456789abcdef" for c in value):
-            raise ValueError("The checksum should be a 32-character long MD5 hash.")
-        self._md5sum = value
-
-    @property
-    def has_md5sum(self) -> bool:
-        """Whether the md5sum for this class has been computed and stored."""
-        return self._md5sum is not None
+        return checksum
```
kiramclean, line 191: Should this return the manifest, too? Seems like either that or we should update the abstract class description (currently says "Compute the checksum and return it along with manifest.")

polaris/_mixins.py
```diff
@@ -0,0 +1,70 @@
...
+        """Lazily compute the checksum once needed."""
+        if not self.has_md5sum:
+            logger.info("Computing the checksum. This can be slow for large datasets.")
+            self.md5sum = self._compute_checksum()
+        return self._md5sum
+
+    @md5sum.setter
+    def md5sum(self, value: str):
+        """Set the checksum."""
+        if len(value) != 32 or not all(c in "0123456789abcdef" for c in value):
```
jstlaurent, line 33: **Suggestion**
An instance where a regular expression is your friend:
```suggestion
        if not re.fullmatch(r"^[a-f0-9]{32}$", value):
```

jstlaurent (approved):
Very nice! 😄

polaris/_mixins.py
cwognum: I want to get more comfortable with REs. It's always been a little daunting to me!

polaris/dataset/_dataset.py
cwognum: That's a mistake in the docs! Good catch!

PR #103, opened 2024-05-15
Competition evaluation

kiramclean:
## Changelogs

Adds support in Polaris client for evaluating competitions. The client runs the same evaluation logic as it does for benchmarks. Competitions differ from regular benchmarks because they have no knowledge of their own test sets (with targets hidden or otherwise). Test data is downloaded by Polaris Hub and passed to the competition evaluation.

API documentation will be done in a separate PR in the context of the rest of the competitions documentation.

---

_Checklist:_

---

Competitions use the exact same evaluation logic as regular benchmarks, but with the difference that test labels are passed in separately (rather than accessed from the benchmark/competition dataset). In the case of competitions, the target labels literally do not exist (as opposed to being hidden programatically), and are downloaded on the hub then passed to the evaluation function.

This raises some questions about whether competitions are really as similar to regular benchmarks as we thought they would be, but for now this works and the competition/benchmark distinction can be discussed separately.


polaris/evaluate/utils.py
```diff
@@ -0,0 +1,59 @@
+import pandas as pd
+import numpy as np
+
+from polaris.utils.context import tmp_attribute_change
+from polaris.evaluate import BenchmarkResults, ResultsType
+
+def evaluate_benchmark(model, y_pred, test):
```
kiramclean, line 7: This is copied, unmodified, from the implementation previously used in the `BenchmarkSpecification` `evaluate` method.

polaris/hub/client.py
```diff
@@ -751,19 +752,17 @@ def _upload_benchmark(
         )
         return response
 
-    
+
     def upload_competition(
-        self, 
-        dataset: Dataset,
```
kiramclean, line 757: Removed the reference to the dataset here, we don't need it.

polaris/hub/client.py
```diff
@@ -832,4 +831,30 @@ def list_competitions(
...
+
+        Args:
+            competition: The competition to evaluate the predictions for.
+            y_pred: The predictions for the test set, as NumPy arrays.
+                If there are multiple targets, the predictions should be wrapped in a dictionary with the target labels as keys.
+
+        Returns:
+             A `BenchmarkResults` object.
+        """
+        return self._base_request_to_hub(
```
kiramclean, line 850: This will call the corresponding hub endpoint (we should decide on what the URL should be), download the labels, then run the evaluation, returning a benchmark results object.

polaris/evaluate/utils.py
Andrewq11: Let's start a thread here to continue the conversations we've been having over Slack to decide if/how/when we should modify this utility to support both standard and competition benchmarks. The main issue here is that we want to share all evaluation logic between the two types of benchmarks, but the amount of data available/needed in the competition evaluation service is much less than the client. With the current utility function, we need to hack around this and create dummy data to bypass Pydantic schema checks and get this function to work correctly.

A couple options that come to mind:
- Extend this function to accept both pydantic objects or `dict`/`array` arguments (for standard and competition benchmarks, respectively). We can have type checks at the start of the function which will check the type of the argument passed and then extract all necessary data points according to the type (e.g. for pydantic objects `model.target_cols` and a regular dict `model["target_cols"]`). This should side step the need to change surrounding code which uses this function. 
- Update this function to expect `dict`/`array` arguments only. This will require some changes to the existing client code that evaluates standard benchmarks. However, it will standardize what is expected during evaluation which I like. 

Personally, I think option 2 is the cleanest, and if the changes to existing client code aren't huge, I would prefer to do this as part of the competition feature development. I also am ok with option 1. 

Curious to hear our thoughts on this and other options as well.

polaris/competition/_competition.py
```diff
@@ -1,33 +1,81 @@
 from datetime import datetime
 import os
 from typing import Optional, Union
+
+import numpy as np
```
Andrewq11, line 5:
```suggestion

```

polaris/competition/_competition.py
```diff
@@ -1,33 +1,81 @@
...
+            name=f'{self.name}_test_set',
+            description=f"Target labels for competition {self.name}. Used internally to evaluate competition predictions."
+        )
+        test_subset = Subset(
+            dataset=dataset,
+            indices=list(range(len(self.split[1]))),
+            input_cols=self.input_cols,
+            target_cols=self.target_cols,
+            hide_targets=False
+        )
```
Andrewq11, line 76: I don't think this is the end of the world if we decide to keep our evaluate utility expecting pydantic objects, but I think taking the extra couple hours to avoid being hacky makes sense. 

Basically, I +1 the thoughts @kiramclean has been sharing on this.

polaris/hub/client.py
Andrewq11: Nice catch!

polaris/hub/client.py
```diff
@@ -832,4 +831,30 @@ def list_competitions(
...
+            url=f"/v2/competition/evaluate",
```
Andrewq11, line 851:
```suggestion
            url="/v2/competition/evaluate",
```

polaris/hub/client.py
```diff
@@ -832,4 +831,30 @@ def list_competitions(
...
+            method="PUT",
+            json={
+                "competition": competition.model_dump(exclude_none=True,
+                                                      exclude=["dataset",
+                                                               "split"],
+                                                      by_alias=True),
```
Andrewq11, line 857: Just revisiting our thoughts on what we need to send over the wire here. This approach works, but I'm more in favour of just sending the competition identifier (owner/name) along with the predictions.

Using the identifier, we can then call the Hub server side to get the necessary data we need into the evaluation service. 

Once we decide on a server to server auth mechanism, we'll be able to implement that!

Andrewq11:
Nice stuff! Left a few comments and I hope we can use this PR as a thread for discussing what we will do regarding the subtle but clear differences in standard and competition benchmark evaluations.

polaris/evaluate/utils.py
kiramclean: I think you're right and it wouldn't be very difficult to implement this in a more "functional" (data-in/data-out) kind of way.. it might make sense to just do in this PR, assuming there are no other potentially conflicting changes in progress to benchmark evaluation upstream.

polaris/hub/client.py
kiramclean: This makes so much more sense than trying to reconstruct a valid `CompetitionSpecification` on the hub side -- I was initially trying to avoid having to figure out how to auth and access the db from the python endpoints run as verbal serverless functions, but you're right that we'll have to figure that out anyway regardless, so may as well do it here. It'll simplify a lot of things!

polaris/evaluate/utils.py
```diff
@@ -0,0 +1,62 @@
...
+
+def is_multi_task_single_test_set(vals: PredictionsType, target_cols: list[str]):
+    """Check if the given values are for a multiple-task benchmark with a single
+    test set. This is inferred by comparing the target names with the keys of the
+    given data. If all keys in the given data match the target column names, we
+    assume they are target names (as opposed to test set names for a single-task,
+    multiple test set benchmark)."""
+    return not isinstance(vals, dict) or set(vals.keys()) == set(target_cols)
+
+def evaluate_benchmark(y_pred: PredictionsType,
```
kiramclean, line 17: This depends on fewer objects now, with the evaluation logic not requiring any unnecessary or wrapper-only data types. Thanks for being up for a bit of re-jiggering things around for the sake of simplicity.

polaris/benchmark/_base.py
```diff
@@ -351,6 +350,37 @@ def task_type(self) -> TaskType:
...
+
+        test_split = self.split[1]
+        if isinstance(test_split, dict):
+            test = {k: make_test_subset(v) for k, v in test_split.items()}
+        else:
+            test = make_test_subset(test_split)
+
+        return test
+
     def get_train_test_split(
```
Andrewq11, line 384: I think the cleanup you've done here is quite nice. In combination with pulling out the evaluation logic into a separate utility, it makes the base logic here more easily understandable.

polaris/competition/_competition.py
```diff
@@ -1,33 +1,71 @@
 from datetime import datetime
 import os
 from typing import Optional, Union
+
+import pandas as pd
```
Andrewq11, line 5:
```suggestion
```

polaris/evaluate/utils.py
```diff
@@ -0,0 +1,62 @@
...
+from polaris.evaluate import BenchmarkResults, ResultsType
+from polaris.utils.types import PredictionsType
+from polaris.evaluate import Metric
+
+def is_multi_task_single_test_set(vals: PredictionsType, target_cols: list[str]):
+    """Check if the given values are for a multiple-task benchmark with a single
+    test set. This is inferred by comparing the target names with the keys of the
+    given data. If all keys in the given data match the target column names, we
+    assume they are target names (as opposed to test set names for a single-task,
+    multiple test set benchmark)."""
```
Andrewq11, line 14: This makes sense, but I can see this becoming a little messy in the future. Maybe this is something we can think about making more robust after we finish competitions.

polaris/hub/client.py
```diff
@@ -751,19 +752,17 @@ def _upload_benchmark(
...
         return response
 
-    
+
     def upload_competition(
-        self, 
-        dataset: Dataset, 
-        competition: CompetitionSpecification, 
+        self,
+        competition: CompetitionSpecification,
```
Andrewq11, line 758: Nice catch on removing the redundant dataset parameter

Andrewq11:
I like the cleanup done here! The complexity around the various prediction objects that can be passed for evaluation is making me think we should spend some time on this after we get a working version of competitions (and before we launch it).

I know you and @cwognum have already been sharing some good thoughts on this as well.

polaris/evaluate/utils.py
kiramclean: Yeah I totally agree.. this is at least isolated and explained here now, but we may want to re-think the data format we accept for predictions. There's a [brief discussion about it here](https://valencelabsworkspace.slack.com/archives/C05ASCA5456/p1715960350512859?thread_ts=1715785094.870159&cid=C05ASCA5456). I think it would be reasonable to always expect something like:
```
{"test_set_name": {"target_col_name": [1, 2, 3, 4, 5]}"
```

I get the desire to be as succinct as possible too though.. you could imagine it would be annoying to have to submit that ☝️ every time you just have a list of numbers to submit.

kiramclean:
Sorry for this list ☝️ Sad attempt to peacefully resolve merge conflicts went awry..

PR #109, opened 2024-05-29
Use evaluation logic directly in hub, no need for wrapper

kiramclean:
## Changelogs

This PR removes the wrapper around the evaluation logic for use by the hub. Instead we can just call `evaluate_benchmark` directly from there. This requires testing in the context of working competitions, which we will do next. This requires the accompanying PR in the hub, https://github.com/polaris-hub/polaris-hub/pull/338

---

_Checklist:_

---

### DIFF
--- a/polaris/evaluate/__init__.py
+++ b/polaris/evaluate/__init__.py
@@ -1,4 +1,9 @@
 from polaris.evaluate._metric import Metric, MetricInfo
 from polaris.evaluate._results import BenchmarkResults, ResultsType
+from polaris.evaluate.utils import evaluate_benchmark
 
-__all__ = ["Metric", "MetricInfo", "BenchmarkResults", "ResultsType"]
+__all__ = ["Metric",
+           "MetricInfo",
+           "BenchmarkResults",
+           "ResultsType",
+           "evaluate_benchmark"]
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -874,6 +874,6 @@ class PolarisHubClient(OAuth2Client):
                 "competition": competition.artifact_id,
                 "predictions": y_pred
             })
-        return BenchmarkResults(results=response["scores"],
+        return BenchmarkResults(results=pd.read_json(response["scores"]),
                                 benchmark_name=competition.name,
                                 benchmark_owner=competition.owner)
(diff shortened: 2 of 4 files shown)


tests/test_competition.py
```diff
@@ -11,17 +10,3 @@ def test_competition_from_json(test_competition, tmpdir):
     path = test_competition.to_json(str(tmpdir))
     new_competition = CompetitionSpecification.from_json(path)
     assert new_competition == test_competition
-
-def test_competition_evaluation(test_competition):
```
kiramclean, line 15: This is redundant now since the evaluation logic is the exact same for regular benchmarks.

polaris/evaluate/__init__.py
```diff
@@ -1,4 +1,9 @@
 from polaris.evaluate._metric import Metric, MetricInfo
 from polaris.evaluate._results import BenchmarkResults, ResultsType
+from polaris.evaluate.utils import evaluate_benchmark
 
-__all__ = ["Metric", "MetricInfo", "BenchmarkResults", "ResultsType"]
+__all__ = ["Metric",
+           "MetricInfo",
+           "BenchmarkResults",
+           "ResultsType",
+           "evaluate_benchmark"]
```
kiramclean, line 9: Adding this `evaluate_benchmark` function so it can be used from the hub.

polaris/hub/client.py
```diff
@@ -866,10 +867,13 @@ def evaluate_competition(
...
         """
-        return self._base_request_to_hub(
+        response = self._base_request_to_hub(
             url=f"/v2/competition/evaluate",
             method="PUT",
             json={
                 "competition": competition.artifact_id,
                 "predictions": y_pred
             })
+        return BenchmarkResults(results=pd.read_json(response["scores"]),
```
kiramclean, line 877: I realized the hub can just return the scores, so we don't need to faff with constructing a valid competition object there. This still returns a `BenchmarkResults`, but now the hub just does the evaluation and nothing more.

cwognum (commented):
Hi @kiramclean, this change makes sense to me, but I think I lack the broader context of this PR to properly review it! :smile: 

I will do a thorough review once `feat/competitions` is ready for review!

Andrewq11 (approved):

PR #111, opened 2024-06-05
integrating results for comps

Andrewq11:
## Changelogs

- updating `evalute_competition` method in client to handle result from Polaris Hub


polaris/hub/client.py
```diff
@@ -866,12 +869,15 @@ def evaluate_competition(
              A `BenchmarkResults` object.
         """
         response = self._base_request_to_hub(
-            url="/v2/competition/evaluate",
+            url=f"/v2/competition/{competition.owner}/{competition.name}/evaluate",
```
kiramclean, line 872: Thanks for updating this! Totally forgot about it after I updated the hub endpoint. It's just more convenient to follow the conventions already in place there for getting the artifact owner/id out of a request.

polaris/hub/client.py
```diff
@@ -866,12 +869,15 @@ def evaluate_competition(
              A `BenchmarkResults` object.
```
kiramclean, line 869: This doesn't return a `BenchmarkResults` anymore right? Just the JSON payload from the hub? We could turn it into one, but I don't think we need to. It's probably fine to just say this returns a 201 response on success or other error response from the hub.

cwognum (approved):

polaris/hub/client.py
```diff
@@ -866,12 +869,15 @@ def evaluate_competition(
...
-            json={"competition": competition.artifact_id, "predictions": y_pred},
+            json={"competition": competition.artifact_id, "predictions": y_pred, "access": access},
         )
-        return BenchmarkResults(
-            results=pd.read_json(response["scores"]),
-            benchmark_name=competition.name,
-            benchmark_owner=competition.owner,
+
+        # Inform the user about where to find their newly created artifact.
+        result_url = urljoin(
```
kiramclean, line 878: Can we grab this URL from the response Location header instead of constructing it manually?

Andrewq11: We could, I was just following the pattern from the Benchmark evaluate function

polaris/hub/client.py
Andrewq11: Nice catch!

PR #121, opened 2024-06-28
Competitions feature

kiramclean:
## Changelogs

This branch contains all of the changes to Polaris client required to support competitions in the hub, including:
- uploading new competition datasets and benchmarks to the hub
- retrieving competition datasets and benchmarks
- evaluating competition benchmarks

Some key notes about the structure of the whole system (i.e. including Polaris hub and supporting services):
- Competition datasets and competition benchmarks are new entities in the system. They behave very similarly to current datasets and benchmarks except for how results are evaluated.
- Evaluation logic is shared with regular benchmarks.
- Preparing competitions (splitting the target labels into a separate dataset) is done in a separate, standalone service.
- Evaluating competition benchmarks is done in a separate, standalone service.

Corresponding changes to the hub and the implementation of these new external services will be discussed in separate PRs in the relevant repos.

---

_Checklist:_

---

## Discussion

The original [design proposal is discussed in this doc](https://docs.google.com/document/d/1aGwhsKYl22ecjKzX97pmCLrgtHqpq6G2fbwc7FmMqHM/edit#heading=h.rffmldekabg), with further details in this [initial project proposal doc](https://docs.google.com/document/d/1t6R8DLkxDRsLuVW30lWOWOiCMdV0bqoLIQgE8euQ730/edit?pli=1#heading=h.rffmldekabg). This implementation closely follows the agreed-upon design in the first linked doc.

Briefly, this feature allows users to launch a competition on Polaris Hub, which are essentially regular benchmarks with hidden test labels. Predictions and evaluation results associated with a particular competition can be uploaded to the hub for a given period of time, after which the competition benchmark labels are made available and the dataset is converted to a "regular" dataset on Polaris Hub. A typical user workflow would involve uploading a dataset to polaris hub, launching the competition, and then collecting results. 

Competition evaluation logic is shared with regular benchmarks, and there has been some refactoring to facilitate re-using this code. **Note:** Currently competition benchmarks accept only predictions, not probabilities. Support for probabilities was adding part way through the development of the competitions feature and will need to be implemented in the evaluation service before competitions can support probabilities. 

Other large and potentially interesting changes to discuss are highlighted below in comments in the relevant context.

### Deployment

We will need to coordinate deployment of these changes to make sure the the corresponding changes to the hub and other services are available before we release a version of the client containing these changes. No existing functionality should break, but the newly available competition-related APIs will not work until the updated hub and other services are live.


polaris/benchmark/_base.py
```diff
@@ -418,70 +434,15 @@ def evaluate(
         # Instead of having the user pass the ground truth, we extract it from the benchmark spec ourselves.
         # This simplifies the API, but also was added to make accidental access to the test set targets less likely.
         # See also the `hide_targets` parameter in the `Subset` class.
-        test = self.get_train_test_split()[1]
-
-        if not isinstance(test, dict):
-            test = {"test": test}
-
-        y_true = {}
```
kiramclean, line 426: All of this logic beginning here has been moved to `evaluate_benchmark` in `polaris/evaluate/utils.py`. The functionality is unchanged, with some minor refactoring to account for the possibility that `y_prob` may be `None`.

polaris/competition/_competition.py
```diff
@@ -0,0 +1,99 @@
...
+    field_validator,
+    ValidationInfo
+)
+from polaris.benchmark import BenchmarkSpecification
+from polaris.hub.settings import PolarisHubSettings
+from polaris.utils.types import AccessType, HubOwner, PredictionsType, TimeoutTypes, ZarrConflictResolution
+from polaris.utils.errors import InvalidCompetitionError
+
+
+class CompetitionSpecification(BenchmarkSpecification):
```
kiramclean, line 16: This is the new `CompetitionBenchmark` class, to accommodate the differences between regular benchmarks and competition benchmarks, notably:
- Competition benchmarks have a start and end time, since they only accept results for the duration of the competition.
- We validate that all test rows have at least one value across the target columns to ensure there are no rows in the test set with missing test data.
- Evaluation is handled in a separate, standalone service. This is to ensure that test labels are never processed in the client so there is no way users can access them. Users will only see the results reported from the evaluation service, but the test labels never exist on their machines.
- Uploading to hub is handled in a separate, standalone service, because some pre-processing needs to be done on the dataset file before it is available (the test labels are separated and masked in the dataset, so they are not available until a competition is over.

polaris/dataset/_competition_dataset.py
```diff
@@ -0,0 +1,56 @@
...
+            raise InvalidDatasetError("There are default adapters for columns that do not exist")
+
+        # Set a default for missing annotations and convert strings to Modality
+        for c in m.table.columns:
+            if c not in m.annotations:
+                m.annotations[c] = ColumnAnnotation()
+            m.annotations[c].dtype = m.table[c].dtype
+
+        # Verify the checksum
+        actual = m.masked_md5sum
```
kiramclean, line 40: This is the difference with regular datasets -- competition datasets have certain masked values that need to be accounted for.

polaris/evaluate/utils.py
```diff
@@ -0,0 +1,122 @@
...
+    else:
+        return {"test": {target_cols[0]: vals}}
+
+def safe_mask(dict, test_label, target_label, mask):
+    if dict is None or dict[test_label] is None or dict[test_label][target_label] is None:
+        return None
+    else:
+        return dict[test_label][target_label][mask]
+
+def evaluate_benchmark(
```
kiramclean, line 70: This is the shared evaluation logic. Regular and competition benchmarks use the exact same evaluation logic, but there has been some refactoring to facilitate sharing this code.

polaris/benchmark/_base.py
```diff
@@ -96,7 +95,7 @@ class BenchmarkSpecification(BaseArtifactModel):
 
     # Public attributes
     # Data
-    dataset: Union[Dataset, str, dict[str, Any]]
+    dataset: Union[Union[Dataset, CompetitionDataset], str, dict[str, Any]]
```
jstlaurent, line 98: **Suggestion**
This can be simplified to this (since our minimum is Python 3.10):
```suggestion
    dataset: Dataset | CompetitionDataset | str | dict[str, Any]
```

polaris/hub/client.py
```diff
@@ -662,18 +720,193 @@ def upload_benchmark(
...
             owner: Which Hub user or organization owns the artifact. Takes precedence over `benchmark.owner`.
         """
+        return self._upload_benchmark(benchmark, ArtifactType.STANDARD.value, access, owner)
+
+    def _upload_benchmark(
+        self,
+        benchmark: BenchmarkSpecification | CompetitionSpecification,
+        artifact_type: ArtifactType,
+        access: AccessType = "private",
+        owner: Optional[Union[HubOwner, str]] = None,
```
jstlaurent, line 730: **Suggestion**
Could be simplified to:
```suggestion
        owner: HubOwner | str | None = None,
```

polaris/hub/client.py
```diff
@@ -662,18 +720,193 @@ def upload_benchmark(
...
+            f"View it here: {urljoin(self.settings.hub_url, f'{artifact_type}/{benchmark.owner}/{benchmark.name}')}"
         )
         return response
+
+    def upload_competition(
+        self,
+        competition: CompetitionSpecification,
+        timeout: TimeoutTypes = (10, 200),
+        owner: Optional[Union[HubOwner, str]] = None,
+        if_exists: ZarrConflictResolution = "replace",
```
jstlaurent, line 776: **Comment**
Since we don't support Zarr-based competitions, this shouldn't be here.

polaris/hub/client.py
```diff
@@ -662,18 +720,193 @@ def upload_benchmark(
...
+            A `CompetitionSpecification` instance, if it exists.
+        """
+
+        response = self._base_request_to_hub(url=f"/v2/competition/{owner}/{name}", method="GET")
+
+        # TODO (cwognum): Currently, the competition endpoints do not return the owner info for the underlying dataset.
+        # TODO (jstlaurent): Use the same owner for now, until the competition returns a better dataset entity
+        response["dataset"] = self._get_dataset(
+            owner, response["dataset"]["name"], ArtifactType.COMPETITION, verify_checksum=verify_checksum
+        )
```
jstlaurent, line 834: **Suggestion**
The Hub now does return the proper owner for the related dataset:
```suggestion
        # TODO (jstlaurent): response["dataset"]["artifactId"] is the owner/name unique identifier,
        #  but we'd need to change the signature of get_dataset to use it
        response["dataset"] = self._get_dataset(
            response["dataset"]["owner"]["slug"], 
            response["dataset"]["name"], 
            ArtifactType.COMPETITION, 
            verify_checksum=verify_checksum
        )
```

polaris/hub/client.py
```diff
@@ -662,18 +720,193 @@ def upload_benchmark(
...
+        # Upload competition benchmark
+        competition_response = self._upload_benchmark(
+            competition, ArtifactType.COMPETITION.value, ACCESS, owner
+        )
+
+        return {
+            "dataset_response": dataset_response,
+            "competition_response": competition_response,
+        }
```
jstlaurent, line 812: **Comment**
It's a little late in the game, but I think this illustrates what I was saying early in the design process, where a competition isn't `dataset + benchmark`, it's `competition`. It _looks_ like it's two things (because we have two similar things, data wise), but it doesn't _behave_ like two things. It's one thing.

Unfortunately, that means we end up with some leaking abstraction here, where if the second request to the Hub fails, the user is stuck with half a competition and possibly can't recover.

Having a single API endpoint to upload all the metadata at once, and then following up with the blob content upload to the bucket, would provide a better user experience.

polaris/hub/client.py
```diff
@@ -662,18 +720,193 @@ def upload_benchmark(
...
+                "access": results_access,
+                "polaris_version": po.__version__,
+                "name": result_name,
+                "description": description,
+                "tags": tags,
+                "user_attributes": user_attributes,
+                "github_url": github_url,
+                "paper_url": paper_url,
+                "contributors": contributors,
+            },
```
jstlaurent, line 899: **Comment**
I'd feel better if this was defined in a Pydantic model that was serialized to JSON for the request.

jstlaurent (commented):
I'll let Cas comment on the domain-specific changes, he's better placed to evaluate. It looks pretty from my part. I did have some questions and comments. I have one concern in particular where the competition upload user experience might not be great.

polaris/hub/client.py
cwognum: I agree! This is essentially the results data model, right? But then just switching the `scores` attribute for the `predictions` one? Maybe we can split off a `ResultsMetadata` datamodel that is used by both the results and the object we would create here?

polaris/hub/client.py
```diff
@@ -662,18 +720,193 @@ def upload_benchmark(
...
+        result_url = urljoin(
+            self.settings.hub_url,
+            f"/v2/competition/{competition.owner}/{competition.name}/{response['id']}",
+        )
+        logger.success(f"Your result has been successfully uploaded to the Hub. View it here: {result_url}")
+
+        scores = response["results"]
+        return CompetitionResults(
+            results=scores, competition_name=competition.name, competition_owner=competition.owner
+        )
```
cwognum, line 912: **Fix:** All the additional metadata that is part of the method signature should also be included in the object that is returned here I believe.

polaris/hub/client.py
```diff
@@ -628,9 +683,12 @@ def upload_dataset(
                     if_exists=if_exists,
                 )
 
+        base_artifact_url = (
+            "datasets" if artifact_type == ArtifactType.STANDARD.value else "/competition/datasets"
```
cwognum, line 687: **Suggestion:** This is a comment for the Hub PR, but I'll leave it here for now so I don't forget. Wouldn't it be more consistent to use: 

```suggestion
            "datasets" if artifact_type == ArtifactType.STANDARD.value else "/competitions/datasets"
```

?

cwognum (commented):
Monumental work and a long time coming! Thank you @Andrewq11 and @kiramclean ! 🚀

polaris/benchmark/_base.py
```diff
@@ -418,70 +434,15 @@ def evaluate(
...
-                        if y_pred[test_label] is not None
-                        else None,
-                        y_prob=y_prob[test_label][target_label][mask]
-                        if y_prob[test_label] is not None
-                        else None,
-                    )
-                    scores.loc[len(scores)] = (test_label, target_label, metric, score)
+        scores = evaluate_benchmark(self.target_cols, self.metrics, y_true,
+                                    y_pred=y_pred,
+                                    y_prob=y_prob)
```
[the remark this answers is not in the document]
kiramclean: Amazing, thanks. I didn't know there was an auto-formatter for python. Thanks!

polaris/competition/_competition.py
```diff
@@ -0,0 +1,99 @@
...
+        """Verifies that the test does not have too many missing values. There must be at
+        least one value per row in the test set across the target columns."""
+        dataset = info.data.get("dataset")
+        target_cols = info.data.get("target_cols")
+        test_indices = split[1]
+
+        if dataset.table.loc[test_indices, target_cols].notna().any(axis=1).all():
+            return split
+        else:
+            raise InvalidCompetitionError("All rows of the test set must have at least one value.")
```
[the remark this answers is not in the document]
kiramclean: We weren't sure if this check was necessary for regular benchmarks, where people might be able to process the test set themselves (e.g. if they wanted to fill in missing values rather than remove them). But it seems reasonable to disallow missing test labels if you think that won't cause too much friction for people trying to upload benchmarks.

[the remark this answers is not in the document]
cwognum: Ah, I see! To make sure results are comparable, it's important they use the exact same test set. Therefore I would not want to allow e.g. imputing missing test values.

polaris/hub/client.py
kiramclean: I brought this up with Andrew yesterday -- you're right that it's possible to end up in a situation where the dataset and benchmark parts of a competition are out of sync and it's not fixable (because if re-try it will say the dataset already exists even though there's no benchmark (competition) with it. I think we should think through how to handle the case where a dataset gets orphaned and can't be used (or the competition gets orphaned, e.g. if the dataset pre-processing fails).

Andrewq11: Yeah this could be troublesome. It's also a difficult problem to completely fix/overcome because we at the least have two independent writes happening where the name must be unique (1 for the R2 blob, 1 transaction for dataset and competition records, assuming we bundle them in a transaction).

We'll have to put one of these in front of the other, and if it fails, the user still won't be able to retry with the same name. Definitely worth a chat as Kira mentioned.

polaris/hub/client.py
```diff
@@ -338,8 +346,36 @@ def get_dataset(self, owner: Union[str, HubOwner], name: str, verify_checksum: b
...
+            verify_checksum: Whether to use the checksum to verify the integrity of the dataset.
+
+        Returns:
+            A `Dataset` instance, if it exists.
+        """
+
+        url = (
+            f"/v1/dataset/{owner}/{name}"
+            if artifact_type == ArtifactType.STANDARD.value
+            else f"/v2/competition/dataset/{owner}/{name}"
```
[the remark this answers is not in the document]
Andrewq11: The idea here was to migrate to an API naming scheme that had two main branches, `v2/competition` and `v2/standard`. This should provide some very clear distinction between `competition` and `standard` actions. 

I'm definitely not married to this and wanted to discuss the API naming with the team prior to launch. I'm not sure we'll have tons of time to do this though.

polaris/competition/_competition.py
```diff
@@ -0,0 +1,99 @@
...
+        from polaris.hub.client import PolarisHubClient
+
+        with PolarisHubClient(
+            env_file=env_file,
+            settings=settings,
+            cache_auth_token=cache_auth_token,
+        ) as client:
+            return client.upload_competition(self.dataset, self, access, timeout, owner, if_exists)
+
+    @field_serializer("start_time", "scheduled_end_time")
```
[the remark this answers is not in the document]
Andrewq11: Oops, forgot to remove `actual_end_time` previously. We don't track this on the hub side so there is no use for it anymore.

polaris/hub/client.py
```diff
@@ -662,18 +720,193 @@ def upload_benchmark(
...
-        url = f"/benchmark/{benchmark.owner}/{benchmark.name}"
+        path_params = "v1/benchmark" if artifact_type == ArtifactType.STANDARD.value else "v2/competition"
+        url = f"/{path_params}/{benchmark.owner}/{benchmark.name}"
         response = self._base_request_to_hub(url=url, method="PUT", json=benchmark_json)
 
         logger.success(
-            "Your benchmark has been successfully uploaded to the Hub. "
-            f"View it here: {urljoin(self.settings.hub_url, f'benchmarks/{benchmark.owner}/{benchmark.name}')}"
+            f"Your {artifact_type} has been successfully uploaded to the Hub. "
+            f"View it here: {urljoin(self.settings.hub_url, f'{artifact_type}/{benchmark.owner}/{benchmark.name}')}"
```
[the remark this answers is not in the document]
Andrewq11: It is not. Good catch!

polaris/hub/client.py
Andrewq11: Benchmark routes are just `/v1/benchmark/*`, so we followed the same here!

polaris/benchmark/_base.py
```diff
@@ -432,72 +470,32 @@ def evaluate(
...
+            #
+            # For multi-set benchmarks
+            y_true = {}
+            for test_set_name, values in test.items():
+                y_true[test_set_name] = {}
+                if isinstance(values.targets, dict):
+                    #
+                    # For multi-task, multi-set benchmarks
+                    for task_name, values in values.targets.items():
+                        y_true[test_set_name][task_name] = values
```
cwognum, line 487: **Suggestion:** I believe this would do the same thing:

```suggestion
                    y_true[test_set_name] = values.targets
```

polaris/benchmark/_base.py
```diff
@@ -432,72 +470,32 @@ def evaluate(
...
+                else:
+                    #
+                    # For single task, multi-set benchmarks
+                    for task_name in self.target_cols:
+                        y_true[test_set_name][task_name] = values.targets
```
cwognum, line 492: **Suggestion:** To clearly communicate that this is a single task benchmark:
```suggestion
                    y_true[test_set_name][self.target_cols[0]] = values.targets
```

polaris/evaluate/utils.py
```diff
@@ -0,0 +1,89 @@
...
+            return {"test": vals}
+        else:
+            return vals
+    elif vals is None:
+        return None
+    else:
+        return {"test": {target_cols[0]: vals}}
+
+
+def safe_mask(dict, test_label, target_label, mask):
```
cwognum, line 31: **Suggestion:** I would use another variable name than `dict` to not clash with the `dict` Python keyword.

polaris/evaluate/utils.py
```diff
@@ -0,0 +1,89 @@
...
+    if dict is None or dict[test_label] is None or dict[test_label][target_label] is None:
```
cwognum, line 32: **Question:** This feels off to me. Seems like it could raise a `KeyError`. Do you mean `dict.get(test_label)` rather than `dict[test_label]`? The difference being that if the `test_label` key is not in the dict, it will return `None` when using `get` and will throw an `KeyError` in the case of `[...]`.

tests/test_integration.py
```diff
@@ -56,10 +57,11 @@ def test_single_task_benchmark_clf_loop_with_multiple_test_sets(
 
     y_prob = {}
     y_pred = {}
+    task_name = test_single_task_benchmark_clf_multiple_test_sets.target_cols[0]
     for k, test_subset in test.items():
         x_test = np.array([dm.to_fp(dm.to_mol(smi)) for smi in test_subset.inputs])
-        y_prob[k] = model.predict_proba(x_test)[:, :1]  # for binary classification
-        y_pred[k] = model.predict(x_test)
+        y_prob[k] = {task_name: model.predict_proba(x_test)[:, :1]}  # for binary classification
```
cwognum, line 63: **Question:** Are you required to specify the prediction as a dictionary here? I would think not from the code.

polaris/evaluate/utils.py
```diff
@@ -0,0 +1,89 @@
...
+    multiple test set benchmark)."""
+    return all(k in target_cols for k in vals)
+
+
+def normalize_predictions_type(vals: PredictionsType, target_cols: list[str]):
+    if isinstance(vals, dict):
+        if is_multi_task_single_test_set(vals, target_cols):
+            return {"test": vals}
+        else:
+            return vals
```
cwognum, line 24: **Question:** Should we differentiate multi-test, single-task and single-test, multi-task here? Seems we would want to standardize these two cases to a 2-level dictionary as well (i.e. `dict[str, dict[str, np.ndarray]]`)

cwognum (commented):
Thanks once more @Andrewq11 and @kiramclean ! :rocket: 

I'm just left with some minor comments and question. The main difficulty may come from the checksum. Since I made those changes to the checksum, I'm of course happy to help if so! :smile: 

I did break my head over the predictions type in reviewing this. This is becoming increasingly complex and having a single interface where the predictions are standardized would give me peace of mind. In another PR though!

polaris/competition/_competition.py
```diff
@@ -0,0 +1,50 @@
...
+            cache_auth_token=cache_auth_token,
+            **kwargs,
+        ) as client:
+            return client.evaluate_competition(self, predictions)
+
+    @field_serializer("start_time", "end_time")
+    def _serialize_start_date(self, v):
+        """Convert from datetime to string to make sure it's serializable"""
+        if v:
+            return v.isoformat()
```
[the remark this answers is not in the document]
Andrewq11: Honestly we probably don't even need this. The `start_time` and `end_time` data is defined in the web app and is never used for operations on the client side. We can just leave it as a string if people are curious and want to access the attribute (e.g. to print it out). What do you think?

polaris/competition/_competition.py
```diff
@@ -0,0 +1,50 @@
...
+    anything within this class serves as a point of differentiation between the two.
+
+    Attributes:
+        owner: A slug-compatible name for the owner of the competition. This is redefined such
+            that it is required.
+        start_time: The time at which the competition becomes active and interactable.
+        end_time: The time at which the competition ends and is no longer interactable.
+    """
+
+    # Additional properties specific to Competitions
```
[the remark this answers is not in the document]
Andrewq11: Yes, good catch.

polaris/dataset/_competition_dataset.py
```diff
@@ -0,0 +1,42 @@
...
+
+        if actual is None:
+            m.masked_md5sum = expected
+        elif actual != expected:
+            raise PolarisChecksumError(
+                "The dataset md5sum does not match what was specified in the meta-data. "
+                f"{actual} != {expected}"
+            )
+
+        return m
```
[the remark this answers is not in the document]
Andrewq11: Actually, no I don't think so. I think we can just rename the fetched `masked_md5sum` attribute to `md5sum` when instantiating a standard `Dataset` object and do away with this entire `CompetitionDataset` class.

polaris/dataset/_competition_dataset.py
```diff
@@ -0,0 +1,42 @@
+from typing import Optional
+
+from pydantic import model_validator
+from polaris.dataset import Dataset
+from polaris.utils.errors import PolarisChecksumError
+
+_CACHE_SUBDIR = "datasets"
+
+
+class CompetitionDataset(Dataset):
```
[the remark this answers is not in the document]
Andrewq11: True, good idea. This could be a reason to keep the `CompetitionDataset` class even though we don't need the `masked_md5sum` attribute anymore. I would prefer to add this and keep it.

polaris/benchmark/_base.py
```diff
@@ -213,12 +214,23 @@ def _validate_split(cls, v, info: ValidationInfo):
...
-        return v
+        # For a given row in the test set, all target columns must not be missing/empty.
+        # NOTE: This could break a bunch of benchmarks already on the hub UNLESS we turn off
+        # validation on download from the Hub.
+        target_cols = m.target_cols
+        test_indices = list(full_test_idx_set)
+        if not dataset.table.loc[test_indices, target_cols].notna().any(axis=1).all():
+            raise InvalidBenchmarkError(
+                "All rows of the test set must have at least one target column with a value."
+            )
```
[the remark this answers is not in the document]
Andrewq11: I especially agree with the `Auroris` comment. I'll remove it.

polaris/benchmark/_base.py
```diff
@@ -432,72 +470,32 @@ def evaluate(
...
+                        y_true[test_set_name][task_name] = values
+                else:
+                    #
+                    # For single task, multi-set benchmarks
+                    for task_name in self.target_cols:
+                        y_true[test_set_name][task_name] = values.targets
+        else:
+            #
+            # For single set benchmarks (single and multiple task)
+            y_true = test.targets
```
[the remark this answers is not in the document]
Andrewq11: Yes, quite confusing. The idea here is that the structure of the `y_true` object must match that of the `y_pred` and `y_prob` objects, and those depend on the number of tasks and test sets. I hope that comment you found showing examples of this will help users in the time being. 

And I agree, I think the `BenchmarkPredictions` class mentioned in #169 with the validators would be awesome for UX.

polaris/competition/_competition.py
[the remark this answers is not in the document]
cwognum: Given the pressure to ship this feature, I think that is a reasonable suggestion!

polaris/benchmark/_base.py
Andrewq11: I believe I tried that initially. For multi-task, multi-target benchmarks, `values.targets` held a structure like this:

`{task_name_1: [values...], task_name_2: [values...]}`

tests/test_integration.py
Andrewq11: Yes, the new `evaluate_benchmark` method does expect it to be a dictionary for single task, multi-set benchmarks.

polaris/evaluate/utils.py
Andrewq11: I think we should do this another PR. I'll create a ticket.

polaris/competition/_competition.py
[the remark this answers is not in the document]
Andrewq11: On second thought, this breaks some functionality related to the  `_compute_checksum` function that exists in the parent `BenchmarkSpecification` class. The `_compute_checksum` function references `self.dataset` which is equal to `None` if we redefine `dataset` in the child `CompetitionSpecification`.

We could pull that checksum function into a util folder and call it like that across the code base, but I think that is out of scope of this PR. I'll write a ticket for it.

polaris/hub/client.py
```diff
@@ -302,14 +308,41 @@ def get_dataset(
...
-            response = self._base_request_to_hub(url=f"/dataset/{owner}/{name}", method="GET")
-            storage_response = self.get(response["tableContent"]["url"])
+            url = (
+                f"/v1/dataset/{owner}/{name}"
+                if artifact_type == ArtifactSubtype.STANDARD.value
+                else f"/v2/competition/dataset/{owner}/{name}"
+            )
+            response = self._base_request_to_hub(url=url, method="GET")
+            dataset_info = (
+                "tableContent" if artifact_type == ArtifactSubtype.STANDARD.value else "maskedDatasetInfo"
```
jstlaurent, line 343: **Question**
Could we use the same attribute name in the endpoint response?

polaris/hub/client.py
```diff
@@ -302,14 +308,41 @@ def get_dataset(
...
-            success_msg="Fetched dataset.",
+            start_msg="Fetching artifact...",
+            success_msg="Fetched artifact.",
             error_msg="Failed to fetch dataset.",
         ):
-            response = self._base_request_to_hub(url=f"/dataset/{owner}/{name}", method="GET")
-            storage_response = self.get(response["tableContent"]["url"])
+            url = (
+                f"/v1/dataset/{owner}/{name}"
+                if artifact_type == ArtifactSubtype.STANDARD.value
```
jstlaurent, line 338: **Suggestion**
Instead of passing in an enum to denote the concrete type we're looking for, we could pass in the class itself (`Dataset` or `CompetitionDataset`), and set the endpoint URL pattern, dataset info field name, and md5 sum field name, as class attributes.

We have specific classes that identify each artifact type, but we scatter their specific information across the code.

If we want to keep the information related to the client separate from the artifact data structure, I'd suggest putting all these information in an `ArtifactType` enum. Something like this:
```python
class ArtifactType(Enum):
    DATASET = ("dataset", "{api_url}/v1/dataset/{owner}/{name}", Dataset)
    BENCHMARK = ("benchmark", "{api_url}/v1/benchmark/{owner}/{name}", BenchmarkSpecification)
    COMPETITION = ("competition", "{api_url}/v2/competition/{owner}/{name}", Competition)
    COMPETITION_DATASET = ("competition_dataset", "{api_url}/v2/competition/datasets/{owner}/{name}", CompetitionDataset)

    def __init__(self, name: str, url_pattern: str, model: BaseArtifactModel):
        self.name = name
        self.url_pattern = url_pattern
        self.model = model

    def get_url(self, owner: str, name: str):
        return self.url_pattern.format(api_url=settings.api_url, owner=owner, name=name)
```

jstlaurent (approved):
This is good stuff, very nice! A few things here and there, nothing major. I do think with the multiplication of artifact types, we should start to centralize their relevant attributes in one place, rather than having conditionals all over the code.

polaris/benchmark/_base.py
```diff
@@ -399,6 +418,16 @@ def evaluate(
...
 
+        info: Expected structure for `y_pred` and `y_prob` arguments
+            The supplied `y_pred` and `y_prob` arguments must adhere to a certain structure depending on the number of
+            tasks and test sets included in the benchmark. Refer to the following for guidance on the correct structure when
+            creating your `y_pred` and `y_prod` objects:
+
+            - Single task, single set: `[values...]`
+            - Multi-task, single set: `{task_name_1: [values...], task_name_2: [values...]}`
+            - Single task, multi-set: `{test_set_1: {task_name: [values...]}, test_set_2: {task_name: [values...]}}`
+            - Multi-task, multi-set: `{test_set_1: {task_name_1: [values...], task_name_2: [values...]}, test_set_2: {task_name_1: [values...], task_name_2: [values...]}}`
```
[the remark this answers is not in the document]
Andrewq11: Exactly, and very much agreed. I'll be writing a bunch of tickets for things we need to do post-merge of competitions. Standardizing this structure is at the top of that list.

polaris/benchmark/_base.py
```diff
@@ -101,14 +100,12 @@ class BenchmarkSpecification(BaseArtifactModel, ChecksumMixin):
...
-    main_metric: Optional[Union[str, Metric]] = None
+    metrics: Union[str, Metric, list[str | Metric]]
+    main_metric: str | Metric | None = None
 
     # Additional meta-data
     readme: str = ""
-    target_types: dict[str, Optional[Union[TargetType, str]]] = Field(
-        default_factory=dict, validate_default=True
-    )
+    target_types: dict[str, Union[TargetType, str, None]] = Field(default_factory=dict, validate_default=True)
```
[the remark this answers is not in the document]
Andrewq11: I believe you had a similar suggestion in your first round of review. I agree, but I think we should address this in a follow up PR where we apply this change to the entire codebase for consistency. 

It won't be a heavy lift, I would just prefer to do it all at once and that would be out of scope for this already monster PR. What do you think?

polaris/hub/client.py
Andrewq11: We could, but we'll need to add the `tableContent` field to the response object just before it is returned from the Hub. I'm currently addressing PR feedback on the Hub so I can add this in.

polaris/hub/client.py
Andrewq11: Great idea, I'll write a ticket for this.

PR #122, opened 2024-06-28
feat: JWT-based authentication for Hub API requests

### LINKED ISSUE #113: Support JWT-based authentication with the Hub
# Context

The Polaris Hub implemented an [OAuth 2 compliant endpoint](https://github.com/polaris-hub/polaris-hub/issues/325) that will allow this client to obtain a JSON Web Token (JWT), in exchange from the current Clerk-issued, non-JWT token.

This new JWT will contain [claims](https://datatracker.ietf.org/doc/html/rfc7519#section-4) that the Hub can use to authenticate requests made to its API, using a lighter, lower latency process.

# Description

When the user authenticates with the Hub through the client, implement a token exchange flow, where the initial, Clerk-issued token is exchanged for a Hub JWT.

When sending requests to the Hub API, use this new JWT as the authentication header.

# Acceptance Criteria

- When the client sends a request to the Hub, it uses the new JWT token for authentication.

# Links

- [Hub implementation issue](https://github.com/polaris-hub/polaris-hub/issues/325)
- [OAuth 2.0 Token Exchange](https://datatracker.ietf.org/doc/html/rfc8693)

jstlaurent:
# Description

This PR implements a JWT-based authentication system when making request to the Polaris Hub API.

It does this by leveraging the OAuth Token Exchange flow ([RFC8693](https://datatracker.ietf.org/doc/html/rfc8693)) in order to exchange the non-JWT token provided by the external authentication service and obtaining a Hub-issued JSON Web Token. That JWT is then used in the authentication header for requests made to the Hub API.

I've split the responsabilities between two authlib clients:

- An external auth client: This client mediates the authentication, and the token fetch/refresh with the external authentication service. (Currently Clerk.)
- A Hub client: This client mediates the exchange of tokens, and ensures Hub API calls are using the JWt for authentication.

Along the way, I've refactored out the token file caching mechanism into subclasses of authlib's `TokenAuth` class, to lighten the client code, and make it easier to test. 

# Links
- Closes #113


polaris/hub/client.py
```diff
@@ -229,44 +215,8 @@ def request(self, method, url, withhold_token=False, auth=httpx.USE_CLIENT_DEFAU
...
-
-        return self._user_info
-
-    @property
-    def user_as_owner(self) -> HubOwner:
-        """Easily get the currently logged-in user a `HubOwner` instance."""
-        return HubOwner(user_id=self.user_info["user_id"], slug=self.user_info["username"])
 
-    def login(self, overwrite: bool = False, auto_open_browser: bool = True):
+    def interactive_login(self, overwrite: bool = False, auto_open_browser: bool = True):
```
cwognum, line 219: **Documentation**: Although it is rare, we do have some places where we also show `PolarisHubClient().login()`. For example in [this tutorial](https://polaris-hub.github.io/polaris/stable/tutorials/basics.html) or in some of the tutorials we've given, e.g. at the ML4DD hackathon. Let's at least update the tutorial, but I would keep `login()` with a deprecation warning for the time being as well.

polaris/hub/client.py
```diff
@@ -178,41 +195,10 @@ def _normalize_owner(
     #     Overrides
     # =========================
```
cwognum, line 196: **Documentation**: This "header" isn't correct anymore, since some of the overrides are placed above it. Feel free to remove it too. Don't think it's needed, especially now that this PR simplifies the code by splitting it over different classes!

polaris/hub/external_auth_client.py
```diff
@@ -0,0 +1,135 @@
...
+from authlib.integrations.httpx_client import OAuth2Client
+from authlib.oauth2 import TokenAuth
+from loguru import logger
+
+from polaris.hub.oauth import ExternalCachedTokenAuth
+from polaris.hub.settings import PolarisHubSettings
+from polaris.utils.errors import PolarisUnauthorizedError
+
+
+class ExternalAuthClient(OAuth2Client):
```
cwognum, line 15: **Suggestion:** I would include this in the generated docs site!

cwognum (approved):
Looks good to me, @jstlaurent ! I just left some minor comments.

I like how this is simplifying the code due to the restructuring, while adding functionality! :pray:

polaris/hub/client.py
```diff
@@ -88,40 +86,59 @@ def __init__(
...
+            cert=self.settings.ca_bundle,
             timeout=self.settings.default_timeout,
             # Extra
             **kwargs,
         )
 
-    def _load_from_signed_url(self, url: URLTypes, load_fn: Callable, headers: Optional[HeaderTypes] = None):
+        # We use an external client to get an auth token that can be exchanged for a Polaris Hub token
+        self.external_client = ExternalAuthClient(
+            settings=self.settings, cache_auth_token=cache_auth_token, **kwargs
```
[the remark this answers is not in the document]
jstlaurent: Yes, they would be. I don't think we actually use these ourselves, and I don't know how users would leverage them, but the option's there.

polaris/hub/client.py
jstlaurent: Good point! I'll revert the name change.

polaris/hub/client.py
```diff
@@ -404,7 +383,7 @@ def upload_results(
         """
 
         # Get the serialized model data-structure
-        results.owner = self._normalize_owner(results.owner, owner)
+        results.owner = HubOwner.normalize(results.owner and owner)
```
cwognum, line 386: **Compliment:** Good idea! Love it! 💅

polaris/hub/external_auth_client.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    This authentication client is used to obtain OAuth 2 tokens from Polaris's external OAuth2 server.
+    These can in turn be used to obtain Polaris Hub tokens.
+
+    This class is intended for internal use by the `PolarisHubClient` class, and you should not have to
+    interact with it directly.
```
cwognum, line 21: **Suggestion:** Minor, but I like using [admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) for comments like these so that they stand out more in the generated docs. 

For example:

```suggestion
    Note: Internal use
        This class is intended for internal use by the `PolarisHubClient` class, and you should not have to
        interact with it directly.
```

cwognum (approved):

polaris/utils/types.py
```diff
@@ -125,6 +125,13 @@ class HubOwner(BaseModel):
     def __str__(self):
         return self.slug
 
+    @staticmethod
```
cwognum, line 128: **Question:** Should this be a `@classmethod` instead?

jstlaurent: I mean, you could implement it as a class method and do:
```python
@classmethod
def normalize(cls, owner: Union[str, "HubOwner"]) -> "HubOwner":
    """
    Normalize a string or `HubOwner` instance to a `HubOwner` instance.
    """
    return owner if isinstance(owner, cls) else cls(slug=owner)
```

Feels slightly less clear to me, since we need to the class as a type in the signature anyway.

cwognum: I think this may be a misunderstanding on my end. I actually thought a `@classmethod` _had to_ return an object of that class. I do thus strongly associate class methods with factory-like methods, but if that association is less for you I'm okay with keeping it this way!

jstlaurent: I see! No worries. Class methods can be useful when you have inheritance. Children classes inherit the method and will get the their class passed in. So the parent class can implement generic logic that could work for all derived children.

Typing becomes a bit more complex. Before Python 3.11, you have to do something like this:
```python
from typing import Type, TypeVar

T = TypeVar('T', bound='Base')

class Base:
    def __init__(self, value: int) -> None:
        self.value = value

    @classmethod
    def create(cls: Type[T], value: int) -> T:
        return cls(value)

class Child(Base):
    def method_not_present_on_base(self) -> None:
        print("This method is only in Child")

base = Base.create(1)
child = Child.create(2)
child.method_not_present_on_base()  # This works correctly
```

Python 3.11 introduces a [Self](https://docs.python.org/3/library/typing.html#typing.Self) type that makes it cleaner:
```python
from typing import Self

class Base:
    def __init__(self, value: int) -> None:
        self.value = value

    @classmethod
    def create(cls) -> Self:
        return cls(42)

class Child(Base):
    def method_not_present_on_base(self) -> None:
        print("This method is only in Child")

base = Base.create()
child = Child.create()
child.method_not_present_on_base()  # This works correctly
```

Issue #131, opened 2024-07-17
Support zero-shot task splits


fteufel:
There are prediction problems that are often evaluated in a zero-shot scenario, with no data used for training. All data would be considered test data, and performance typically evaluated as a correlation efficient between the true values and a score from the model (e.g. perplexity).

Right now, polaris requires that there are at least two non-empty partitions specified in the `split` for `SingleTaskBenchmarkSpecification`. Could this be made more flexible, or is this out of scope for polaris?

Thanks!

cwognum:
Hey @fteufel,

Thanks a lot for the suggestion! It's an interesting proposal and I see how that would be useful. Technically, it shouldn't be too big a challenge either.

It does raise some more "philosophical" questions. One of the main goals we've set out to solve with Polaris is to make sure results are comparable. We want to be comparing methods and e.g. not datasets or preprocessing workflows. If we don't impose a fixed train set, it's not clear to me how we can make sure results remain comparable. For example, a method that is pre-trained on more data or data that comes from the same chemical space as the test set is expected to do better, but is that a meaningful difference _between methods_? Probably not? I'm not super familiar with the one-shot literature. Do you know if this is something that is addressed in a paper you could reference for me to read?

Or given the above, maybe we should not allow test-set only datasets on Polaris? Instead, we could ask people in a zero-shot setting to always explicitly define a possibly unsupervised (pre)train set? What do you think? In that case though, maybe we can think of other features that make defining such a benchmark easier, e.g. by allowing people to specify benchmarks based on two or more datasets?

Thanks again for the suggestion! Let me know what you think of this.

cwognum:
> people might use pretrained representations to build their supervised model on top of

That's a great point!

I think you're right that there is no way to fully prevent this and that may indeed not be as big a problem.

I do still think we should at some
point address the pre-training case. Maybe we could e.g. add a metadata field in the results object that lets you specify the precise data you pretrained on. Simply being transparent about this seems useful in comparing methods.

Thanks, @fteufel ! These are all interesting points!

fteufel:
That makes a lot of sense - definitely would be valuable to have pretraining info available. 

So, can we have zero-shot? 🙂

cwognum:
Yes! Do you want to make a PR?

PR #135, opened 2024-07-17
Allow an empty train partition to support test-set only (e.g. zero-shot) benchmarks

fteufel:
This PR enables the definition of zero-shot benchmarks by allowing the train partition of `BenchmarkSpecification` to be an empty list. See #131 

## Changelogs

- No more `InvalidBenchmarkError` is raised when `v[0]` is empty in `_validate_split`
- Error message and docstring was adjusted accordingly
---

_Checklist:_

- [x] _Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR._ #131 

---

### DIFF
--- a/polaris/benchmark/_base.py
+++ b/polaris/benchmark/_base.py
@@ -174,9 +174,8 @@ class BenchmarkSpecification(BaseArtifactModel, ChecksumMixin):
 
         # Train partition can be empty (zero-shot)
         # Test partitions cannot be empty
-        if (
-            (isinstance(v[1], dict) and any(len(v) == 0 for v in v[1].values()))
-            or (not isinstance(v[1], dict) and len(v[1]) == 0)
+        if (isinstance(v[1], dict) and any(len(v) == 0 for v in v[1].values())) or (
+            not isinstance(v[1], dict) and len(v[1]) == 0
         ):
             raise InvalidBenchmarkError("The predefined split contains empty test partitions")
 
--- a/polaris/benchmark/_base.py
+++ b/polaris/benchmark/_base.py
@@ -6,6 +6,7 @@ import fsspec
 import numpy as np
 import pandas as pd
 from datamol.utils import fs
+from loguru import logger
 from pydantic import (
     Field,
     ValidationInfo,
@@ -178,6 +179,11 @@ class BenchmarkSpecification(BaseArtifactModel, ChecksumMixin):
             not isinstance(v[1], dict) and len(v[1]) == 0
         ):
             raise InvalidBenchmarkError("The predefined split contains empty test partitions")
+
+        if len(v[0]) == 0:
+            logger.info(
+                "This benchmark only specifies a test set. It will return an empty train set in `get_train_test_split()`"
+            )
 
         train_indices = v[0]
         test_indices = [i for part in v[1].values() for i in part] if isinstance(v[1], dict) else v[1]
--- a/polaris/benchmark/_base.py
+++ b/polaris/benchmark/_base.py
@@ -166,19 +166,19 @@ class BenchmarkSpecification(BaseArtifactModel, ChecksumMixin):
     def _validate_split(cls, v, info: ValidationInfo):
         """
         Verifies that:
-          1) There is at least two, non-empty partitions
+          1) There are no empty test partitions
           2) All indices are valid given the dataset
           3) There is no duplicate indices in any of the sets
           3) There is no overlap between the train and test set
         """
 
-        # There is at least two, non-empty partitions
+        # Train partition can be empty (zero-shot)
+        # Test partitions cannot be empty
         if (
-            len(v[0]) == 0
-            or (isinstance(v[1], dict) and any(len(v) == 0 for v in v[1].values()))
+            (isinstance(v[1], dict) and any(len(v) == 0 for v in v[1].values()))
             or (not isinstance(v[1], dict) and len(v[1]) == 0)
         ):
-            raise InvalidBenchmarkError("The predefined split contains empty partitions")
+            raise InvalidBenchmarkError("The predefined split contains empty test partitions")
 
         train_indices = v[0]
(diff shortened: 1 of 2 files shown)


polaris/benchmark/_base.py
```diff
@@ -166,19 +166,18 @@ def _validate_main_metric(cls, v):
...
-          1) There is at least two, non-empty partitions
+          1) There are no empty test partitions
           2) All indices are valid given the dataset
           3) There is no duplicate indices in any of the sets
           3) There is no overlap between the train and test set
         """
 
-        # There is at least two, non-empty partitions
-        if (
-            len(v[0]) == 0
```
cwognum, line 177: **Suggestion**: I could see an empty train set be confusing, so for now I suggest we clearly log on benchmark creation that this is a test set only benchmark that will return an empty train set. E.g.:

```py
if len(v[0]) == 0: 
    logger.info("This benchmark only specifies a test set. It will return an empty train set in `get_train_test_split()`")
```

cwognum (commented):
Thanks @fteufel for the quick PR! 🙏 

Besides the minor comment that I left, could you also add a simple test case (e.g. another check in `test_split_verification`) making sure that creating a benchmark with an empty train partition works as expected. i.e. successfully creates a train Subset in `get_train_test_split()`, but that the returned object is empty?

fteufel:
I've added the case in `test_split_verification` - not completely sure about the logic in there though, doesn't look to me like the return values of `get_train_test_split` are checked there, just the instantiation.

cwognum:
Thanks again!

> doesn't look to me like the return values of get_train_test_split are checked there

That's right. You could add such a check, e.g. (didn't check if this works) 
```py
benchmark = cls(split=([], test_split), **default_kwargs)
train, _ = benchmark.get_train_test_split()
assert len(train) == 0
```

----

On another note, this PR will require a similar change to the web platform's API. This may take a while! I'll let you know once that has been completed.

fteufel:
That works test works :)

cwognum (approved):
The required changes on the Hub's side have been completed by @Andrewq11 (for internal reference, see https://github.com/polaris-hub/polaris-hub/pull/395). 

@fteufel Would you mind running a final test by trying to upload a test-set only benchmark to the Hub using the code in your branch? After that, this PR is ready to be merged. Congratulations on your first contribution ! 🚀

fteufel:
Will do - but can you tell me how to upload to an organization instead of an user? Can't find it in the docs.

cwognum:
Good to know that that's not clear. We will update the docs.

To answer your question: Simply set the owner parameter in the upload method to the slug of the organization.

E.g.
```py
dataset.upload_to_hub(owner="org")
```

fteufel:
```python
dataset.upload_to_hub(owner='mlls')
```
doesn't do it for me

```
Traceback (most recent call last):
  File "/novo/users/fegt/polaris/polaris/hub/client.py", line 161, in _base_request_to_hub
    response.raise_for_status()
  File "/novo/users/fegt/miniconda3/envs/bend/lib/python3.10/site-packages/httpx/_models.py", line 761, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '403 Forbidden' for url 'https://polarishub.io/api/v1/dataset/mlls/bend-variant-effects-expression'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/novo/users/fegt/BEND/scripts/upload_polaris_datasets.py", line 103, in <module>
    dataset.upload_to_hub(owner='mlls')
  File "/novo/users/fegt/polaris/polaris/dataset/_dataset.py", line 372, in upload_to_hub
    self.client.upload_dataset(self, access=access, owner=owner)
  File "/novo/users/fegt/polaris/polaris/hub/client.py", line 497, in upload_dataset
    response = self._base_request_to_hub(
  File "/novo/users/fegt/polaris/polaris/hub/client.py", line 172, in _base_request_to_hub
    raise PolarisHubError(
polaris.utils.errors.PolarisHubError: The request to the Polaris Hub failed. See the error message below for more details:
{
  "message": "Cannot create a dataset for owner 'mlls'"
}

```

cwognum:
Since the error thrown is a 403, it seems the user credentials you're making the API request with don't have access to the `mlls` organization. Looking at some logs, however, this seems strange: You made the request with an account that should have access to the `mlls` org... 

Could it be that the user credentials you're using were created before you created the organization? If so, you may need to refresh your credentials with: 
```
polaris login --overwrite
```

@jstlaurent It seems to me the JWTs don't include the most up-to-date membership information. I think this is expected, but I wonder if there could be a mechanism to invalidate a JWT if such updates happen. For reference, I was e.g. looking at [this trace](https://ui.honeycomb.io/valence-labs/environments/prod/result/bVJiNq84Jgn/trace/dRxY9SvCVSi?fields[]=s_name&fields[]=s_serviceName&span=49c9e2fbe28b77d0).

jstlaurent:
@cwognum : I believe you're correct. The token cached by the client contains the organization membership list. That allows for much lower latency when accessing the API, but I suspect in this case, that list has become stale.

I'll see how we can better balance this with the need to keep the data in sync.

cwognum:
Known issue! Fix will be deployed shortly!

cwognum:
> Known issue! Fix will be deployed shortly!

Sorry, this is taking a little longer than expected, but I hope to merge it today! 

In the meantime, this PR seems ready to be merged! Congratulations on your first contribution @fteufel and thanks for the great work! I will wait a little to release a new version, because I just released one and there are some other changes that I expect to be merged soon, but your work will make it into the next Polaris release! 🎉

cwognum:
@fteufel The fix has been deployed!

You're dataset should be ready now, which will also allow you to make the dataset public! If the issue persists, however, please file a new bug report (we have an issue template in this repo!).

Excited to see the BEND benchmark on Polaris soon! Thanks again! 🚀

PR #137, opened 2024-07-17
Only adding zarrRootPath field to datasets that use Zarr

Andrewq11:
## Changelogs

Updating the `client.upload_dataset` method to only add the `zarrRootPath` field to datasets which make use of Zarr. Otherwise, datasets which do not use make use of Zarr will fail a validation check and cannot be uploaded. 

---

_Checklist:_

~- [ ] _Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR._~

---

### DIFF
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -474,8 +474,9 @@ class PolarisHubClient(OAuth2Client):
         dataset.owner = HubOwner.normalize(owner or dataset.owner)
         dataset_json = dataset.model_dump(exclude={"cache_dir", "table"}, exclude_none=True, by_alias=True)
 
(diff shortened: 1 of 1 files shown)


cwognum (approved):

cwognum:
I'm okay with merging this to unblock everyone, but this piece of code feels brittle to me (my bad!).

Maybe we should instead improve the check on the Hub side to use the `usesZarr` field? This is already part of the payload anyways. And maybe the Hub should actually create and return a `zarrRootPath` to upload to? Something to think about!

Andrewq11:
I'll merge this and create a ticket to improve how we handle validation on the Hub side when it comes to Zarr datasets.

PR #140, opened 2024-07-18
Adding a loading indicator for upload/download operations

Andrewq11:
## Changelogs

- Using the `yaspin` [package](https://github.com/pavdmyt/yaspin), this PR adds a loading indicator to all client side upload/download operations. 
- The indicator will update to a success or failure status depending on the outcome of the given method that is running. This required wrapping numerous client methods in try/except blocks. 

---

_Checklist:_

- [x] _Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR. [Link to issue](https://github.com/orgs/polaris-hub/projects/2?pane=issue&itemId=70508095)_ 

---

NOTE: While the `tqdm` package is already a dependency of `polaris-lib`, the lift required to implement it given the current time crunch is too high. We should however revisit this if we want to have slightly fancier messaging for users. A ticket will be created for this shortly.


polaris/utils/progress_indicator.py
```diff
@@ -0,0 +1,26 @@
+from yaspin import yaspin
+from yaspin.spinners import Spinners
+
+
+class ProgressIndicator:
```
jstlaurent, line 5: **Suggestion**
It's fine as is, but it might be easier to use if it was written as a [context manager](https://realpython.com/python-with-statement/#creating-function-based-context-managers). Something like:
```python
@contextmanager
def progress_indicator(start_msg: str = 'In progress...', success_msg: str | None = None, error_msg: str | None = None):
    spinner = yaspin()
    spinner.spinner = Spinners.dots
    spinner.text = start_msg
    try:
        spinner.start()
        yield spinner
        spinner.color = "green"
        spinner.text = ""
        spinner.ok(f"✅ SUCCESS: \033[1m{success_msg}\033[0m\n")
    except Exception as err:
        spinner.color = "red"
        spinner.fail(f"💥 ERROR: {error_msg}")
        raise err
    finally:
        spinner.stop()
```

That would let you do this in the calling code:
```python
with progress_indicator("Uploading benchmark... ", "Your benchmark has been successfully uploaded to the Hub..."):
    ...do the work here...
```

jstlaurent (approved):
Nice! I had one suggestion, if you have the time.

polaris/utils/progress_indicator.py
Andrewq11: I'll give this a go, it looks cleaner 👍

mercuryseries (approved):
Thanks @Andrewq11. Looks great!
I added some minor suggestions on top of what Julien suggested.

polaris/utils/progress_indicator.py
mercuryseries: Some minor suggestions:

```python
@contextmanager
def progress_indicator(start_msg: str = 'In progress...', success_msg: str | None = None, error_msg: str | None = None):
    spinner = yaspin(Spinners.dots, text=start_msg)

    try:
        spinner.start()
        yield spinner
        spinner.text = ""
        spinner.green.ok(f"✅ SUCCESS: \033[1m{success_msg}\033[0m\n")
    except Exception as err:
        spinner.text = ""  # <-- Do we want to add this here too? 
        spinner.red.fail(f"💥 ERROR: {error_msg}")
        raise err
    finally:
        spinner.stop()
```

Andrewq11: Yes! Included in the latest commit.

polaris/utils/progress_indicator.py
```diff
@@ -0,0 +1,32 @@
...
+@contextmanager
+def progress_indicator(start_msg: str = "In progress...", success_msg: str = "", error_msg: str = ""):
+    spinner = yaspin()
+    spinner.spinner = Spinners.dots
+    spinner.text = start_msg
+    result_message = {"message": ""}
+
+    try:
+        spinner.start()
+        yield result_message
```
Andrewq11, line 15: Any thoughts on using this to pass a message back up to the context manager? This preserves the current flow we have for upload operations but might seem a little hacky(?) 

Let me know what you think 😄

mercuryseries: @Andrewq11 How about yielding a callback?

```python
@contextmanager
def progress_indicator(start_msg: str = 'In progress...', success_msg: str | None = None, error_msg: str | None = None):
    spinner = yaspin(Spinners.dots, text=start_msg)

    result_message = {"message": success_msg}

    def update_message(msg):
        result_message["message"] = msg

    try:
        spinner.start()
        yield update_message
        spinner.text = ""
        spinner.green.ok(f"✅ SUCCESS: \033[1m{result_message['message']}\033[0m\n")
    except Exception as err:
        spinner.text = "" 
        spinner.red.fail(f"💥 ERROR: {error_msg}")
        raise err
    finally:
        spinner.stop()


with progress_indicator("Starting task...", success_msg="Task completed!", error_msg="Task failed!") as update_message:
    # Simulate some work
    import time
    time.sleep(2)
    # raise Exception

    update_message("All steps were successfully executed.")
```

This keeps the logic inside the `progress_indicator` function, and you can use the `update_message` callback to set a custom success message.

Andrewq11: I like it! Will update.

polaris/utils/progress_indicator.py
cwognum, line 1: **Suggestion:** Minor, but I would move this to `polaris/utils/context.py`

polaris/utils/progress_indicator.py
cwognum: Returning a callback function from a context manager feels a little un-Pythonic to me. 

To offer an alternative: You could also implement the context manager as a class. In this case, you need to implement the `__enter()__` and `__exit__()` magic methods. The enter method could then return `self` and you can add the method to the class.

cwognum:
```py

class ProgressIndicator:
    def __init__(self, start_msg: str = 'In progress...', success_msg: str | None = None, error_msg: str | None = None):
        self._start_msg = start_msg
        self._success_msg = success_msg
        self._error_msg = error_msg

    def __enter__(self):
        spinner = yaspin(Spinners.dots, text=self._start_msg)
        spinner.start()
        return self

    def __exit__(self, exc_type, exc_value, traceback):
        spinner.text = "" 
        if exc_type is not None: 
            spinner.red.fail(f"💥 ERROR: {self._error_msg}")
        else: 
             spinner.green.ok(f"✅ SUCCESS: \033[1m{self._success_msg}\033[0m\n")
        spinner.stop()

    def update_success_msg(self, msg: str): 
        self._success_msg = msg


with ProgressIndicator("Starting task...", success_msg="Task completed!", error_msg="Task failed!") as indicator:
    indicator.update_success_msg("All steps were successfully executed.")
```

polaris/utils/progress_indicator.py
Andrewq11: Agreed. Will update.

polaris/hub/client.py
```diff
@@ -458,120 +490,133 @@ def upload_dataset(
...
+            # 3. Upload the associated Zarr archive
+            # TODO: Revert step 1 in case step 2 fails - Is this needed? Or should this be taken care of by the hub?
+
+            # Prepare the parquet file
+            buffer = BytesIO()
+            dataset.table.to_parquet(buffer, engine="auto")
+            parquet_size = len(buffer.getbuffer())
+            parquet_md5 = md5(buffer.getbuffer()).hexdigest()
+
+            # If the dataset uses Zarr, we will save the Zarr archive to the Hub as well
```
kiramclean, line 531: Is this duplicated? Looks like it appears on L515 as well.

polaris/utils/progress_indicator.py
```diff
@@ -0,0 +1,32 @@
+from contextlib import contextmanager
+from yaspin import yaspin
+from yaspin.spinners import Spinners
+
+
+@contextmanager
+def progress_indicator(start_msg: str = "In progress...", success_msg: str = "", error_msg: str = ""):
```
kiramclean, line 7: Would it make sense to have sensible defaults here (as opposed to empty strings)? Just something simple like `success_msg="Success", error_msg="Error"`?

polaris/utils/progress_indicator.py
```diff
@@ -0,0 +1,32 @@
...
+    spinner = yaspin()
+    spinner.spinner = Spinners.dots
+    spinner.text = start_msg
+
+    result_message = {"message": success_msg}
```
kiramclean, line 12: Out of curiosity, why is this a dict and not just a string?

polaris/hub/client.py
Andrewq11: Yes, I think it snuck in when resolving some merge conflicts. Great catch.

polaris/utils/progress_indicator.py
Andrewq11: I personally prefer the original in terms of readability, but that is likely because I'm not used to the pattern you suggested. I do agree that yielding a function doesn't seem very Pythonic though. If we feel strongly about this, I'm happy to implement it.

cwognum: My preference is the latter. Also would make it easy to add additional functionality to the context, such as `tqdm` integration or setting a custom error message. 

No strong preference though. We can add that complexity once it becomes needed as well.

Andrewq11: Let's do that then, not the biggest lift to implement it now.

polaris/utils/progress_indicator.py
Andrewq11: I put them as empty strings because the `spinner.ok` and `spinner.fail` methods already print out `SUCCESS` and `FAIL`. 

I think it might actually be better to not make these arguments optional as every call of `progress_indicator` in the client defines them (and they should for clarity).

cwognum (approved):
Thanks, @Andrewq11 ! Great UX improvement!

PR #144, opened 2024-07-19
Adding more informative error messaging due to potential JWT and Clerk data discrepancy 

Andrewq11:
## Changelogs

Due to the potential discrepancy between claims in JWTs issued from the Hub and the true data in Clerk, we want to provide users with a more informative error message in the event of forbidden errors (for upload) and not found errors (for downloads). More specifically, we would like to suggest to the user to retry logging in to get a fresh JWT with up-to-date claims from Clerk.

- Added error messaging suggesting a re-login in the case of 403 forbidden errors when a user tries to upload a dataset, benchmark or result. 
- Added error messaging suggesting a re-login in the case of 404 not found errors when a user tries to retrieve a dataset or benchmark. 
- Updated the doc strings for the above client methods with a helpful note regarding the potential discrepancy.

---

_Checklist:_

    - This came about from another PR: https://github.com/polaris-hub/polaris/pull/135

---


polaris/hub/client.py
```diff
@@ -153,22 +154,28 @@ def _load_from_signed_url(self, url: URLTypes, load_fn: Callable, headers: Heade
...
-            if response.status_code == 500:
+            if response_status_code == 500:
                 raise
 
             # If not an internal server error, the hub should always return a JSON response
             # with additional information about the error.
             response = response.json()
             response = json.dumps(response, indent=2, sort_keys=True)
+
+            if show_informative_error and (response_status_code == 403 or response_status_code == 404):
```
Andrewq11, line 176: I'm not sure if we want the `response_status_code == 404` here. I added it because users facing this data discrepancy issue could run into 404s despite them having the correct permissions. 

So in that case, it seems suitable here, but I wanted to gather some additional opinions on it.

kiramclean: When does the user get a `404` when they shouldn't? When the user has gained a membership in Clerk that isn't reflected in the JWT they'd get a `403`, right?

cwognum: > When the user has gained a membership in Clerk that isn't reflected in the JWT they'd get a 403, right?

When retrieving datasets you don't have access to, you would get a 404. So if you just joined an organization and try download a private dataset, that could result in a 404 when the JWT is outdated.

cwognum (commented):
Thanks for the initiative, @Andrewq11 ! That is a nice quality of life improvement.

polaris/hub/client.py
```diff
@@ -153,22 +154,28 @@ def _load_from_signed_url(self, url: URLTypes, load_fn: Callable, headers: Heade
...
+                raise InformativePolarisHubError(response, method)
```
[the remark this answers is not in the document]
cwognum: Ah, I just realized this was probably copied from [here](https://github.com/polaris-hub/polaris/blob/main/polaris/hub/client.py#L173-L175). I believe I may have written that code, but looking at it now I would drop the response from there as well.

polaris/hub/client.py
```diff
@@ -153,22 +154,28 @@ def _load_from_signed_url(self, url: URLTypes, load_fn: Callable, headers: Heade
         content = BytesIO(response.content)
         return load_fn(content)
 
-    def _base_request_to_hub(self, url: str, method: str, **kwargs):
+    def _base_request_to_hub(self, url: str, method: str, show_informative_error: bool = False, **kwargs):
```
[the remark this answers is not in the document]
Andrewq11: I added this early on when I was under the impression that only some client methods would be impacted by this change. Not the case though, good catch!

This does make me think though, the list operations will remain silently impacted by this data discrepancy issue. Since they do not throw 403s and they instead just don't return names of artifacts the user doesn't have access to, we can't intervene (as is) and suggest a re-login. 

While we made other changes to mitigate this (e.g. TTL), it is something we may want to consider in the near future.

polaris/utils/errors.py
```diff
@@ -37,3 +42,24 @@ class TestAccessError(Exception):
...
+
+    def __init__(self, response: Response, method: str):
+        suggestion_message = (
+            self.SUGGEST_LOGIN_ON_CREATE
+            if method == "PUT" or method == "POST"
+            else self.SUGGEST_LOGIN_ON_RETRIEVE
+        )
+        suggestion_message = format_text(suggestion_message, [self.BOLD_CODE, self.YELLOW_CODE])
+
+        error_message = f"The request to the Polaris Hub failed. See the error message below for more details:\n{response}\n\n{suggestion_message}."
```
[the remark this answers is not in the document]
Andrewq11: I think it generally makes sense to, but the only conflicting thing is that the PolarisHubError is thrown for 409s when business rules are violated. I think it would be slightly misleading/confusing to show a message calling that an "issue" and to reach out to us for it. 

What do you think?

polaris/_mixins.py
```diff
@@ -4,7 +4,9 @@
 from loguru import logger
 from pydantic import BaseModel, PrivateAttr, computed_field
 
-from polaris.utils.errors import PolarisChecksumError
+
+class PolarisChecksumError(ValueError):
```
Andrewq11, line 8: Because we are using the `FormattingMixin` in the errors.py file but also using the `PolarisChecksumError` in this file, we run into issues with circular imports. 

Having `PolarisChecksumError` defined here doesn't feel correct, I've just done it to get past errors. We could solve this by creating a `mixins` folder and defining each mixin in an individual file. What do we think of that?

Andrewq11: I've converted the PR to a draft until we decide what to do here.

polaris/hub/client.py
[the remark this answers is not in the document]
cwognum: Good point. I don't think it will be too big an issue (simply trying again a few minutes later would solve it), but it could indeed be confusing. 

More generally, with these PRs I've been thinking a bit more about how we can streamline error handling. I think there is a lot to gain here, e.g. also by parsing the Zod error message the Hub returns in case of a 409. At the summer school, I felt that a lot of questions arose due to people not understanding the stack trace.

polaris/_mixins.py
cwognum: > We could solve this by creating a mixins folder and defining each mixin in an individual file

That works!

An alternative is to move the import of the `PolarisChecksumError` inside of `FormattingMixin.format()`.

polaris/_mixins.py
```diff
@@ -69,3 +71,17 @@ def verify_checksum(self, md5sum: str | None = None):
             raise PolarisChecksumError(
                 f"The specified checksum {md5sum} does not match the computed checksum {self.md5sum}"
             )
+
+
+class FormattingMixin:
```
cwognum, line 76: **Suggestion**: Now that we have this, let's use it for the `ProgressIndicator` as well!

polaris/hub/client.py
```diff
@@ -162,17 +169,24 @@ def _base_request_to_hub(self, url: str, method: str, **kwargs):
...
             response = json.dumps(response, indent=2, sort_keys=True)
-            raise PolarisHubError(
-                f"The request to the Polaris Hub failed. See the error message below for more details:\n{response}"
-            ) from error
+
+            if response_status_code == 403:
+                raise PolarisCreateArtifactError() from error
+
+            if response_status_code == 404:
+                raise PolarisRetrieveArtifactError() from error
```
cwognum, line 187: **Suggestion:** As we've seen in this PR, I think the mapping between HTTP status codes and actions (i.e. create / retrieve) is not self-explanatory. I would add a comment briefly documenting this. We may need to revisit this in the future as the Hub and Client evolve (e.g. if we want to support updating artifacts through the client).

polaris/utils/errors.py
```diff
@@ -10,12 +13,12 @@ class InvalidResultError(ValueError):
     pass
 
 
-class PolarisChecksumError(ValueError):
-    pass
-
+class PolarisHubError(Exception, FormattingMixin):
+    def __str__(self):
+        response_message = self.__cause__.response.text
```
cwognum, line 18: **Suggestion:** This feels brittle. There are no guarantees that the cause of the `PolarisHubError` has the `response` field. If you would like to show the `response.text` at the bottom of the stack trace, I would change it back such that you explicitly pass in the `response`. 

Maybe move it to the `PolarisHubError`? I think it makes sense to always require a `response` for these errors! 

```py
class PolarisHubError(Exception, FormattingMixin):
    def __init__(self, message: str, response: Response):
        super().__init__(message)
        self.response = response
```

polaris/utils/errors.py
```diff
@@ -37,3 +40,21 @@ class TestAccessError(Exception):
 
 class InvalidZarrChecksum(Exception):
     pass
+
+
+class PolarisCreateArtifactError(PolarisHubError):
+    SUGGEST_LOGIN_ON_CREATE = "Note: If you can confirm that you are authorized to perform this action, please call 'polaris login --overwrite' and try again. If this issue persists, please reach out to the Polaris team for support"
+
+    def __str__(self):
+        base_message = super().__str__()
```
cwognum, line 49: **Suggestion:** Instead of the `super.__str__()`, I think we should use the `__init__()` constructors. That way we benefit from any other functionality that may be based on `self.message`. I also don't think we need to define the messages as class variables, since we only use them in one spot. 

Something like: 

```py
class PolarisCreateArtifactError(PolarisHubError):
    def __init__(self): 
        message = self.format(
            "Note: If you can confirm that you are authorized to perform this action, " 
            "please call 'polaris login --overwrite' and try again.",
            [self.BOLD, self.YELLOW]
        )
        super().__init__(message)
```

In `PolarisHubError`, you could similarly add a prefix (and suffix) with the same approach.

cwognum (commented):
Thanks again, @Andrewq11 ! Throwing custom exceptions in Python is not something I've really done before, so I'm learning with you here. I feel like the current approach is not very Pythonic and did some further suggestions that may help in that regard. Curious to hear what you think! Appreciate your patience.

I'm curious how other libs do this, e.g. numpy. Maybe there is some inspiration to be found there...

polaris/utils/errors.py
Andrewq11: Did I misinterpret what you meant in these previous comments? That's why I went down this path.


Screenshot 2024-07-22 at 7 45 15 AM

polaris/utils/errors.py
Andrewq11: We use the `super.__str__()` because I thought we didn't want to explicitely pass in a `response` object. We need to use the `__str__` method in that case because the `self.__cause__` variable is not defined until `__init__` has run.

Andrewq11: Also, at least from the research I've done, it's not discouraged to use the `__str__ ` method to format the error message you'd like to display when the exception is called. When you pass in a message to `super().__init__(message)`, that message is just stored and later outputted by the default `__str__` method.

polaris/utils/errors.py
cwognum: My bad! I was thinking that by raising from another error, we could drop the `response.text` from the message in the error we throw (reason being that can be found in the stack trace now). 

However, thinking about this again, I prefer us adding the `response` parameter back to the `PolarisHubError` as you had it earlier. 
- Having actionable feedback at the end of the stack trace, rather than hiding it in the stack trace, is way more user friendly!
- I also like that this is adding value to the custom class we're introducing (e.g. we could now also more easily reformat the Zod errors in the JSON body) and that it makes the expected context in which you can throw such an error explicit, i.e. a `PolarisHubError` should only be thrown in the context of a request and response in the client!

polaris/utils/errors.py
cwognum: So if we add back the `response`, this no longer is an issue? 

> When you pass in a message to super().__init__(message), that message is just stored and later outputted by the default __str__ method.

This is true now, but may change. E.g. the `PolarisHubError` may at some point add custom logic using the `message` variable. It think using the constructor is the more appropriate way of doing this.

polaris/utils/errors.py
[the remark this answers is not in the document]
cwognum: I see your point!

I do think mentioning a default communication channel people can reach us at for any issues relating to the Hub (which is not open source and over which users thus have less control) could nevertheless be useful. For example, the Zod errors were confusing to a lot of users at the Summer School and as long as we haven't improved the formatting of these errors, I would expect people to get stuck on these.

I would be okay with something like: 
```
If the issue persists, please reach out to the Polaris team via https://github.com/polaris-hub/polaris.
```

But feel free to rephrase or to leave it as is for now!

polaris/utils/errors.py
Andrewq11: This is something we should have by Thursday given the TTL increase, and my time is thin given where this ranks in the priority stack, so let's confirm some things to prevent more confusion:
- We want each error to have a `response` parameter 
- We want to only use the `__init__` method, not the `__str__` method
- We have no use in raising errors with the `from` keyword, and we can raise it like this `raise PolarisHubError("some message")` instead

Is this correct?

cwognum:
@Andrewq11 With me being OOO, the time difference and the fast approaching deadline, I decide to make a PR into your branch to make my suggestions very explicit: https://github.com/polaris-hub/polaris/pull/153.

polaris/utils/errors.py
```diff
@@ -10,30 +15,56 @@ class InvalidResultError(ValueError):
...
+        super().__init__(message, response)
+
+
+class PolarisRetrieveArtifactError(PolarisHubError):
+    def __init__(self, response: Response | None = None):
+        message = (
+            "Note: If this artifact exists and you can confirm that you are authorized to retrieve it, "
+            "please call 'polaris login --overwrite' and try again. If the issue persists, please reach out to the Polaris team for support."
+        )
+        message = self.format(message, [self.BOLD, self.YELLOW])
```
mercuryseries, line 69: Not required for now, but adding helper methods for frequent formats could improve readability and maintainability.

polaris/benchmark/_base.py
```diff
@@ -18,7 +18,7 @@
 from sklearn.utils.multiclass import type_of_target
 
 from polaris._artifact import BaseArtifactModel
-from polaris._mixins import ChecksumMixin
+from polaris.mixins._checksum import ChecksumMixin
```
mercuryseries, line 21: To maintain consistency with the current codebase, we could create a `polaris/mixins/__init__.py` file with the following content:

```python
# Import classes from the internal modules
from polaris.mixins._checksum import ChecksumMixin
from polaris.mixins._format_text import FormattingMixin

# Define what is available for import when the package is imported
__all__ = ["ChecksumMixin", "FormattingMixin"]
```

This will allow us to do:

```python
from polaris.mixins import ChecksumMixin
from polaris.mixins import FormattingMixin
```

mercuryseries (approved):
Thanks @Andrewq11! I added some minor comments/suggestions.

polaris/benchmark/_base.py
Andrewq11: Good call, thank you!

PR #146, opened 2024-07-21
Fix: `PolarisFileSystem` uses the dataset name instead of the dataset slug, causing a 404

zhu0619:
## Changelogs
- Sluggify the dataset name in the `PolarisFileSystem`.
- Serialize the `cache_dir` to a string.
- Updated Zarr dataset tutorial to reflect recent changes.
---

_Checklist:_

<s> - [ ] _Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR._ </s>
<s>- [ ] _Add tests to cover the fixed bug(s) or the newly introduced feature(s) (if appropriate)._</s>
<s>- [ ] _Update the API documentation if a new function is added, or an existing one is deleted._</s>

---

When uploading the dataset, the dataset slug has to be used to get the correct path to the R2 bucket. Otherwise, the Hub returns a `404`.


polaris/dataset/_dataset.py
```diff
@@ -516,8 +516,8 @@ def __getitem__(self, item):
 
     def _repr_dict_(self) -> dict:
         """Utility function for pretty-printing to the command line and jupyter notebooks"""
-        repr_dict = self.model_dump()
-        repr_dict.pop("table")
+        repr_dict = self.model_dump(exclude={"table"})
+        repr_dict["cache_dir"] = str(repr_dict["cache_dir"])
```
cwognum, line 520: **Suggestion:** Add a [`@field_serializer`](https://docs.pydantic.dev/latest/api/functional_serializers/#pydantic.functional_serializers.field_serializer) for the `cache_dir` to the `Dataset` class. That way, we centralize the logic for serialization.

polaris/dataset/_dataset.py
```diff
@@ -516,8 +516,8 @@ def __getitem__(self, item):
 
     def _repr_dict_(self) -> dict:
         """Utility function for pretty-printing to the command line and jupyter notebooks"""
-        repr_dict = self.model_dump()
-        repr_dict.pop("table")
+        repr_dict = self.model_dump(exclude={"table"})
```
cwognum, line 519: **Suggestion:** Let's exclude the checksum manifest as well. This doesn't hold information that the user needs and can get quite lengthy.

```suggestion
        repr_dict = self.model_dump(exclude={"table", "zarr_md5sum_manifest"})
```

polaris/hub/client.py
```diff
@@ -320,7 +320,7 @@ def open_zarr_file(
         polaris_fs = PolarisFileSystem(
             polaris_client=self,
             dataset_owner=owner,
-            dataset_name=name,
+            dataset_name=sluggify(name),
```
cwognum, line 323: **Suggestion:** This is a good catch! I would move it inside the `PolarisFileSystem` constructor, however, to centralize the logic.

docs/tutorials/dataset_zarr.ipynb
cwognum, line 1: **Suggestion:** Nice! Thank you for updating this! 🙏 

Just one minor suggestion:  `# consolidate` is not very informative as a comment. I would explain why we consolidate the archive (in short: Polaris requires this because it is more efficient when loading the data from a cloud bucket).

Could you also run the notebook again after excluding the manifest from the dataset repr?
-

cwognum (approved):
Thanks @zhu0619 ! Just some minor suggestions.

docs/tutorials/dataset_zarr.ipynb
zhu0619: Yep.  Updated the notebook. The manifest is no longer in the dataset viz table.

cwognum:
Hey @zhu0619 ! Seems I'm too late, but one more thought came to mind when I saw you push the latest changes: Do we need to sluggify the owner too? I don't think so, because even though we type-hint `dataset_owner` as a `str`, I believe it's actually a [`HubOwner`](https://github.com/polaris-hub/polaris/blob/5cf6092615a49af08e7a88b8d1b44c2b0af1574e/polaris/utils/types.py#L116-L138) object which returns its slug as string representation. 

Maybe we should just drop the `dataset_name` and `dataset_owner` and use the [`artifact_id`](https://github.com/polaris-hub/polaris/blob/5cf6092615a49af08e7a88b8d1b44c2b0af1574e/polaris/_artifact.py#L52-L55) instead. It's only used for the prefix anyways.

Issue #127, opened 2024-07-10
Optimize field validator for the `split` field in the `BenchmarkSpecification` class


Andrewq11:
# Context

Currently, the `split` field validator runs a series of iterative checks to determine whether a user has specified valid train and test rows. These checks work fine for datasets which contain a reasonable number of rows (thousands), but begin to take exceedingly long when run on datasets with millions of rows. 

For context, a `Benchmark` object containing a dataset with 200,000,000 rows took greater than 30 minutes to initialize due to this validation step. When this step was removed, it took less than 5 minutes.

# Description

This validation step must be optimized to reduce the time it takes to complete. This will likely come in the form of ensuring that all current checks which are in place are required, refactoring existing checks with a focus on removing redundant work and improving time complexity, and also deciding whether there are different business rules we can use to judge the validity of the `split` field.

# Acceptance Criteria

- Time to init a benchmark object takes < 10 minutes with a dataset containing 200,000,000 rows

# Links

- [Link to validation step](https://github.com/polaris-hub/polaris/blob/main/polaris/benchmark/_base.py#L167)

jstlaurent:
I have a strong suspicion that this is the worst offender:
```python
# The train and test indices do not overlap
if any(i in train_indices for i in test_indices):
    raise InvalidBenchmarkError("The predefined split specifies overlapping train and test sets")
```

Maybe building the sets of indices first, and then using them would yield better perf. Something like:
```python
@field_validator("split")
def _validate_split(cls, v: SplitType, info: ValidationInfo):
    """
    Verifies that:
      1) There is at least two, non-empty partitions
      2) All indices are valid given the dataset
      3) There is no duplicate indices in any of the sets
      3) There is no overlap between the train and test set
    """
    train_idx_list = v[0]
    test_idx_list = list(i for part in v[1].values() for i in part) if isinstance(v[1], dict) else v[1]

    # There is at least two, non-empty partitions
    if len(train_idx_list) == 0 or len(test_idx_list) == 0:
        raise InvalidBenchmarkError("The predefined split contains empty partitions")
    
    train_idx_set = set(train_idx_list)
    test_idx_set = set(test_idx_list)

    # The train and test indices do not overlap
    if len(train_idx_set & test_idx_set) > 0:
        raise InvalidBenchmarkError("The predefined split specifies overlapping train and test sets")

    # Duplicate indices
    if len(train_idx_set) != len(train_idx_list):
[...]

PR #148, opened 2024-07-22
optimizing time complexity of split field validator

Andrewq11:
## Changelogs

- Updates to the `split` field validator in the `Benchmark` model which significantly improve the validators time complexity (thank you for the suggestions and analysis, @jstlaurent and @cwognum)

---

_Checklist:_

- [x] _Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR._ https://github.com/polaris-hub/polaris/issues/127

---

### DIFF
--- a/polaris/benchmark/_base.py
+++ b/polaris/benchmark/_base.py
@@ -204,8 +204,8 @@ class BenchmarkSpecification(BaseArtifactModel, ChecksumMixin):
 
         # All indices are valid given the dataset
         if info.data["dataset"] is not None:
-            max = len(info.data["dataset"])
-            if any(i < 0 or i >= max for i in chain(train_idx_list, test_idx_list)):
+            max_i = len(info.data["dataset"])
+            if any(i < 0 or i >= max_i for i in chain(train_idx_list, test_idx_list)):
                 raise InvalidBenchmarkError("The predefined split contains invalid indices")
 
         return v
--- a/polaris/benchmark/_base.py
+++ b/polaris/benchmark/_base.py
@@ -1,3 +1,4 @@
+from itertools import chain
 import json
 from hashlib import md5
 from typing import Any, Callable, Optional, Union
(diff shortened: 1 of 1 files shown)


polaris/benchmark/_base.py
```diff
@@ -180,28 +181,33 @@ def _validate_split(cls, v, info: ValidationInfo):
...
+        if len(train_idx_set) != len(train_idx_list):
             raise InvalidBenchmarkError("The training set contains duplicate indices")
-        if len(set(test_indices)) != len(test_indices):
+        if len(test_idx_set) != len(test_idx_list):
             raise InvalidBenchmarkError("The test set contains duplicate indices")
 
         # All indices are valid given the dataset
         if info.data["dataset"] is not None:
-            if any(i < 0 or i >= len(info.data["dataset"]) for i in train_indices + test_indices):
+            max = len(info.data["dataset"])
```
cwognum, line 207: **Suggestion:** I would rename this variable.

```suggestion
            max_i = len(info.data["dataset"])
```


Naming a variable `max` overwrites the built-in `max()` function. E.g.:

```
>>> max(1, 3)
3
>>> max = 3
>>> max(1, 3)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'int' object is not callable
```

cwognum (approved):
Thanks @Andrewq11 ! 

Did you try splitting a large dataset again? How long did that take?

polaris/benchmark/_base.py
Andrewq11: Good catch!

Andrewq11:
> Did you try splitting a large dataset again? How long did that take?

@cwognum I did try splitting a large dataset! I believe 200,000,000 rows took ~20 seconds and 100,000,000 took ~8 seconds. Much, much better.

It would be great to have more logging for when these validations are happening too. Just something to let the user know that we are running validations which can take some time (10-20 seconds of "hanging" could be confusing for a user). What do you think? Happy to create an issue.

cwognum:
Yeah, that makes sense to me. Not super high priority because I don't expect such large datasets any time soon, but improving the logging is an important, low hanging fruit to improve the UX!

PR #153, opened 2024-07-23
Suggestions for PR #144

cwognum:
Opening this PR to clearly communicate my suggestions for #144.

I feel strongly about setting the message using the constructor. In the future, we may want to add custom functionality to parent classes based on the `message` field, e.g. we may want to add observability in the parent class that logs the exception or we may want to use some global formatting.

> [!WARNING]
> I did not have the time to extensively test the changes!


polaris/_mixins.py
```diff
@@ -68,6 +64,9 @@ def verify_checksum(self, md5sum: str | None = None):
         self.md5sum = self._compute_checksum()
 
         if self.md5sum != md5sum:
+            # Import here to prevent circular imports
+            from polaris.utils.errors import PolarisChecksumError
```
Andrewq11, line 68: I think we should still create the `mixins` folder unless we have a strong preference not to. Inline imports can get messy if the codebase grows and we become ok with doing this elsewhere.

polaris/utils/errors.py
```diff
@@ -13,48 +15,59 @@ class InvalidResultError(ValueError):
...
-        return f"{base_message}\n{suggestion_message}."
+        if response is not None:
+            prefix += f" The Hub responded with: {response.text}."
 
+        suffix = "If the issue persists, please reach out to the Polaris team for support."
 
-class PolarisRetrieveArtifactError(PolarisHubError):
-    SUGGEST_LOGIN_ON_RETRIEVE = "Note: If this artifact exists and you can confirm that you are authorized to retrieve it, please call 'polaris login --overwrite' and try again. If this issue persists, please reach out to the Polaris team for support"
+        super().__init__("\n".join([prefix, message, suffix]))
+        self.response = response
```
Andrewq11, line 43: Going to remove this, it's unused

polaris/utils/errors.py
```diff
@@ -13,48 +15,59 @@ class InvalidResultError(ValueError):
...
+        prefix = "The request to the Polaris Hub failed."
 
-    def __str__(self):
-        base_message = super().__str__()
-        suggestion_message = self.format(self.SUGGEST_LOGIN_ON_CREATE, [self.BOLD, self.YELLOW])
-        return f"{base_message}\n{suggestion_message}."
+        if response is not None:
+            prefix += f" The Hub responded with: {response.text}."
 
+        suffix = "If the issue persists, please reach out to the Polaris team for support."
```
Andrewq11, line 40: This seems out of place with certain errors (e.g. invalid auth credentials).

polaris/utils/errors.py
```diff
@@ -13,48 +15,59 @@ class InvalidResultError(ValueError):
...
+
+
 class InvalidZarrChecksum(Exception):
     pass
 
 
-class PolarisCreateArtifactError(PolarisHubError):
-    SUGGEST_LOGIN_ON_CREATE = "Note: If you can confirm that you are authorized to perform this action, please call 'polaris login --overwrite' and try again. If this issue persists, please reach out to the Polaris team for support"
+class PolarisHubError(Exception, FormattingMixin):
+    def __init__(self, message: str, response: Response | None = None):
```
Andrewq11, line 34: We won't always have a message, so we should make this optional

polaris/hub/client.py
```diff
@@ -295,7 +299,9 @@ def get_dataset(
                 try:
                     storage_response.raise_for_status()
                 except HTTPStatusError as error:
-                    raise PolarisHubError("Could not get signed URL from Polaris Hub.") from error
+                    raise PolarisHubError(
+                        "Could not get signed URL from Polaris Hub.", response=storage_response
```
Andrewq11, line 303: We should name these arguments for consistency

Andrewq11 (commented):
Thank you for this. 

These comments are for me. I'll address them either today or tomorrow, test everything, and then handle merging everything in.

Issue #149, opened 2024-07-22
Misleading exception thrown when Hub error doesn't contain a JSON body


fteufel:
### Polaris version

0.7.3

### Python Version

3.10

### Operating System

Linux

### Installation

pip

### Description

A seemingly standard benchmark upload fails. Not sure why exactly.

```
⠋ Uploading benchmark...2024-07-22 16:49:03.027 | INFO     | polaris._mixins:md5sum:27 - Computing the checksum. This can be slow for large datasets.
💥 ERROR: Failed to upload benchmark. 
Traceback (most recent call last):
  File "/novo/users/fegt/BEND/scripts/upload_polaris_benchmarks.py", line 62, in <module>
    benchmark.upload_to_hub(owner='mlls', visibility='public')
  File "/novo/users/fegt/miniconda3/envs/bend/lib/python3.10/site-packages/polaris/benchmark/_base.py", line 499, in upload_to_hub
    return client.upload_benchmark(self, access=access, owner=owner)
  File "/novo/users/fegt/miniconda3/envs/bend/lib/python3.10/site-packages/polaris/hub/client.py", line 657, in upload_benchmark
    response = self._base_request_to_hub(url=url, method="PUT", json=benchmark_json)
  File "/novo/users/fegt/miniconda3/envs/bend/lib/python3.10/site-packages/polaris/hub/client.py", line 171, in _base_request_to_hub
    response = response.json()
  File "/novo/users/fegt/miniconda3/envs/bend/lib/python3.10/site-packages/httpx/_models.py", line 764, in json
    return jsonlib.loads(self.content, **kwargs)
  File "/novo/users/fegt/miniconda3/envs/bend/lib/python3.10/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/novo/users/fegt/miniconda3/envs/bend/lib/python3.10/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/novo/users/fegt/miniconda3/envs/bend/lib/python3.10/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

### Steps to reproduce

```python
dataset = po.load_dataset("mlls/bend-histone-modification")
dataset.cache()
split = [
[...]

PR #155, opened 2024-07-23
Updating `base_request_to_hub` method to allow handling of responses without json

Andrewq11:
## Changelogs

- Wrapped the JSON parsing of an HTTP response's body in a `try...except` block. This prevents a `JSONDecodeError` in the event the response does not include JSON. This allows the error to still be handled in a meaningful manner. Errors returned from the Hub should be unaffected and output in the same fashion. 

---

_Checklist:_

- [x] _Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR._ Yes: https://github.com/polaris-hub/polaris/issues/149

---

### DIFF
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -175,8 +175,8 @@ class PolarisHubClient(OAuth2Client):
                 raise
 
             # If JSON is included in the response body, we retrieve it and format it for output. If not, we fallback to
-            # retrieving plain text from the body. This is important for handling large payloads that our backend may reject,
-            # resulting in a non-JSON error response.
+            # retrieving plain text from the body. This is important for handling certain errors thrown from the backend
+            # which do not contain JSON in the response.
             try:
                 response = response.json()
                 response = json.dumps(response, indent=2, sort_keys=True)
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -174,14 +174,17 @@ class PolarisHubClient(OAuth2Client):
             if response_status_code == 500:
                 raise
 
-            # If not an internal server error, the hub should always return a JSON response
-            # with additional information about the error.
-            response = response.json()
-            response = json.dumps(response, indent=2, sort_keys=True)
+            # If JSON is included in the response body, we retrieve it and format it for output. If not, we retrieve
+            # the string representation of the response body and move forward
+            try:
+                response = response.json()
+                response = json.dumps(response, indent=2, sort_keys=True)
+            except json.JSONDecodeError:
+                response = response.text
+                pass
 
             # The below two error cases can happen due to the JWT token containing outdated information.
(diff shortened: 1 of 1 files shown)


polaris/hub/client.py
```diff
@@ -174,14 +174,16 @@ def _base_request_to_hub(self, url: str, method: str, **kwargs):
...
-            # If not an internal server error, the hub should always return a JSON response
-            # with additional information about the error.
-            response = response.json()
-            response = json.dumps(response, indent=2, sort_keys=True)
+            # If JSON is included in the response body, we retrieve it and format it for output. If not, we retrieve
+            # the string representation of the response body and move forward
+            try:
+                response = response.json()
+                response = json.dumps(response, indent=2, sort_keys=True)
+            except (json.JSONDecodeError, TypeError):
```
Andrewq11, line 182: Given that the Hub should always return valid JSON, we shouldn't need to worry about a `TypeError` in the case `json.dumps(...)` fails. However, I added it here to be sure. I also like the readability aspect of directly specifying which errors _could_ happen.

Minor, but let me know if anyone has thoughts.

mercuryseries: Out of curiosity – What prompted this change?

Andrewq11: If a user defines a benchmark split with >150,000 rows and tries to upload it to the hub, our infra provider will deny the request because the payload is too large. 

In their response, there will be no JSON and so the user will get a very confusing error message related to a JSONDecodeError. This change prevents that.

mercuryseries: I see. How about we add a comment for future reference to make it clear why we're handling these exceptions?

```python
try:
    response_json = response.json()
    response = json.dumps(response_json, indent=2, sort_keys=True)
except (json.JSONDecodeError, TypeError):
    # If the response is not valid JSON or if json.dumps fails, fallback to plain text.
    # This is important for handling large payloads that our infra provider may reject,
    # resulting in a non-JSON error response.
    response = response.text
```

mercuryseries (approved):

polaris/hub/client.py
Andrewq11: You don't think the comment on lines 177-178 is descriptive enough?

I'm also not sure if mentioning the limitations of our infra provider in the codebase is advisable. What do you think? 

If we feel strongly, I'll add something a bit more descriptive.

mercuryseries: > You don't think the comment on lines 177-178 is descriptive enough?
The comment on lines 177-178 is helpful, but I felt it lacked some context, which is why I asked about the reason behind the changes.

> I'm also not sure if mentioning the limitations of our infra provider in the codebase is advisable. What do you think?
I agree.

As long as we remember why we added this, that's what matters most.

PR #167, opened 2024-07-29
Adding correct name of package during version fetch

Andrewq11:
## Changelogs

- Updating the name of the package for which we fetch the client version from. It is currently trying to look for `polaris` when it should be looking for `polaris-lib`. This is needed because all incoming bug reports have the version `dev` instead of the actual release version.

---

_Checklist:_

---


mercuryseries (approved):
Good catch! Thanks @Andrewq11!

PR #171, opened 2024-08-09
Add a converter from PDB to Zarr to the DatasetFactory

zhu0619:
## Changelogs

- Added `PDBConverter` which is able to convert pdb files to zarr file
- Added `ARRAY_TO_PDB` which is able to load pdb from zarr file
- Allow `add_from_file` to handle multiple files. 
- Added simple tutorial for creating dataset from pdb file. 
---

_Checklist:_

---
Issue #172 

During the conversion, only the most essential structural information is retained, including 3D coordinates, chain ID, residue ID, insertion code, residue name, heteroatom indicator, atom name, element, atom ID, B-factor, occupancy, and charge.


zhu0619:
Currently, fastpdb can only be installed via pip. I'll create an issue in their repository to request support for conda installation.

polaris/dataset/_dataset.py
```diff
@@ -485,7 +485,9 @@ def _split_index_from_path(self, path: str) -> Tuple[str, Optional[int]]:
             index = index.split(":")
 
             if len(index) == 1:
-                index = int(index[0])
+                index = index[0]
+                if index.isdigit():
+                    index = int(index)
```
cwognum, line 490: **Suggestion**:  I found this piece of code a little confusing. 

After looking at it for a bit, I realize it's because currently the "pointer" looks something like: 
```
pdb#tutorial
```

Whereas based on the name of the variables, I would expect
```
pdb/tutorial/
```

We should only use `#` when we want to index an array, not when we want to access a group.

I actually think that without these lines of code, it should work with the path I suggested.

polaris/dataset/converters/_pdb.py
```diff
@@ -0,0 +1,158 @@
...
+    "b_factor",
+    "occupancy",
+    "charge",
+]
+
+
+class PDBConverter(Converter):
+    """
+    Converts PDB files into a Polaris dataset.
+    During the conversion, only the most essential structural information is retained, including 3D coordinates, chain ID, residue ID, insertion code, residue name, heteroatom indicator, atom name, element, atom ID, B-factor, occupancy, and charge.
```
cwognum, line 47: > only the most essential structural information is retained

**Question:** What information do we _lose_? It sounds pretty complete to me! Wasn't this lossless?

polaris/dataset/converters/_pdb.py
```diff
@@ -0,0 +1,158 @@
...
+            pdb_dict["hetero"].append(row.hetero)
+            pdb_dict["atom_name"].append(row.atom_name)
+            pdb_dict["element"].append(row.element)
+            pdb_dict["atom_id"].append(row.atom_id)
+            pdb_dict["b_factor"].append(row.b_factor)
+            pdb_dict["occupancy"].append(row.occupancy)
+            pdb_dict["charge"].append(row.charge)
+        return pdb_dict
+
+    def _load_pdf(self, path: str, pdb_pointer=None) -> dict:
```
cwognum, line 91: **Suggestion:** Minor

```suggestion
    def _load_pdb(self, path: str, pdb_pointer=None) -> dict:
```

polaris/dataset/converters/_pdb.py
```diff
@@ -0,0 +1,158 @@
...
+
+        if not isinstance(path, list):
+            path = [path]
+
+        pdb_pointers = []
+
+        # load single pdb and convert to zarr
+        for pdb_path in path:
+            # use the pdb file name as pointer
+            pdb_pointer = Path(pdb_path).stem
```
cwognum, line 145: **Suggestion:** Using `Path(...).stem` as pointer may lead to duplicate pointers. I think conversion should fail in that case. I think this is an issue for the other converters as well.

polaris/dataset/converters/_pdb.py
```diff
@@ -0,0 +1,158 @@
...
+            pdb_group = factory.zarr_root.create_group(self.pdb_column)
+        else:
+            pdb_group = factory.zarr_root[self.pdb_column]
+
+        group = pdb_group.create_group(pdb_pointer)
+        for col_name, col_val in pdb_dict.items():
+            col_val = np.array(col_val)
+            # get the accepted dtype
+            dtype = DTYPE_DICT.get(col_name, col_val.dtype)
+            group.create_dataset(col_name, data=col_val, dtype=dtype)
```
cwognum, line 132: **Suggestion:** For consistency and in line with official Zarr docs, I would use:
```suggestion
            group.array(col_name, data=col_val, dtype=dtype)
```

polaris/dataset/converters/_pdb.py
```diff
@@ -0,0 +1,158 @@
...
+        # load pdb as fastpbd and convert to dict
+        pdb_dict = self._load_pdf(path)
+
+        # Create group and add datasets
+        if self.pdb_column not in factory.zarr_root:
+            pdb_group = factory.zarr_root.create_group(self.pdb_column)
+        else:
+            pdb_group = factory.zarr_root[self.pdb_column]
+
+        group = pdb_group.create_group(pdb_pointer)
```
cwognum, line 127: **Suggestion:** For consistency and in line with official Zarr docs, I would use:

```suggestion
        group = pdb_group.group(pdb_pointer)
```

polaris/dataset/converters/_pdb.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        Convert a single pdb to zarr file
+        """
+
+        # load pdb as fastpbd and convert to dict
+        pdb_dict = self._load_pdf(path)
+
+        # Create group and add datasets
+        if self.pdb_column not in factory.zarr_root:
+            pdb_group = factory.zarr_root.create_group(self.pdb_column)
```
cwognum, line 123: **Suggestion**: For consistency and in line with official Zarr docs, I would use:

```suggestion
            pdb_group = factory.zarr_root.group(self.pdb_column)
```

Additionally, I think creating this group happen in the `convert()` method.

polaris/dataset/converters/_pdb.py
```diff
@@ -0,0 +1,158 @@
...
+
+class PDBConverter(Converter):
+    """
+    Converts PDB files into a Polaris dataset.
+    During the conversion, only the most essential structural information is retained, including 3D coordinates, chain ID, residue ID, insertion code, residue name, heteroatom indicator, atom name, element, atom ID, B-factor, occupancy, and charge.
+
+    Info: Numpy array for serialization
+        This class converts the 3D structure to fastpdb array (for ML purposes with key structual information).
+        This might not be the most storage efficient, but is fastest and easiest to maintain.
+        See[fastpdb](https://github.com/biotite-dev/fastpdb) and [biotite](https://github.com/biotite-dev/biotite/blob/main/src/biotite/structure/atoms.py) for more info.
```
cwognum, line 52: **Suggestion**:

```suggestion
    Info: PDBs as ND-arrays using `biotite`
        To save PDBs in a Polaris-compatible format, we convert them to ND-arrays using `fastpdb` and `biotite`.
        We then save these ND-arrays to Zarr archives. This conversion should be lossless.
        See [fastpdb](https://github.com/biotite-dev/fastpdb) and [biotite](https://github.com/biotite-dev/biotite/blob/main/src/biotite/structure/atoms.py) for more info.
```

cwognum (changes requested):
Thanks @zhu0619 !

I know it took some searching, but I think the solution you came up using `fastpdb` with is very polished! :nail_care: 

I did have some comments. In addition to these comments, would you also mind adding test cases?

polaris/dataset/_adapter_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    arrays = {}
+    for key, item in group.items():
+        if isinstance(item, zarr.core.Array):
+            arrays[key] = item[:]
+        elif isinstance(item, zarr.hierarchy.Group):
+            arrays[key] = load_group_as_numpy_arrays(item)
+    return arrays
+
+
+def zarr_to_pdb(group: zarr.Group):
```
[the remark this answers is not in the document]
zhu0619: This func is required by `_adaptor`, it would end up with a cycle import if moved to `polaris/dataset/converters/_pdb.py`. 
How about rename the file to `_pdb_utils.py`?

docs/tutorials/dataset_pdb.ipynb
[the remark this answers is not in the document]
zhu0619: Reconsidering the organization of the tutorials from the user’s perspective, i can image they would prefer to search by modality or file type—such as small molecules (SDF) or proteins (PDB)—rather than by Zarr files, which they typically don’t need to handle.

cwognum (approved):

cwognum:
FYI - We'll hold of on merging this to give #121 priority!

PR #176, opened 2024-08-13
Visualization Layer: Data Parsing Service 

kunkunzhu:
## Changelogs

- Add content type field to Column Annotation object

---

_Checklist:_

---

Issue [#422](https://github.com/polaris-hub/polaris-hub/issues/422)


polaris/dataset/converters/_sdf.py
```diff
@@ -141,6 +141,7 @@ def _get_name(mol: dm.Mol):
         annotations = {self.mol_column: ColumnAnnotation(is_pointer=True, modality=Modality.MOLECULE_3D)}
         if self.smiles_column is not None:
             annotations[self.smiles_column] = ColumnAnnotation(modality=Modality.MOLECULE)
+            annotations[self.smiles_column] = ColumnAnnotation(content_type=ContentType.SMILES)
```
jstlaurent, line 144: **Suggestion**
I think the second line might be overwriting the first line here. This might be better:
```suggestion
            annotations[self.smiles_column] = ColumnAnnotation(modality=Modality.MOLECULE, content_type=ContentType.SMILES)
```

polaris/dataset/_column.py
```diff
@@ -18,6 +18,12 @@ class Modality(enum.Enum):
     IMAGE = "image"
 
 
+class ContentType(str, enum.Enum):
+    """Used to specify column's IANA content type in a dataset."""
+
+    SMILES = Literal["chemical/x-daylight-smiles"]
```
jstlaurent, line 24: **Suggestion**
This type definition might be better defined with others we have in `polaris.utils.types`. We have other types there that leverage `Literal`, so maybe something like:
```python
KnownContentType: TypeAlias = Literal["chemical/x-daylight-smiles"]
ContentType: TypeAlias = KnownContentType | str
```

Also, not sure the enum adds a lot here... Maybe just using the string "chemical/x-daylight-smiles" directly? What do you think?

jstlaurent (commented):
Left a few comments. 😄

polaris/dataset/_column.py
cwognum, line 24: **Comment:**

More of a note for future reference, but I was curious how universal the `chemical/x-daylight-smiles` is.

The `x-daylight-` is throwing me off here. I would expect something more general like `chemical/smiles` (as these [notes](https://www.daylight.com/meetings/summerschool98/course/dave/smiles-relatives.html) suggest as well, see the quote :point_down: ). After a bit of searching it does seem that Daylight invented SMILES, so maybe it's okay? Wikipedia seems to say so ([ref](https://en.wikipedia.org/wiki/Chemical_file_format#The_Chemical_MIME_Project)) as does the [paper](https://pubs.acs.org/doi/10.1021/ci9803233).


> ### The "chemical/smiles" MIME type
> The "chemical" MIME type, championed by H. Rzepa of Imperial College, London, is (hopefully) nearing ratification. Should this happen, SMILES will almost certainly appear as MIME subtype, i.e., the MIME type: chemical/smiles. Until that blessed day, the IETF recommendation is to use something like application/x-smiles or x-chemical/x-smiles.

polaris/dataset/converters/_sdf.py
cwognum, line 1: **Suggestion:** 

Now that https://github.com/polaris-hub/polaris/pull/171 is merged, should we also add `chemical/x-pdb` as a known content type and update the PDB converter to use this ? What do you think?

polaris/dataset/_column.py
```diff
@@ -33,10 +39,11 @@ class ColumnAnnotation(BaseModel):
...
 
     is_pointer: bool = False
-    modality: Union[str, Modality] = Modality.UNKNOWN
-    description: Optional[str] = None
+    modality: str | Modality = Modality.UNKNOWN
+    description: str | None = None
     user_attributes: Dict[str, str] = Field(default_factory=dict)
-    dtype: Optional[Union[np.dtype, str]] = None
+    dtype: np.dtype | str | None = None
+    content_type: str | ContentType | None = None
```
cwognum, line 46: **Suggestion:** Include this newly added field in the docstring for the class.

I can imagine that having to set the content type, especially to something specific like `chemical/x-daylight-smiles`, is something the average ML engineer won't be used to or expect. Documenting the field is a first step, but it makes me wonder if there's other ways through which we could make for a better UX. E.g. allowing people to edit this field on the Hub!

cwognum (approved):
Looks great! I like that we're using the term _content type_ and known MIME types. We're not reinventing the wheel here!

polaris/dataset/_column.py
cwognum: On the use of an enum, one question came to mind: Should a known `content-type` be associated with a modality? If so, having an enum could define this mapping that we could then use in a Pydantic validator for the `ColumnAnnotation` ?

jstlaurent: I'm ok with shortening it if we want to. The `chemical` namespace was never ratified officially, even after nearly 30 years, so I don't think it's happening. So we either use the proposed types in https://en.wikipedia.org/wiki/Chemical_file_format, or we use `application/x-[something]`.

For reference, here are all currently official MIME types: https://www.iana.org/assignments/media-types/media-types.xhtml

PR #177, opened 2024-08-20
fix: Issue with Numpy version requirement from fastpdb/biotite

jstlaurent:
# Description

This PR pins the Numpy version to 1.x.y, to match the requirement from Biotite, itself required by FastPDB. On a clean virtual env install, the Numpy 2+ raises an error when importing Polaris.

### DIFF
--- a/env.yml
+++ b/env.yml
@@ -22,8 +22,8 @@ dependencies:
   - filelock
 
   # Scientific
-  - numpy
-  - pandas < 2.2.0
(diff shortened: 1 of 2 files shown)


cwognum (approved):
Thank you @jstlaurent !

Let's see how this evolves, but as we support more modalities we will depend on more and more niche packages like `biotite` and `datamol`. At some point we probably should split our dependencies in optional and required ones.

PR #181, opened 2024-08-20
Option to Append to an Existing Array When Loading from Multiple Files

zhu0619:
## Changelogs

- Add option to append an existing sdf array 
- Enable appending when loading from multiple files
- Added more tests for `add_from_files`

---

_Checklist:_

---

### DIFF
--- a/polaris/dataset/_factory.py
+++ b/polaris/dataset/_factory.py
@@ -221,7 +221,7 @@ class DatasetFactory:
             adapter = adapters.get(name)
             self.add_column(series, annotation, adapter)
 
-    def add_from_file(self, path: str, append: bool = False):
+    def add_from_file(self, path: str):
         """
         Uses the registered converters to parse the data from a specific file and add it to the dataset.
         If no converter is found for the file extension, it raises an error.
@@ -235,7 +235,7 @@ class DatasetFactory:
         if converter is None:
             raise ValueError(f"No converter found for extension {ext}")
 
-        table, annotations, adapters = converter.convert(path, self, append=append)
+        table, annotations, adapters = converter.convert(path, self)
         self.add_columns(table, annotations, adapters)
 
     def add_from_files(self, paths: List[str], axis: Literal[0, 1, "index", "columns"]):
@@ -256,9 +256,11 @@ class DatasetFactory:
                 raise ValueError(f"No converter found for extension {ext}")
 
             tables = []
+            append = False
             for path in paths:
-                table, annotations, adapters = converter.convert(path, self, append=True)
+                table, annotations, adapters = converter.convert(path, self, append=append)
                 tables.append(table)
+                append = True
             self.add_columns(pd.concat(tables, axis=0, ignore_index=True), annotations, adapters)
         else:
             for path in paths:
--- a/polaris/dataset/_factory.py
+++ b/polaris/dataset/_factory.py
@@ -228,7 +228,6 @@ class DatasetFactory:
 
         Args:
             path: The path to the file that should be parsed.
-            append: Allow to append an exisiting array.
         """
         ext = dm.fs.get_extension(path)
         converter = self._converters.get(ext)
--- a/polaris/dataset/_factory.py
+++ b/polaris/dataset/_factory.py
@@ -221,20 +221,21 @@ class DatasetFactory:
             adapter = adapters.get(name)
             self.add_column(series, annotation, adapter)
 
-    def add_from_file(self, path: str):
+    def add_from_file(self, path: str, append: bool = False):
         """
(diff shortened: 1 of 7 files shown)


polaris/dataset/_factory.py
```diff
@@ -221,20 +221,21 @@ def add_columns(
             adapter = adapters.get(name)
             self.add_column(series, annotation, adapter)
 
-    def add_from_file(self, path: str):
+    def add_from_file(self, path: str, append: bool = False):
```
cwognum, line 224: **Suggestion**: Do we need a _new_ parameter? Could we use `axis` here as well ?

polaris/dataset/converters/_sdf.py
```diff
@@ -103,25 +103,31 @@ def _get_name(mol: dm.Mol):
         if self.mol_prop_as_cols:
             props &= ~Chem.PropertyPickleOptions.MolProps
         bytes_data = [mol.ToBinary(props) for mol in df[tmp_col]]
-
         df.drop(columns=[tmp_col], inplace=True)
 
-        # Create the zarr array
-        factory.zarr_root.array(self.mol_column, bytes_data, dtype=bytes, chunks=self.chunks)
+        if self.mol_column in factory.zarr_root and append:
```
cwognum, line 108: **Suggestion**: This could lead to confusing behavior when `append == True` but the column doesn't exist. In that case we deviate from what the user asked and do so silently. I think it would be better to just fail if that happens. 

```py
if append:  # Or: if axis == 1
    if self.mol_column not in factory.zarr_root:
        raise RuntimeError(...)
```

polaris/dataset/converters/_sdf.py
```diff
@@ -103,25 +103,31 @@ def _get_name(mol: dm.Mol):
...
-                    )
+                start = pointer_start + group.index[0]
+                end = pointer_start + group.index[-1]
+                # Lu: is the check neccessary ?
+                # if group.nunique().sum() != len(group.columns): # deduct group column
+                #     print(group)
+                #     raise ValueError(
+                #         f"After grouping by {self.groupby_key}, values for other columns are not unique within a group. "
+                #         "Please handle this manually to ensure aggregation is done correctly."
+                #     )
```
cwognum, line 130: **Comment:** As discussed in person - This check is indeed necessary because in these cases you need to come up with a way to manually group the data. However, people are not required to use the factory, so this is not a blocker.

polaris/dataset/converters/_pdb.py
```diff
@@ -169,7 +169,7 @@ def _convert_pdb(
             dtype = DTYPE_DICT.get(col_name, col_val.dtype)
             group.array(col_name, data=col_val, dtype=dtype)
 
-    def convert(self, path: Union[str, List[str]], factory: "DatasetFactory") -> FactoryProduct:
+    def convert(self, path: Union[str, List[str]], factory: "DatasetFactory", **kwargs) -> FactoryProduct:
```
cwognum, line 172: **Suggestion:** It seems to me that appending is also relevant for the PDB converter. Maybe we can introduce functionality in the super class that can be shared between the different converters.

polaris/dataset/converters/_zarr.py
```diff
@@ -27,7 +27,7 @@ class ZarrConverter(Converter):
     to a single column and each array contains the values for all datapoints in that column.
     """
 
-    def convert(self, path: str, factory: "DatasetFactory") -> FactoryProduct:
+    def convert(self, path: str, factory: "DatasetFactory", **kwargs) -> FactoryProduct:
```
cwognum, line 30: **Suggestion:** Similar comment as for PDB converter. I think appending is also useful functionality for the Zarr converter.

cwognum (commented):
Nice work on the test cases! Thanks, @zhu0619 ! 

I did have some questions about the `append` argument.

polaris/dataset/_factory.py
zhu0619: `append` was already added in 'convert' function. This parameter should be removed from here. 
My mistake.

polaris/dataset/converters/_sdf.py
zhu0619: If the column doesn't exist, it will create a new array. 
Here the `append` parameter is rather `allow_appending`. So the existing array is allowed to be appended when user set it to True. 
I can rename it to 'allow_appending' if that clarifies.

cwognum (approved):
Thanks @zhu0619, looks good to me! 

As we start adding more complex functionality to this repository, I think we should sit down and think about the design a bit more. If we ever have some spare time... :sweat_smile:

Issue #132, opened 2024-07-17
Support downloading larger-than-memory datasets 


Andrewq11:
### Is your feature request related to a problem? Please describe.
Currently, there is a requirement to load an entire dataset into memory when fetching it from Polaris Hub. This introduces limitations related to the size of dataset a given user can interact with via the Polaris client. 

Depending on one's machine specs, specifically the amount of RAM, a given user's experience can be degraded if Polaris attempts to use too many system resources during dataset retrieval/processing. 

### Describe the solution you'd like
We need to ideate and implement a solution which enables the Polaris client to download and validate datasets from Polaris Hub in a more memory conscious manner. The aforementioned solution should remove the need for the entire dataset to be loaded into memory on client-side retrieval or interaction. 

### Describe alternatives you've considered
There is an option where we can make directly downloading datasets possible without the need to go through the client. The user could then take on the work of how to efficiently process/manipulate that dataset such that their system resources aren't drained. 

However, because benchmarks (and soon-to-be competitions) wrap a dataset with additional metadata that the client has been developed to support and validate, this option begins to breakdown. The client is required in these cases to ensure valid and comparable benchmarking results. 

Thus, for just getting access to a dataset, we could consider making the direct downloads available to users. However, in the context of benchmarks or competitions, this is not a realistic option for now.

PR #186, opened 2024-08-27
XL Datasets: Minimal Zarr-only dataset implementation

cwognum:
## Changelogs

- Introduce a new `DatasetV2` class in a new `polaris.experimental` module.
- Explicitly rename the old dataset to `DatasetV1`, but create the `Dataset` alias for it.
- Extract a `BaseDataset` class that encompasses all functionality that is shared between V1 and V2.
- Restricted use of `__getitem__` to only allow indexing a specific value or a row. 
- Add a new Zarr manifest for upload integrity. 
  - See #185

---

_Checklist:_

---

Closes https://github.com/polaris-hub/polaris/issues/132

The implementation has been relatively straight-forward given the [PoC](https://github.com/polaris-hub/polaris-private/blob/experiment/belka-dataset/scripts/cas/belka/nb/03.%20Subset%20API.ipynb). There are two points that required some further thinking.

1. **Naming and code organization**: I wasn't sure how to name the new class, where to introduce it and whether to rename the old class. For clarity, I decided to rename the old class, but we do maintain the `Dataset` alias for it to not break backwards compatibility. I also introduced a new `experimental` module to house the V2 implementation. 
2. **Indexing**: With https://github.com/polaris-hub/polaris/pull/171, we made it possible for the Zarr root to contain groups. This raises the question how to index those. For these cases, I decided to explicitly introduce a index array that needs to be part of that group.

### On V1 and V2 compatibility

You can think of a Pandas DataFrame as a set of named NumPy arrays. For storage, the conversion from Pandas to a Zarr archive is therefore relatively straight-forward. Note that this is limited to the storage - It's not straight-forward to implement the Pandas API on top of a Zarr archive, but this was decided to be out of scope. 

An example of this conversion was implemented in [this test case](https://github.com/polaris-hub/polaris/blob/feat/dataset-v2/tests/test_dataset_v2.py#L103-L127). A toy example would look like: 

```py
    df = pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]})

    root = zarr.open("path/to/arhive.zarr", "w")
    for col in df.columns:
        root.array(col, data=df[col].values)
```

### DIFF
--- a/polaris/dataset/_base.py
+++ b/polaris/dataset/_base.py
@@ -276,7 +276,8 @@ class BaseDataset(BaseArtifactModel, abc.ABC):
         """Uploads the dataset to the Polaris Hub."""
         raise NotImplementedError
 
-    @abc.abstractclassmethod
+    @classmethod
+    @abc.abstractmethod
     def from_json(cls, path: str):
         """
         Loads a dataset from a JSON file.
--- a/polaris/dataset/_base.py
+++ b/polaris/dataset/_base.py
@@ -1,8 +1,8 @@
 import abc
 import json
+import uuid
 from pathlib import Path
 from typing import Dict, List, MutableMapping, Optional, Union
-import uuid
 
 import fsspec
 import numpy as np
@@ -105,6 +105,24 @@ class BaseDataset(BaseArtifactModel, ChecksumMixin, abc.ABC):
 
     @model_validator(mode="after")
     def _validate_base_dataset_model(cls, m: "BaseDataset"):
+        # Verify that all annotations are for columns that exist
+        if any(k not in m.columns for k in m.annotations):
+            raise InvalidDatasetError(
+                f"There are annotations for columns that do not exist. Columns: {m.columns}. Annotations: {list(m.annotations.keys())}"
+            )
+
+        # Verify that all adapters are for columns that exist
+        if any(k not in m.columns for k in m.default_adapters.keys()):
+            raise InvalidDatasetError(
+                f"There are default adapters for columns that do not exist. Columns: {m.columns}. Adapters: {list(m.annotations.keys())}"
+            )
+
+        # Set a default for missing annotations and convert strings to Modality
+        for c in m.columns:
+            if c not in m.annotations:
+                m.annotations[c] = ColumnAnnotation()
+            m.annotations[c].dtype = m.dtypes[c]
+
         #
         # Set the default cache dir if none and make sure it exists
         if m.cache_dir is None:
--- a/polaris/dataset/_base.py
+++ b/polaris/dataset/_base.py
@@ -47,11 +47,6 @@ class BaseDataset(BaseArtifactModel, ChecksumMixin, abc.ABC):
     A Dataset can have multiple modalities or targets, can be sparse and can be part of one or multiple
      [`BenchmarkSpecification`][polaris.benchmark.BenchmarkSpecification] objects.
 
-    Info: Pointer columns
-        Whereas a `Dataset` contains all information required to construct a dataset, it is not ready yet.
-        For complex data, such as images, we support storing the content in external blobs of data.
-        In that case, the table contains _pointers_ to these blobs that are dynamically loaded when needed.
-
     Attributes:
         default_adapters: The adapters that the Dataset recommends to use by default to change the format of the data
             for specific columns.
--- a/polaris/dataset/_base.py
+++ b/polaris/dataset/_base.py
@@ -1,7 +1,7 @@
 import abc
 import json
 from pathlib import Path
-from typing import Dict, List, MutableMapping, Optional, Union
+from typing import Any, Dict, List, MutableMapping, Optional, Union
 
 import fsspec
 import numpy as np
@@ -26,6 +26,7 @@ from polaris.utils.dict2html import dict2html
 from polaris.utils.errors import InvalidDatasetError
 from polaris.utils.types import (
     AccessType,
+    DatasetIndex,
     HttpUrlString,
     HubOwner,
     SupportedLicenseType,
@@ -252,7 +253,9 @@ class BaseDataset(BaseArtifactModel, abc.ABC):
         self._zarr_data = load_zarr_group_to_memory(data)
 
     @abc.abstractmethod
-    def get_data(self, row: str | int, col: str, adapters: dict[str, Adapter] | None = None) -> np.ndarray:
+    def get_data(
+        self, row: str | int, col: str, adapters: dict[str, Adapter] | None = None
+    ) -> np.ndarray | Any:
         """Since the dataset might contain pointers to external files, data retrieval is more complicated
         than just indexing the `table` attribute. This method provides an end-point for seamlessly
         accessing the underlying data.
@@ -329,10 +332,16 @@ class BaseDataset(BaseArtifactModel, abc.ABC):
     def size(self) -> tuple[int, int]:
         return self.n_rows, self.n_columns
 
-    @abc.abstractmethod
-    def __getitem__(self, item):
+    def __getitem__(self, item: DatasetIndex) -> Any | np.ndarray | dict[str, np.ndarray]:
         """Allows for indexing the dataset directly"""
-        raise NotImplementedError
+
+        # If a tuple, we assume it's the row and column index pair
+        if isinstance(item, tuple):
+            row, col = item
+            return self.get_data(row, col)
+
+        # Otherwise, we assume you're indexing the row
+        return {col: self.get_data(item, col) for col in self.columns}
 
     @abc.abstractmethod
     def _repr_dict_(self) -> dict:
--- a/polaris/dataset/_base.py
+++ b/polaris/dataset/_base.py
@@ -8,7 +8,6 @@ import fsspec
 import numpy as np
 import pandas as pd
 import zarr
-from datamol.utils import fs as dmfs
 from loguru import logger
 from pydantic import (
     Field,
@@ -302,6 +301,7 @@ class BaseDataset(BaseArtifactModel, ChecksumMixin, abc.ABC):
         self,
         destination: str,
         if_exists: ZarrConflictResolution = "replace",
+        load_zarr_from_new_location: bool = False,
     ) -> str:
         """
         Save the dataset to a destination directory as a JSON file.
@@ -310,6 +310,8 @@ class BaseDataset(BaseArtifactModel, ChecksumMixin, abc.ABC):
             destination: The _directory_ to save the associated data to.
             if_exists: Action for handling existing files in the Zarr archive. Options are 'raise' to throw
                 an error, 'replace' to overwrite, or 'skip' to proceed without altering the existing files.
+            load_zarr_from_new_location: Whether to update the current instance to load data from the location
+                the data is saved to. Only relevant for Zarr-datasets.
 
         Returns:
             The path to the JSON file.
@@ -331,11 +333,7 @@ class BaseDataset(BaseArtifactModel, ChecksumMixin, abc.ABC):
         if cache_dir is not None:
             self.cache_dir = cache_dir
 
-        self.to_json(self.cache_dir)
-
-        if self.uses_zarr:
-            self.zarr_root_path = dmfs.join(self.cache_dir, "data.zarr")
-            self._zarr_root = None
+        self.to_json(self.cache_dir, load_zarr_from_new_location=True)
 
         if verify_checksum:
             self.verify_checksum()
--- a/polaris/dataset/_base.py
+++ b/polaris/dataset/_base.py
@@ -1,9 +1,8 @@
 import abc
 import json
 from pathlib import Path
-from typing import Any, Dict, List, MutableMapping, Optional, Union
+from typing import Any, Dict, MutableMapping, Optional, Union
 
-import fsspec
 import numpy as np
 import zarr
 from loguru import logger
@@ -92,7 +91,7 @@ class BaseDataset(BaseArtifactModel, abc.ABC):
         return {k: Adapter[v] if isinstance(v, str) else v for k, v in value.items()}
 
     @field_serializer("default_adapters")
-    def _serialize_adapters(self, value: List[Adapter]):
+    def _serialize_adapters(self, value: dict[str, Adapter]):
         """Serializes the adapters"""
         return {k: v.name for k, v in value.items()}
 
@@ -104,26 +103,26 @@ class BaseDataset(BaseArtifactModel, abc.ABC):
         return value
 
     @model_validator(mode="after")
-    def _validate_base_dataset_model(cls, m: "BaseDataset"):
+    def _validate_base_dataset_model(self):
         # Verify that all annotations are for columns that exist
-        if any(k not in m.columns for k in m.annotations):
+        if any(k not in self.columns for k in self.annotations):
             raise InvalidDatasetError(
-                f"There are annotations for columns that do not exist. Columns: {m.columns}. Annotations: {list(m.annotations.keys())}"
+                f"There are annotations for columns that do not exist. Columns: {self.columns}. Annotations: {list(self.annotations.keys())}"
             )
 
         # Verify that all adapters are for columns that exist
-        if any(k not in m.columns for k in m.default_adapters.keys()):
+        if any(k not in self.columns for k in self.default_adapters.keys()):
             raise InvalidDatasetError(
-                f"There are default adapters for columns that do not exist. Columns: {m.columns}. Adapters: {list(m.annotations.keys())}"
+                f"There are default adapters for columns that do not exist. Columns: {self.columns}. Adapters: {list(self.annotations.keys())}"
             )
 
         # Set a default for missing annotations and convert strings to Modality
-        for c in m.columns:
-            if c not in m.annotations:
-                m.annotations[c] = ColumnAnnotation()
-            m.annotations[c].dtype = m.dtypes[c]
+        for c in self.columns:
+            if c not in self.annotations:
+                self.annotations[c] = ColumnAnnotation()
+            self.annotations[c].dtype = self.dtypes[c]
 
-        return m
+        return self
 
     @property
     def client(self):
@@ -277,19 +276,15 @@ class BaseDataset(BaseArtifactModel, abc.ABC):
         """Uploads the dataset to the Polaris Hub."""
         raise NotImplementedError
 
-    @classmethod
+    @abc.abstractclassmethod
     def from_json(cls, path: str):
-        """Loads a dataset from a JSON file.
-        Overrides the method from the base class to remove the caching dir from the file to load from,
-        as that should be user dependent.
+        """
+        Loads a dataset from a JSON file.
 
         Args:
-            path: Loads a dataset specification from a JSON file.
+            path: The path to the JSON file to load the dataset from.
         """
-        with fsspec.open(path, "r") as f:
-            data = json.load(f)
-        data.pop("cache_dir", None)
-        return cls.model_validate(data)
+        raise NotImplementedError
 
     @abc.abstractmethod
     def to_json(
--- a/polaris/dataset/_base.py
+++ b/polaris/dataset/_base.py
@@ -1,12 +1,10 @@
 import abc
 import json
-import uuid
 from pathlib import Path
 from typing import Dict, List, MutableMapping, Optional, Union
 
 import fsspec
 import numpy as np
-import pandas as pd
 import zarr
 from loguru import logger
 from pydantic import (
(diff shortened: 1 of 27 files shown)


polaris/utils/v2_manifest.py
```diff
@@ -0,0 +1,71 @@
+from hashlib import md5
```
Andrewq11, line 1: What do you think about renaming this file to `manifest` and moving it to `polaris/dataset/zarr`? `utils` seems a little out of place but not terrible.

tests/test_dataset_v2.py
```diff
@@ -0,0 +1,271 @@
...
+from polaris.dataset.converters._pdb import PDBConverter
+from polaris.experimental._dataset_v2 import _INDEX_ARRAY_KEY, DatasetV2
+from polaris.utils.v2_manifest import generate_zarr_manifest
+
+
+def test_dataset_v2_get_columns(test_dataset_v2):
+    assert set(test_dataset_v2.columns) == {"A", "B"}
+
+
+def test_dataset_v2_get_rows(test_dataset_v2, zarr_archive):
```
Andrewq11, line 23:
```suggestion
def test_dataset_v2_get_rows(test_dataset_v2):
```

tests/test_evaluate.py
```diff
@@ -150,7 +151,7 @@ def test_absolute_average_fold_error():
 
 
 def test_metric_y_types(
-    tmpdir: str, test_single_task_benchmark_clf: SingleTaskBenchmarkSpecification, test_data: Dataset
+    tmpdir: str, test_single_task_benchmark_clf: SingleTaskBenchmarkSpecification, test_data: DatasetV1
```
Andrewq11, line 154: I know we can use the `tmpdir` argument with Pytest for getting easy access to clean directories, but is it needed in the few functions in this file that don't access it?

tests/test_dataset_v2.py
```diff
@@ -0,0 +1,271 @@
...
+
+    t2 = perf_counter()
+    for x in subset:
+        pass
+    d2 = perf_counter() - t2
+
+    assert d2 < d1
+
+
+def test_dataset_v2_checksum(test_dataset_v2, tmpdir):
```
Andrewq11, line 68: Is `tmpdir` needed here?

Andrewq11 (commented):
Thanks @cwognum! Left some comments but only a couple may be blocking.

jstlaurent (commented):
I have a few comments and suggestions. Overall, good, solid work. 😄

polaris/utils/v2_manifest.py
```diff
@@ -0,0 +1,71 @@
...
+    Recursive function that traverses a Zarr archive to build a V2 manifest file.
+
+    Parameters:
+        dir_path: The path to the current directory being processed in the archive
+        writer: Writer object for incrementally adding rows to the manifest Parquet file
+        zarr_root_path: The root path which triggered the first recursive call
+    """
+
+    # Get iterator of items located in the directory at `dir_path`
+    with os.scandir(dir_path) as it:
```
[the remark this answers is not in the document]
Andrewq11: We used `os.walk` in the original implementation and it does return a generator that yields a tuple for each directory level, but the `files` item in the tuple is an in-memory array. If there is a directory with millions of files/chunks, it could lead to issues with memory (though probably unlikely). 

We switched to `os.scandir` because each item in the directory is yielded individually.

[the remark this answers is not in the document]
jstlaurent: That's fair. I guess in the end that will depend on whether the Zarr structure is broader or deeper. I'd be curious to benchmark each approach once we get an actual XL dataset.

polaris/dataset/_column.py
```diff
@@ -42,29 +42,29 @@ class ColumnAnnotation(BaseModel):
...
 
     is_pointer: bool = False
-    modality: Union[str, Modality] = Modality.UNKNOWN
+    modality: Modality = Modality.UNKNOWN
     description: Optional[str] = None
     user_attributes: Dict[str, str] = Field(default_factory=dict)
-    dtype: Union[np.dtype, str, None] = None
-    content_type: Union[KnownContentType, str, None] = None
+    dtype: np.dtype | None = None
+    content_type: KnownContentType | None = None
```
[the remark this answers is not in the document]
jstlaurent: Also, maybe `KnownContentType` should be a Python `Literal` type, instead of an enum.

polaris/dataset/_base.py
```diff
@@ -0,0 +1,403 @@
...
+)
+
+# Constants
+_CACHE_SUBDIR = "datasets"
+
+
+class BaseDataset(BaseArtifactModel, ChecksumMixin, abc.ABC):
+    """Base data-model for a Polaris dataset, implemented as a [Pydantic](https://docs.pydantic.dev/latest/) model.
+
+    At its core, a dataset in Polaris is a tabular data structure that stores data-points in a row-wise manner.
```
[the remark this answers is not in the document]
cwognum: Even though we don't store it in an explicit table (i.e. Pandas DataFrame) anymore, I do think this conceptually still works as a mental model. I'll update the docstring to add that nuance! 👍

polaris/dataset/_base.py
```diff
@@ -0,0 +1,403 @@
...
+
+        if self._warn_about_remote_zarr:
+            saved_remote = saved_on_hub or not Path(self.zarr_root_path).exists()
+
+            if saved_remote:
+                logger.warning(
+                    f"You're loading data from a remote location. "
+                    f"To speed up this process, consider caching the dataset first "
+                    f"using {self.__class__.__name__}.cache()"
+                )
```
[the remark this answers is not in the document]
cwognum: Good idea, but I'll leave it as is for now.

polaris/dataset/_base.py
```diff
@@ -0,0 +1,403 @@
...
+            The path to the cache directory.
+        """
+
+        if cache_dir is not None:
+            self.cache_dir = cache_dir
+
+        self.to_json(self.cache_dir, load_zarr_from_new_location=True)
+
+        if verify_checksum:
+            self.verify_checksum()
```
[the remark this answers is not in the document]
cwognum: Good catch!

polaris/dataset/_base.py
```diff
@@ -0,0 +1,403 @@
...
+
+        return self.cache_dir
+
+    def size(self):
+        return self.rows, self.n_columns
+
+    def __getitem__(self, item):
```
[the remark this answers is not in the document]
cwognum: Oh, good catch. Yes, it should be!

tests/test_evaluate.py
cwognum: No, it's not! 👍

polaris/utils/v2_manifest.py
cwognum: Like it!

polaris/dataset/_base.py
```diff
@@ -0,0 +1,403 @@
...
+
+        Returns:
+            A numpy array with the data at the specified indices. If the column is a pointer column,
+                the content of the referenced file is loaded to memory.
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def upload_to_hub(
+        self, access: Optional[AccessType] = "private", owner: Union[HubOwner, str, None] = None
```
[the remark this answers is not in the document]
cwognum: The `owner` parameter is optional because it could also be set as a `Dataset` attribute. See e.g. [here](https://github.com/polaris-hub/polaris/blob/main/polaris/hub/client.py#L600).

polaris/dataset/_subset.py
```diff
@@ -77,10 +77,9 @@ def __init__(
         self._adapters = adapters
         self._featurization_fn = featurization_fn
 
-        # NOTE (cwognum): Note to future self. As we're starting to think about competition-style benchmarks,
-        #  we will likely split up datasets. In that case, this default iloc_to_loc mapping won't work.
-        #  By that time, we should probably be able to overwrite this mapping.
-        self._iloc_to_loc = self.dataset.table.index
+        # Storing all indices in memory can be memory consuming for XXL datasets.
+        # This is why we constrain the iloc to loc mapping to be the identity function for Dataset V2.
+        self._iloc_to_loc = self.dataset.rows if isinstance(self.dataset, DatasetV1) else None
```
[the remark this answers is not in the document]
cwognum: Love this! Good idea!

polaris/dataset/_subset.py
```diff
@@ -167,9 +166,15 @@ def as_array(self, data_type: Union[Literal["x"], Literal["y"], Literal["xy"]]):
...
         if data_type == "x":
-            ret = [self._get_single_input(self._iloc_to_loc[idx]) for idx in self.indices]
+            ret = [
+                self._get_single_input(idx if self._iloc_to_loc is None else self._iloc_to_loc[idx])
+                for idx in self.indices
+            ]
         else:
-            ret = [self._get_single_output(self._iloc_to_loc[idx]) for idx in self.indices]
+            ret = [
+                self._get_single_output(idx if self._iloc_to_loc is None else self._iloc_to_loc[idx])
```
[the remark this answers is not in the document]
cwognum: This is addressed by Julien his suggestion!

polaris/experimental/_dataset_v2.py
```diff
@@ -0,0 +1,218 @@
...
+
+    @property
+    def zarr_manifest_path(self) -> str:
+        if self._zarr_md5sum_manifest_path is None:
+            zarr_manifest_path = generate_zarr_manifest(self.zarr_root_path, self.cache_dir)
+            self._zarr_md5sum_manifest_path = zarr_manifest_path
+
+        return self._zarr_md5sum_manifest_path
+
+    def _compute_checksum(self) -> str:
```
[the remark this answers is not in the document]
cwognum: This is an override of the `ChecksumMixin` base class!

polaris/dataset/_base.py
```diff
@@ -0,0 +1,403 @@
...
+    def __eq__(self, other):
+        """Whether two datasets are equal is solely determined by the checksum"""
+        if not isinstance(other, BaseDataset):
+            return False
+        return self.md5sum == other.md5sum
+
+    def __del__(self):
+        """Close the connection of the client"""
+        if self._client is not None:
+            self._client.close()
```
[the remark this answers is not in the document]
cwognum: The issue is that we don't use the client directly, but rather use it to open a Zarr file hosted on the Hub in the `zarr_root` property. We could turn that property in a context manager, but we would then also have to do this for the `zarr_data` property, which uses the `zarr_root` property. 

I assume this part of the code will be revisited once we start using JWT tokens on the client. Maybe we could leave this as is for now?

polaris/experimental/_dataset_v2.py
```diff
@@ -0,0 +1,218 @@
...
+            col: The label of a direct child of the Zarr root.
+            adapters: The adapters to apply to the data before returning it.
+                If None, will use the default adapters specified for the dataset.
+
+        Returns:
+            A numpy array with the data at the specified indices. If the column is a pointer column,
+                the content of the referenced file is loaded to memory.
+        """
+        # Fetch adapters for dataset and a given column
+        adapters = adapters or self.default_adapters
```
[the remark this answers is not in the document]
cwognum: Like the idea of a partial override! 👍

polaris/mixins/_checksum.py
```diff
@@ -66,6 +64,9 @@ def verify_checksum(self, md5sum: str | None = None):
         self.md5sum = self._compute_checksum()
 
         if self.md5sum != md5sum:
+            # Imported here to prevent circular import
+            from polaris.utils.errors import PolarisChecksumError
```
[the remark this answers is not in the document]
cwognum: > (Ironically, one of the rare instance where having barrel files all over the place revealed a problem with the module structure, rather than caused one...)

I'm not sure what _barrel files_ are, but I would be interested to learn more about "module structure". We are more frequently running into circular imports in Polaris, so there is something valuable to learn there I think! 😄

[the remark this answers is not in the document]
cwognum: On this issue - I removed the formatting from the error messages. In addition to the module structure, I feel like improving our logging and error system are two other topics that I know little about, but that seem inevitably important to "production-ready" software and that I would thus love to learn more about!

polaris/dataset/_base.py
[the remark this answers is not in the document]
jstlaurent: I see... It feels a little confusing to me and I'd rather we have one way of setting ownership. There's a scenario here where I'm passing in a different owner than is set in the instance, and then the model I have locally doesn't match what's been uploaded to the Hub.

polaris/dataset/_base.py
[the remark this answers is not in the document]
jstlaurent: Ok, I'll refactor as part of the storage token work.

polaris/mixins/_checksum.py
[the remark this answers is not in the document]
jstlaurent: A barrel file is a name (that originated from the JavaScript/TypeScript world) used to describe a file that only imports and then reexports other elements. This is what the `__init__.py` files do in this package.

The module structure is basically how you split the code of the package. Ideally, it would look like a pyramid of successive layers, where higher-level layers, coarser layers of functionality call lower-level, finer grained layers of functionality. More formally, your dependency graph needs to be a DAG to avoid issues, and also to simplify the mental model of how pieces work together.

If you find yourself fighting circular imports, that means that you don't have a DAG and some lower level layer is calling on a higher level layer.

polaris/dataset/_base.py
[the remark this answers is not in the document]
cwognum: I'll update it to require explicitly setting the `Dataset.owner` attribute! :+1:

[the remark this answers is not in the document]
cwognum: Actually, that would affect the code examples on the Hub, which we have shown in various presentations and social posts. Removing the `owner` attribute from the Dataset class is also not an option because we use it in the `PolarisFileSystem` class. Finally, we could update the local `dataset.owner` attribute to match what was uploaded to the Hub, but that may be an equally confusing side-effect.

Since this becomes a more challenging issue than I initially thought and because this code wasn't introduced in this PR, I'm going to leave it as is for now. I did create https://github.com/polaris-hub/polaris/issues/192 to track this.

cwognum:
@jstlaurent @Andrewq11 Thank you for the review! 

This ended up being a good opportunity to also review the Dataset code that was mostly written over a year ago. One of the things I realized, is that the `__getitem__()` method was partially broken and implementing it properly would've taken a good amount of time and complicate the code. This would've become even more tricky for the `DatasetV2` class. I decided to therefore change this part of the API and simplify the type of indexing we allow.

Unfortunately, this does imply that some of the code examples we've shown in presentations are no longer accurate. See also https://github.com/polaris-hub/polaris-hub/pull/471. I am personally okay with this, since the Dataset API generally does not get a lot of attention compared to the Benchmark API. What do you think?

polaris/dataset/_base.py
```diff
@@ -123,14 +122,6 @@ def _validate_base_dataset_model(cls, m: "BaseDataset"):
                 m.annotations[c] = ColumnAnnotation()
             m.annotations[c].dtype = m.dtypes[c]
 
-        #
-        # Set the default cache dir if none and make sure it exists
-        if m.cache_dir is None:
```
Andrewq11: Why'd we move this back out of`BaseDataset` into the separate dataset classes again?

polaris/experimental/_dataset_v2.py
```diff
@@ -117,12 +127,32 @@ def zarr_manifest_path(self) -> str:
 
         return self._zarr_md5sum_manifest_path
 
-    def _compute_checksum(self) -> str:
-        """Compute the checksum of the Zarr manifest file."""
-        manifest_md5 = calculate_file_md5(self.zarr_manifest_path)
-        return manifest_md5
+    @computed_field
+    @property
+    def md5sum(self) -> str:
```
Andrewq11, line 132: Why'd we move this out of the ChecksumMixin class?

cwognum: Good question!

`DatasetV2` no longer bases the `ChecksumMixin`. Since the `md5sum` is not deterministic, it is used differently than the checksum for V1. E.g. the `verify_checksum()` method doesn't make much sense. To make that difference in use explicit, I added separate attributes here. See https://github.com/polaris-hub/polaris/pull/186#discussion_r1744684996 for context.

If this is causing confusion, perhaps we could rename the attribute here to make clear it's something else. What about `manifest_md5sum` ?

polaris/dataset/_base.py
cwognum: This is related to your other comment: Since the `BaseDataset` no longer inherits from the `ChecksumMixin`, the `_md5sum` property may not exist for subclasses.

polaris/utils/errors.py
```diff
@@ -60,7 +57,6 @@ def __init__(self, response: Response | None = None):
             "Note: If you can confirm that you are authorized to perform this action, "
             "please call 'polaris login --overwrite' and try again. If the issue persists, please reach out to the Polaris team for support."
         )
-        message = self.format(message, [self.BOLD, self.YELLOW])
```
Andrewq11, line 63: Have we lost the formatting of our error messages now? If so, I would've personally rathered the dubious import of the `PolarisChecksumError` in the `ChecksumMixin` class to preserve their styling while we create a separate ticket to more properly address it. Not a blocker though!

cwognum: We indeed lost the text formatting for the errors. I personally think that's okay. We're not using the formatting consistently for all errors and I feel like the whole "error" system (and logging, and module structure) is something we may want to address in the future.

polaris/experimental/_dataset_v2.py
Andrewq11: Makes sense, very nice. I think `manifest_md5sum` is a clearer name too!

polaris/utils/errors.py
Andrewq11: Sounds good. I think we should address it when we get some time. There should be a ticket regarding it floating around somewhere!

polaris/experimental/_dataset_v2.py
cwognum: Renamed the variables!

polaris/dataset/_base.py
```diff
@@ -0,0 +1,360 @@
...
+        Overrides the method from the base class to remove the caching dir from the file to load from,
+        as that should be user dependent.
+
+        Args:
+            path: Loads a dataset specification from a JSON file.
+        """
+        with fsspec.open(path, "r") as f:
+            data = json.load(f)
+        data.pop("cache_dir", None)
+        return cls.model_validate(data)
```
jstlaurent, line 292: **Comment**
I think `to_json` and `from_json` should be symmetrical. If one's implementation is dependent on the sub-class, then both should be abstract.

polaris/dataset/_subset.py
```diff
@@ -61,12 +61,12 @@ class Subset:
...
-        indices: List[Union[int, Sequence[int]]],
-        input_cols: Union[List[str], str],
-        target_cols: Union[List[str], str],
-        adapters: Optional[List[Adapter]] = None,
-        featurization_fn: Optional[Callable] = None,
+        dataset: DatasetV1,
+        indices: List[int | Sequence[int]],
+        input_cols: List[str] | str,
+        target_cols: List[str] | str,
+        adapters: List[Adapter] | None = None,
```
jstlaurent, line 68: **Suggestion**
This is the wrong type:
```suggestion
        adapters: dict[str, Adapter] | None = None,
```

polaris/experimental/_dataset_v2.py
```diff
@@ -0,0 +1,249 @@
...
+    version: ClassVar[Literal[2]] = 2
+    _zarr_manifest_path: str | None = PrivateAttr(None)
+    _zarr_manifest_md5sum: str | None = PrivateAttr(None)
+
+    # Redefine this to make it a required field
+    zarr_root_path: str
+
+    @model_validator(mode="after")
+    @classmethod
+    def _validate_v2_dataset_model(cls, m: "DatasetV2"):
```
jstlaurent, line 53: **Suggestion**
Model validator with `after` mode can be instance methods in Pydantic:
```suggestion
    @model_validator(mode="after")
    def _validate_v2_dataset_model(self) -> Self:
```

polaris/experimental/_dataset_v2.py
```diff
@@ -0,0 +1,249 @@
...
+        return arr
+
+    def upload_to_hub(self, access: AccessType = "private", owner: HubOwner | str | None = None):
+        """Uploads the dataset to the Polaris Hub."""
+
+        # NOTE (cwognum):  Leaving this for a later PR, because I want
+        #  to do it simultaneously with a PR on the Hub side.
+        raise NotImplementedError
+
+    def to_json(
```
jstlaurent, line 195: **Comment**
Not for this PR, but this method is becoming increasingly ill-named.

jstlaurent (approved):
Looks good. Still had a couple of things to say. 😉

polaris/experimental/_dataset_v2.py
cwognum: Maybe `save` and `load`.

polaris/experimental/_dataset_v2.py
jstlaurent: You can with [typing-extensions](https://pypi.org/project/typing-extensions/).

PR #187, opened 2024-09-01
Chore: First class predictions

### LINKED ISSUE #169: Add validators for the structure of predictions supplied to prediction evaluation methods
# Context

The required structure for benchmark predictions can vary based on the number of test sets _and_ tasks in a given benchmark. It is necessary that when users supply their predictions for evaluation (via the `evaluate` method in the `BenchmarkSpecification` class), the structure of the predictions must be compatible with the structure of the target labels (primarily defined in the `BenchmarkSpecification` class by the `target_cols` and `split` attributes).  

# Description

We should add additional validation or checks such that when users attempt to evaluate their results, we can confirm they are of the correct structure required by the benchmark. We should:
-  Create an intermediate `BenchmarkPredictions` class that users supply their predictions to prior to evaluation. It can contain validators which compare the expected structure of predictions relative to the supplied ones. If they do not match, it can return a helpful error message to the user explaining why. This route would require a small lift in updating documentation where we explain that users must use this new class as input to prediction evaluation methods. 

# Acceptance Criteria

- Prior to evaluation, the supplied structure of predictions is validated against the expected structure as defined by the `BenchmarkSpecification` object.
- There is helpful logging which guides users to the correct structure in the event their prediction structure is not correct.

### LINKED ISSUE #178: Standardize the expected type of predictions despite differences in the number of tasks and test sets in a benchmark/competition
# Context

Benchmarks can be created on Polaris which contain N number of tasks and N number of test sets. This introduces 4 classes of benchmarks which Polaris must support during the evaluation of user predictions. These classes are as follows:
- Single task, single set
- Multi-task, single set
- Single task, multi-set
- Multi-task, multi-set

Currently, evaluation methods in the Polaris library expect different structures for the submitted predictions depending on the category of a benchmark. The following is a breakdown of those expectations:
- Single task, single set -> `[values]`
- Multi-task, single set -> `{task_name_1: [values], task_name_2: [values], ...}`
- Single task, multi-set _and_  multi-task, multi-set -> `{test_set_1: {task_name: [values], ...}, ...}`

# Description

This current configuration is not only confusing, but also unneeded. We should standardize the expected structure of predictions to be the same irrespective of the benchmark type. The following should be the expected structure of the predictions across all benchmark types:

- `{test_name: {task_name: [values], ...}, ...}`

This is the structure that is currently supported for single task, multi-set _and_  multi-task, multi-set benchmarks.

# Acceptance Criteria

- Evaluation methods are updated to expect the aforementioned single structure _regardless_ of the benchmark type
- Benchmark tests are updated and pass with the new prediction structure
- The docstring for the `evaluate` method in the `BenchmarkSpecification` class is updated accordingly
- Evaluation methods continue to produce reliable and accurate results after the above changes are applied

kirahowe:
## Changelogs

- This PR extracts all of the normalization and serialization logic that goes into working with predictions into a standalone class for clarity
- Closes #169 
  - Predictions are still accepted in a very generous variety of formats and there no breaking changes introduced here, but the structure of provided predictions is now validated and standardizes, with helpful error messages for common error cases.  
- Closes  #178 
  - Evaluation logic now works with the canonical representation of predictions, for both regular and competition benchmarks.
- Competition documentation is removed from the docs while the feature is gated.

---

_Checklist:_


polaris/benchmark/predictions/_base.py
```diff
@@ -0,0 +1,129 @@
...
+
+        if isinstance(vals, (list, np.ndarray)):
+            return {"test": cls._set_col_name(vals, target_cols)}
+        elif cls._is_multi_task_single_test_set(vals, target_cols):
+            return {"test": vals}
+        elif cls._is_single_task_multi_test_set(vals, target_cols):
+            return {test_set_name: cls._set_col_name(v, target_cols) for test_set_name, v in vals.items()}
+        else:
+            cls._give_feedback(vals)
+        return vals
```
cwognum, line 63: **Suggestion**: 

If there are >1 target columns or >1 test sets, this logic allows for incomplete predictions. That is, predictions for only a subset of the test sets or targets. I think that case should fail. If there are multiple tasks or multiple test sets, you should provide predictions for all.

Another case I think should fail is when there are keys provided that don't correspond to a test set or target. This is less of an issue, but if we silently ignore these keys, that could lead to confusing behavior for the user. 

In summary, I suggest we standardize to a format where the outer dict has keys for _all_ test sets (and only those keys) and the inner dict has keys for _all_ targets (and only those keys).

This implies we will need to pass the test set labels to this class.

We should also add test cases for this.

polaris/benchmark/predictions/_base.py
```diff
@@ -0,0 +1,129 @@
...
+        {"test-set-name": {"target-name": np.ndarray}} regardless of how they are provided."""
+
+        if isinstance(vals, (list, np.ndarray)):
+            return {"test": cls._set_col_name(vals, target_cols)}
+        elif cls._is_multi_task_single_test_set(vals, target_cols):
+            return {"test": vals}
+        elif cls._is_single_task_multi_test_set(vals, target_cols):
+            return {test_set_name: cls._set_col_name(v, target_cols) for test_set_name, v in vals.items()}
+        else:
+            cls._give_feedback(vals)
```
cwognum, line 62: **Suggestion:** Should this always be called? I believe it may now for example happen that we create a predictions object based on a non-numeric list.

polaris/benchmark/predictions/_base.py
```diff
@@ -0,0 +1,129 @@
...
+    def _give_feedback(cls, vals):
+        for test_set, targets in vals.items():
+            if not isinstance(targets, dict):
+                raise ValueError(
+                    f"Invalid structure for test set '{test_set}'. "
+                    "Expected a dictionary of {col_name: predictions}"
+                )
+            for target, predictions in targets.items():
+                if not (
+                    isinstance(predictions, np.ndarray)
```
cwognum, line 120: **Suggestion:** Should we also check if the NumPy array is numeric?

cwognum: Actually, a related comment: I could imagine cases in which predictions can be non-numeric, e.g. if you predict the outcome of a chemical reaction you would likely pass the prediction as a SMILES string. We don't host any such benchmarks right now, but if we want to make sure current versions of the client can support such future benchmarks, maybe we shouldn't add this check?

polaris/evaluate/utils.py
```diff
@@ -3,85 +3,71 @@
...
+    for test_label, y_true_test in y_true.predictions.items():
         # For every metric...
         for metric in metrics:
             if metric.is_multitask:
                 # Multi-task but with a metric across targets
                 score = metric(
-                    y_true=y_true_subset, y_pred=y_pred.get(test_label), y_prob=y_prob.get(test_label)
+                    y_true=y_true_test,
+                    y_pred=y_pred.predictions.get(test_label),
+                    y_prob=y_prob.predictions.get(test_label),
```
cwognum, line 53: **Question:** `y_pred` and `y_prob` may be `None`, right? In which case this would crash?

polaris/evaluate/utils.py
```diff
@@ -3,85 +3,71 @@
...
                 continue
 
-            if not isinstance(y_true_subset, dict):
+            if not isinstance(y_true_test, dict):
                 # Single task
                 score = metric(
-                    y_true=y_true_subset, y_pred=y_pred.get(test_label), y_prob=y_prob.get(test_label)
+                    y_true=y_true_test,
+                    y_pred=y_pred.predictions.get(test_label),
+                    y_prob=y_prob.predictions.get(test_label),
```
cwognum, line 64: **Question**: y_pred and y_prob may be None, right? In which case this would crash?

polaris/evaluate/utils.py
```diff
@@ -3,85 +3,71 @@
...
-    elif vals is None:
-        return None
-    else:
-        return {"test": {target_cols[0]: vals}}
-
-
-def safe_mask(
-    input_values: dict | dict[str, dict], test_label: str, target_label: str, mask: NDArray[np.bool_]
-):
+def safe_mask(input_values: PredictionsType, test_label: str, target_label: str, mask: NDArray[np.bool_]):
```
cwognum, line 12: **Suggestion:** If we standardize as I suggested in another comment to have all keys correspond to a target / test set and to have keys for all targets / test sets, I don't think we need this anymore.

cwognum (commented):
This is already giving me peace of mind! :lotus_position: 

Seeing it all clearly in one spot did raise some questions. Let me know if any of it doesn't make sense. Happy to elaborate!

polaris/utils/types.py
```diff
@@ -24,19 +24,28 @@
 The second item can either be a single test set or a dictionary with multiple, named test sets.
 """
 
-PredictionsType: TypeAlias = Union[np.ndarray, dict[str, Union[np.ndarray, dict[str, np.ndarray]]]]
+ListOrArrayType: TypeAlias = Union[list[Union[int, float]], np.ndarray]
```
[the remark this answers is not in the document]
kiramclean: This is so much nicer, thanks! I didn't know Python supported this kind of syntax.

tests/test_benchmark_predictions.py
```diff
@@ -0,0 +1,98 @@
...
+        BenchmarkPredictions(
+            predictions={
+                "test1": {"col1": [1, 2, 3], "col2": [4, 5, 6]},
+                "test2": {"col1": [7, 8, 9], "col2": [10, 11, 12]},
+            },
+            target_cols=["col1", "col2"],
+        ).predictions,
+    )
+
+    "Invalid predictions"
```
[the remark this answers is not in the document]
kiramclean: Good idea, that actually passed! I added one with a bad column name and updated the feedback to say that column names have to match `target_cols`. I don't think we have a way to know what the test set names should be, but maybe I'm missing something and we do? If we so let me know and we can add a check that the test set names are valid, too.

docs/api/competition.evaluation.md
```diff
@@ -1,7 +1,3 @@
-::: polaris.evaluate.CompetitionPredictions
```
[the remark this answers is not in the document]
kiramclean: yes! I meant to ask how to do that.

polaris/benchmark/_base.py
```diff
@@ -460,31 +453,16 @@ def evaluate(
...
-                    y_true[test_set_name][self.target_cols[0]] = values.targets
+        y_true_subset = self._get_test_set(hide_targets=False)
+
+        if isinstance(y_true_subset, dict):
+            y_true_values = {k: v.targets for k, v in y_true_subset.items()}
         else:
-            #
-            # For single set benchmarks (single and multiple task)
-            y_true = test.targets
+            y_true_values = y_true_subset.targets
```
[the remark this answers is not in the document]
kiramclean: I did that in `evaluate_benchmark`, thinking it made sense to keep all the data manipulation parts together, since `evaluate_benchmark` is already very liberal with what it accepts. But if you think it makes more sense to do that here I'm happy to move it.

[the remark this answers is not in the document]
cwognum: Makes sense! I see it now. Thanks!

docs/api/competition.evaluation.md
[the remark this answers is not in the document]
cwognum: It should be as simple as including the same syntax, but updating the path. 

```
::: polaris.benchmark.predictions.BenchmarkPredictions
```

You can verify it locally with 
```
mkdocs serve
```

If you're curious to learn more, see [`mkdocs`](https://www.mkdocs.org/) and [`mkdocs-material`](https://squidfunk.github.io/mkdocs-material/). Let me know if you get stuck! Happy to help!

tests/test_benchmark_predictions.py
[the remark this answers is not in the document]
cwognum: I actually commented on that, see https://github.com/polaris-hub/polaris/pull/187#discussion_r1742704967 😄

docs/api/competition.evaluation.md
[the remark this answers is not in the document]
kiramclean: gotcha! I thought these were maybe auto-generated. I'll add this.

polaris/benchmark/predictions/_base.py
kiramclean: The `ListOrArrayType` should only allow ints or floats, the idea was to only support numeric predictions. But yeah this would be the place to update if/when we want to support non-numeric predictions. I can separate out a named method to make it more clear that this is meant to encapsulate validating specifically the prediction values.

polaris/benchmark/predictions/_base.py
kiramclean: Non-numeric lists are now supported, thanks for clarifying!

polaris/evaluate/utils.py
kiramclean: yeah that's strange.. that must be true, but I thought there were tests checking the cases where `y_pred` or `y_prob` were `None`.. will check out what's going on there.

polaris/utils/types.py
```diff
@@ -24,19 +24,26 @@
 The second item can either be a single test set or a dictionary with multiple, named test sets.
 """
 
-PredictionsType: TypeAlias = Union[np.ndarray, dict[str, Union[np.ndarray, dict[str, np.ndarray]]]]
+ListOrArrayType: TypeAlias = list | np.ndarray
```
Andrewq11, line 27: Do we currently host benchmarks or competitions (or will we soon) that expect predictions of a type other than `int`? If not, we could tighten the typing to `list[int]`?

tests/test_benchmark_predictions.py
```diff
@@ -0,0 +1,199 @@
...
+        BenchmarkPredictions(
+            predictions=[1, 2, 3], target_labels=["col1"], test_set_labels=["test"]
+        ).predictions,
+    )
+    assert_deep_equal(
+        {"test": {"col1": np.array([1, 2, 3])}},
+        BenchmarkPredictions(
+            predictions={"test": {"col1": [1, 2, 3]}}, target_labels=["col1"], test_set_labels=["test"]
+        ).predictions,
+    )
```
Andrewq11, line 33: Really like that you've put this together in a way where the final form is always accepted as initial input!

Andrewq11 (approved):
Great work @cwognum and @kiramclean 🚀 

I think this PR is very clean and adds much needed structure to prediction objects, thank you. I've added some comments but I don't think any are blocking.

polaris/utils/types.py
cwognum: Yes, we do! For example, in https://github.com/polaris-hub/polaris/pull/199 the prediction is a 3D molecule pose.

polaris/benchmark/_base.py
```diff
@@ -468,31 +471,21 @@ def evaluate(
...
-                    # For multi-task, multi-set benchmarks
-                    for task_name, values in values.targets.items():
-                        y_true[test_set_name][task_name] = values
-                else:
-                    #
-                    # For single task, multi-set benchmarks
-                    y_true[test_set_name][self.target_cols[0]] = values.targets
+        y_true_subset = self._get_test_set(hide_targets=False)
+
+        if isinstance(y_true_subset, dict):
```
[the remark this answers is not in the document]
cwognum: I think that's an excellent idea!

polaris/evaluate/_predictions.py
```diff
@@ -0,0 +1,174 @@
...
+    @field_validator("target_labels", "test_set_labels")
+    @classmethod
+    def _validate_labels(cls, v: list[str]) -> list[str]:
+        if len(set(v)) != len(v):
+            raise ValueError("The predictions contain duplicate columns")
+        return v
+
+    @model_validator(mode="before")
+    @classmethod
+    def _validate_predictions(cls, data: dict) -> dict:
```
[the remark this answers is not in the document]
cwognum: I think that's another great idea!

polaris/evaluate/_predictions.py
```diff
@@ -0,0 +1,197 @@
...
+
+                if len(target) != test_set_sizes[test_set_label]:
+                    raise ValueError(
+                        f"Predictions size mismatch: The predictions for test set '{test_set_label}' "
+                        f"should have a size of {test_set_sizes[test_set_label]}, but have a size of {len(target)}."
+                    )
+        return predictions
+
+    @classmethod
+    def _convert_lists_to_arrays(cls, predictions: IncomingPredictionsType) -> IncomingPredictionsType:
```
jstlaurent, line 184: **Suggestion**
This could go outside the class. Maybe even in the `utils`module.

polaris/evaluate/_predictions.py
```diff
@@ -0,0 +1,197 @@
...
+        for _, test_set_predictions in predictions.items():
+            if not isinstance(test_set_predictions, dict):
+                return False
+            if set(test_set_predictions.keys()) != set(target_labels):
+                return False
+
+        return True
+
+    @classmethod
+    def _check_test_set_size(
```
jstlaurent, line 168: **Suggestion**
This could also be a function outside the class.

polaris/evaluate/_predictions.py
```diff
@@ -0,0 +1,197 @@
...
+
+        validator = TypeAdapter(dict[str, int])
+        test_set_sizes = validator.validate_python(data.get("test_set_sizes"))
+
+        # Normalize the predictions to a standard representation
+        predictions = cls._convert_lists_to_arrays(predictions)
+        predictions = cls._normalize_predictions(predictions, target_labels, test_set_labels)
+
+        # Normalize the predictions to a standard representation
+        cls._check_test_set_size(predictions, test_set_sizes)
```
jstlaurent, line 82: **Comment**
I feel like this would be a better for an **after** model validator. That would fit better with the intent of the Pydantic validator API.

polaris/evaluate/_predictions.py
```diff
@@ -0,0 +1,197 @@
...
+        # Not a dictionary
+        if not isinstance(predictions, dict):
+            return False
+
+        # Outer-level of the dictionary should correspond to the test set names
+        if set(predictions.keys()) != set(test_set_labels):
+            return False
+
+        # Inner-level of the dictionary should correspond to the target columns
+        for _, test_set_predictions in predictions.items():
```
jstlaurent, line 159: **Suggestion**
```suggestion
        for test_set_predictions in predictions.values():
```

polaris/evaluate/_predictions.py
```diff
@@ -0,0 +1,197 @@
...
+
+        return {
+            "predictions": predictions,
+            "target_labels": target_labels,
+            "test_set_labels": test_set_labels,
+            "test_set_sizes": test_set_sizes,
+        }
+
+    @classmethod
+    def _normalize_predictions(
```
jstlaurent, line 92: **Comment**
This is very nice. Very soothing to me. 😄

jstlaurent (approved):
This looks very fine indeed. I had some suggestions, nothing blocking.

polaris/utils/misc.py
```diff
@@ -16,3 +18,19 @@ def slugify(sluggable: SlugCompatibleStringType) -> SlugStringType:
     Converts a slug-compatible string to a slug.
     """
     return sluggable.lower().replace("_", "-").strip("-")
+
+
+def convert_lists_to_arrays(predictions: ListOrArrayType | dict) -> np.array | dict:
```
jstlaurent, line 23: **Fix**
The rare moment where `Union` is required for Python to make sense of the type signature.

```suggestion
def convert_lists_to_arrays(predictions: Union[ListOrArrayType, dict]) -> np.array | dict:
```

jstlaurent (approved):
Looks good! Suggested a fox for the build error.

PR #189, opened 2024-09-03
Fix/Update polarisfs.base_path woth version number

zhu0619:
## Changelogs

- Updated `polarisfs.base_path` by adding version `v1`. 
---

_Checklist:_

---

### DIFF
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -404,6 +404,9 @@ class PolarisHubClient(OAuth2Client):
         except HTTPStatusError as error:
             # In this case, we can pass the response to provide more information
             raise PolarisHubError(message="Error opening Zarr store", response=error.response) from error
+        # This catches all other types of exceptions
+        except Exception as error:
+            raise PolarisHubError(message="Error opening Zarr store") from error
 
     def list_benchmarks(self, limit: int = 100, offset: int = 0) -> list[str]:
         """List all available benchmarks on the Polaris Hub.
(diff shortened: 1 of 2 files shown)


polaris/hub/polarisfs.py
```diff
@@ -89,6 +90,7 @@ def ls(
             timeout = self.default_timeout
 
         ls_path = self.sep.join([self.base_path, "ls", path])
+        print('ls_path.rstrip("/") - ', ls_path.rstrip("/"))
```
jstlaurent, line 93: **Suggestion**
Remove leftover debug logs.
```suggestion
```

polaris/hub/client.py
```diff
@@ -404,8 +404,6 @@ def open_zarr_file(
         except HTTPStatusError as error:
             # In this case, we can pass the response to provide more information
             raise PolarisHubError(message="Error opening Zarr store", response=error.response) from error
-        except Exception as error:
-            raise PolarisHubError(message="Error opening Zarr store") from error
```
jstlaurent, line 408: **Question**
Was this for debugging? If so, we might want to restore them, to wrap underlying errors into something more readable.

jstlaurent (approved):
Two comments, but ok to merge when they are addressed.

polaris/hub/client.py
zhu0619: There is another `except` right before this one. I thought it was an unintended duplicate.  
So it is only for readability of the error? I will add a comment for it.

Andrewq11: They're both useful. The first block catches errors of type `HTTPStatusError` and passes additional information available in this type to `PolarisHubError` (e.g. `response=error.response`) for display/formatting. The second block catches all other types of exceptions. 

We can't use a single block because `HTTPStatusError` contains additional error info that `Exception` does not.

Issue #173, opened 2024-08-12
Client Bucket Authentication: Get a bucket JWT from the Hub and use it to access the R2 bucket


jstlaurent:
# Context

Once the Hub's `token` endpoint has been updated, the client will be able to obtain a bucket JWT to perform upload or download operations on the bucket.

# Description

Update the way the client interacts with the R2 bucket:

- Prior to performing an operation (or a series of related operations in the case of a Zarr archive), obtain a bucket JWT from the Hub.
- Use a dedicated S3 client to connect to the bucket, instead of the Polaris fsspec implementation
- Use the bucket JWT when authenticating with the S3 client

# Acceptance Criteria

- The client does not use the Polaris fsspec implementation when interacting with the R2 bucket

# Links

- [Polaris Client Authentication Design Document](https://docs.google.com/document/d/12YFscUM_vWBsKlfwGmuVtGW4ULlzYThm-Eq_AcdgQiE/edit#heading=h.f0z109nuyp0x)
- [Hub token endpoint issue](https://github.com/polaris-hub/polaris-hub/issues/426)

PR #191, opened 2024-09-05
XL Datasets: Upload

Andrewq11:
## Changelogs

- Updated the `DatasetV2.upload_to_hub` method to call the `client.upload_dataset` method in the PolarisHubClient class
- Added some conditional logic to call `client._upload_v1_dataset` or `client._upload_v2_dataset` depending on the type of the dataset object passed as an argument
- Added the `client._upload_v2_dataset` method to handle uploading V2 datasets
- Leverage `StorageSession` and the storage tokens to upload the manifest and Zarr archive to the storage backend

~Currently on-hold until https://github.com/polaris-hub/polaris/issues/173 is addressed:~
~- Leveraging storage JWTs in `client._upload_v2_dataset`~
~- Leveraging storage JWTs in `client._get_v2_dataset` ~
**Edit:** Storage JWT work is done.

---

_Checklist:_

---


polaris/dataset/_base.py
```diff
@@ -0,0 +1,364 @@
...
+        if verify_checksum:
+            self.verify_checksum()
+
+        return self.cache_dir
+
+    def size(self) -> tuple[int, int]:
+        return self.n_rows, self.n_columns
+
+    @abc.abstractmethod
+    def __getitem__(self, item):
```
jstlaurent, line 333: **Comment**
Since this isn't implemented in `DatasetV2`, it shouldn't be in the common parent class.

polaris/experimental/_dataset_v2.py
```diff
@@ -0,0 +1,267 @@
...
+        #  https://github.com/polaris-hub/polaris/issues/188
+        super().cache(verify_checksum=False)
+
+    def __getitem__(self, item):
+        """Allows for indexing the dataset directly"""
+        raise NotImplementedError
+
+    def _repr_dict_(self) -> dict:
+        """Utility function for pretty-printing to the command line and jupyter notebooks"""
+        repr_dict = self.model_dump(exclude={"zarr_md5sum_manifest"})
```
jstlaurent, line 266: **Question**
Since `zarr_md5sum_manifest`is a `PrivateAttr`, do we need to exclude it explicitly here?

jstlaurent (commented):
Added some comments and question. Happy to see the new data structure is pretty similar. That means our initial design choices were good. 😄

cwognum (approved):
This looks great! I think it came together very nicely. Thanks @Andrewq11 and @jstlaurent ! 🙏 

My only comment concerns the `_md5sum` field in the `DatasetV2` class. I'm confused... I thought we explicitly decided to remove that from the V2 dataset, see e.g. https://github.com/polaris-hub/polaris/pull/186#discussion_r1747707041 . Looking through the commit history, it seems I may have added it? Could it have been reintroduced due to rebasing this branch?

Context switching is hard... I feel like it was a whole other person that worked on this.

polaris/hub/client.py
```diff
@@ -623,6 +632,85 @@ def _upload_dataset(
...
+        ) as progress_indicator:
+            # Check if a dataset license was specified prior to upload
+            if not dataset.license:
+                raise InvalidDatasetError(
+                    f"\nPlease specify a supported license for this dataset prior to uploading to the Polaris Hub.\nOnly some licenses are supported - {get_args(SupportedLicenseType)}."
+                )
+
+            # Normalize timeout
+            if timeout is None:
+                timeout = self.settings.default_timeout
```
Andrewq11, line 660: We should move the license and timeout checks (lines 653 and 659, respectively) to the calling `upload_dataset` function. They're shared right now and I doubt the logic will change.

polaris/hub/client.py
```diff
@@ -623,6 +632,85 @@ def _upload_dataset(
...
+                    "access": access,
+                    **dataset_json,
+                },
+                timeout=timeout,
+            )
+
+            with StorageSession(self, "write", dataset.urn) as storage:
+                # Step 2: Upload the manifest file
+                logger.info("Copying the dataset manifest file to the Hub.")
+                with open(dataset.zarr_manifest_path, "rb") as manifest_file:
```
Andrewq11, line 684: Not worth changing now, but this may be a bottleneck if the dataset, and thus the manifest, is sufficiently large.

We can leverage some [useful methods](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#pyarrow.parquet.ParquetFile.iter_batches) in PyArrow to bypass reading the entire manifest at once when/if this becomes an issue.

Andrewq11 (commented):
I can't approve it (because I'm assigned to it), but it looks good to me as well. Left just a few comments but none are blocking. Thanks @jstlaurent!

jstlaurent:
> My only comment concerns the `_md5sum` field in the `DatasetV2` class. I'm confused... I thought we explicitly decided to remove that from the V2 dataset, see e.g. [#186 (comment)](https://github.com/polaris-hub/polaris/pull/186#discussion_r1747707041) . Looking through the commit history, it seems I may have added it? Could it have been reintroduced due to rebasing this branch?

@cwognum: It might have been a rebase issue. It got a little confusing at some point. So, it should be removed?

polaris/experimental/_dataset_v2.py
```diff
@@ -264,7 +278,21 @@ def to_json(
             json.dump(serialized, f)
         return dataset_path
 
+    def cache(self, verify_checksum: bool = True) -> str:
```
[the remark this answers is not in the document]
jstlaurent: You shouldn't change the function signature when overriding it in a subclass. But since it basically calls `self.to_json` and returns path, I can push it down to the subclasses.

polaris/hub/client.py
jstlaurent: Good point that. I think I have an idea for that, but it'll wait until the next PR.

polaris/dataset/_base.py
```diff
@@ -80,6 +83,7 @@ class BaseDataset(BaseArtifactModel, abc.ABC):
     # Private attributes
     _zarr_root: zarr.Group | None = PrivateAttr(None)
     _zarr_data: MutableMapping[str, np.ndarray] | None = PrivateAttr(None)
+    _zarr_md5sum_manifest: list[ZarrFileChecksum] = PrivateAttr(default_factory=list)
```
cwognum, line 86: **Comment:** The `_zarr_md5sum_manifest` in V1 is - confusingly - different from the `_zarr_manifest_md5sum` in V2. The former is the list of checksums for all files in the Zarr archive, whereas the latter is the checksum of the manifest file. This thus only exists in the V1 class and shouldn't be in the base class.

polaris/dataset/_base.py
```diff
@@ -116,6 +120,19 @@ def _validate_base_dataset_model(self) -> Self:
...
 
+    @model_validator(mode="after")
+    def _ensure_cache_dir_exists(self) -> Self:
+        """
+        Set the default cache dir if none and make sure it exists
+        """
+        if self._cache_dir is None:
+            dataset_id = self._md5sum if self.has_md5sum else str(uuid4())
+            self._cache_dir = str(Path(DEFAULT_CACHE_DIR) / _CACHE_SUBDIR / dataset_id)
+        fs, path = fsspec.url_to_fs(self._cache_dir)
```
cwognum, line 131: **Comment:** The `_md5sum` and `has_md5sum` properties don't exist for V2, which is why we moved this logic related to the cache dir to the subclasses. We should probably pull out a `get_cache_dir()` method instead.

```suggestion
        cache_dir = self.get_cache_dir()
        fs, path = fsspec.url_to_fs(cache_dir)
```

polaris/dataset/_base.py
```diff
@@ -215,6 +232,18 @@ def dtypes(self) -> dict[str, np.dtype]:
         """Return the dtype for each of the columns for the dataset"""
         raise NotImplementedError
 
+    def should_verify_checksum(self, strategy: ChecksumStrategy) -> bool:
```
cwognum, line 235: **Comment**: Verifying the checksum (i.e. recomputing the checksum and comparing it against the provided checksum), only makes sense for the V1 implementation because the checksum for the V2 is not deterministic.

polaris/dataset/_base.py
```diff
@@ -294,13 +323,20 @@ def to_json(
...
+        Args:
+            verify_checksum: Whether to verify the checksum of the dataset after caching.
+
         Returns:
             The path to the cache directory.
         """
         self.to_json(self._cache_dir, load_zarr_from_new_location=True)
+
+        if verify_checksum:
+            self.verify_checksum()
```
cwognum, line 338: Comment: Same comment here. Verifying the checksum only makes sense for the V1 and shouldn't be in the base class.

polaris/experimental/_dataset_v2.py
```diff
@@ -161,6 +147,31 @@ def has_zarr_manifest_md5sum(self) -> bool:
...
+        """Set the checksum."""
+        if not re.fullmatch(r"^[a-f0-9]{32}$", value):
+            raise ValueError("The checksum should be the 32-character hexdigest of a 128 bit MD5 hash.")
+        self._md5sum = value
+
+    @property
+    def has_md5sum(self) -> bool:
+        """Whether the md5sum for this class has been computed and stored."""
+        return self._md5sum is not None
+
```
cwognum, line 174: **Comment:** To clearly mark the V2 implementation as a departure from the V1 implementation, we renamed the `_md5sum` property for V2 to `_zarr_manifest_md5sum`. These methods and the `_md5sum` property are thus not needed.

polaris/hub/client.py
```diff
@@ -623,6 +632,85 @@ def _upload_dataset(
...
+
+            # Upload dataset meta-data
+            url = f"/v2/dataset/{dataset.owner}/{dataset.name}"
+            response = self._base_request_to_hub(
+                url=url,
+                method="PUT",
+                json={
+                    "zarrManifestFileContent": {
+                        "md5Sum": dataset.md5sum,
+                    },
```
cwognum, line 674: **Comment:** I believe this should be:

```suggestion
                    "zarrManifestFileContent": {
                        "md5Sum": dataset.zarr_manifest_md5sum,
                    },
```

tests/test_dataset_v2.py
```diff
@@ -65,6 +65,13 @@ def test_dataset_v2_load_to_memory(test_dataset_v2):
     assert d2 < d1
 
 
+def test_dataset_v2_checksum(test_dataset_v2):
+    # Make sure the `md5sum` is part of the model dump even if not initiated yet.
+    # This is important for uploads to the Hub.
+    assert test_dataset_v2._md5sum is None
+    assert "md5sum" in test_dataset_v2.model_dump()
```
cwognum, line 72: **Comment:** Since the `_md5sum` property shouldn't be part of V2, this test case doesn't make sense.

cwognum (commented):
@jstlaurent Yes, it should be removed. I think https://github.com/polaris-hub/polaris/pull/186/ is a good PR to reference for what I would expect. 

Since the checksum for the `DatasetV2` dataset is not deterministic, it's a significant departure from the `DatasetV1` class. For example, as @Andrewq11 pointed out, verifying the checksum after download doesn't make much sense.

I've highlighted the changes I think should be reverted, but may have missed some.

polaris/dataset/_base.py
jstlaurent: Nice catch.

polaris/dataset/_base.py
jstlaurent: Let's just not use the md5 checksum for the cache dir. Fewer chances of problems that was as well.

cwognum (approved):
Thanks @jstlaurent ! 👍

PR #198, opened 2024-09-11
Type the return type of `@model_validator(mode="after")` with `Self`

cwognum:
## Changelogs

- Type the return type of `@model_validator(mode="after")` with `Self`

---

_Checklist:_

---

This came up in #186 , but I only saw it after merging. See https://github.com/polaris-hub/polaris/pull/186#discussion_r1755296478 .

### DIFF
--- a/polaris/dataset/_competition_dataset.py
+++ b/polaris/dataset/_competition_dataset.py
@@ -19,3 +19,5 @@ class CompetitionDataset(DatasetV1):
 
         if self.uses_zarr:
             raise InvalidCompetitionError("Pointer columns are not currently supported in competitions.")
+
+        return self
--- a/polaris/dataset/_competition_dataset.py
+++ b/polaris/dataset/_competition_dataset.py
@@ -1,4 +1,5 @@
 from pydantic import model_validator
+from typing_extensions import Self
 
 from polaris.dataset._dataset import DatasetV1
 from polaris.utils.errors import InvalidCompetitionError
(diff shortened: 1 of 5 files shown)


cwognum:
Sorry for not waiting for your response @jstlaurent ! Quickly opened this PR to still address the feedback! :smile:

polaris/dataset/_competition_dataset.py
```diff
@@ -13,7 +14,7 @@ class CompetitionDataset(DatasetV1):
     """
 
     @model_validator(mode="after")
-    def _validate_model(self):
+    def _validate_model(self) -> Self:
```
Andrewq11, line 17: Is `self` implicitly returned from `@model_validator`? We don't do it explicitly here so the `Self` typing may not be applicable.

Andrewq11 (approved):
Thanks @cwognum 🚀 

Left a single comment. Looks good to me once that is addressed.

polaris/dataset/_competition_dataset.py
Andrewq11: A better way to put it, could we add `return self` at the end of this validator?

jstlaurent: The methods needs to `return self`at the end.

jstlaurent (approved):
👍

polaris/dataset/_competition_dataset.py
cwognum: Good catch! Thanks! 🙏

PR #199, opened 2024-09-13
Add ligand RMSD metrics 

zhu0619:
## Changelogs

- Added metric `rmsd_coverage` for docking related benchmarks
- Added unit tests
- Added new target type "docking"

---

_Checklist:_

---
The rmsd metrics are implemented based on the [Posebuster codebase](https://github.com/maabuu/posebusters/blob/main/posebusters/modules/rmsd.py). A few modifications were made, replacing RDKit callables with equivalent Datamol functions. The RMSD calculation was also simplified to evaluate only the first conformer of the predicted molecule and to consider only heavy atoms.

Note: Posebusters provides a series of checkers, known as '[Posebuster Checkers](https://github.com/maabuu/posebusters/tree/main),' to further filter out unwanted docked ligand conformers. To avoid over-complicating Polaris metrics module with case-specific metrics, users should apply these filters before uploading results to the Polaris hub.

### DIFF
--- a/polaris/evaluate/metrics/docking_metrics.py
+++ b/polaris/evaluate/metrics/docking_metrics.py
@@ -19,7 +19,7 @@ def _rmsd(mol_probe: dm.Mol, mol_ref: dm.Mol) -> float:
             present, the lowest RMSD will be reported.
 
     Returns:
-        PoseBusters results dictionary.
+        Returns the RMS between two molecules, taking symmetry into account.
     """
 
     # copy the molecule for modification.
--- a/polaris/evaluate/metrics/docking_metrics.py
+++ b/polaris/evaluate/metrics/docking_metrics.py
@@ -1,20 +1,42 @@
 # This script includes docking related evaluation metrics.
 
 from typing import Union, List
-from copy import deepcopy
 
 import numpy as np
-from rdkit.Chem.rdMolAlign import CalcRMS, GetBestRMS
-
+from rdkit.Chem.rdMolAlign import CalcRMS
 import datamol as dm
 
 
 ## The following rmsd implementation are modified based on ppsebusters https://github.com/maabuu/posebusters/blob/main/posebusters/modules/rmsd.py
 
 
-def _call_rdkit_rmsd(mol_probe: dm.Mol, mol_ref: dm.Mol, conf_id_probe: int, conf_id_ref: int, **params):
+def _rmsd(mol_probe: dm.Mol, mol_ref: dm.Mol) -> float:
+    """Calculate RMSD between predicted molecule and closest ground truth molecule.
+       The RMSD is calculated with first conformer of predicted molecule and only consider heavy atoms for RMSD calculation
+
+    Args:
+        mol_probe: Predicted molecule (docked ligand) with exactly one conformer.
+        mol_ref: Ground truth molecule (crystal ligand) with at least one conformer. If multiple conformers are
+            present, the lowest RMSD will be reported.
+
+    Returns:
+        PoseBusters results dictionary.
+    """
+
+    # copy the molecule for modification.
+    mol_probe = dm.copy_mol(mol_probe)
+    mol_ref = dm.copy_mol(mol_ref)
+
+    # remove hydrogen from molecule
+    mol_probe = dm.remove_hs(mol_probe)
+    mol_ref = dm.remove_hs(mol_ref)
+
+    # calculate RMSD
     try:
-        return _rmsd(mol_probe, mol_ref, conf_id_probe, conf_id_ref, **params)
+        return CalcRMS(
+            prbMol=mol_probe, refMol=mol_ref, symmetrizeConjugatedTerminalGroups=True, prbId=-1, refId=-1
+        )
+
     except RuntimeError:
         pass
     except ValueError:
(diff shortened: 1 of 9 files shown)


polaris/evaluate/metrics/docking_metrics.py
```diff
@@ -0,0 +1,134 @@
...
+
+    return np.nan
+
+
+def _rmsd(
+    mol_probe: dm.Mol, mol_ref: dm.Mol, conf_id_probe: int, conf_id_ref: int, kabsch: bool = False, **params
+):
+    if kabsch is True:
+        return GetBestRMS(prbMol=mol_probe, refMol=mol_ref, prbId=conf_id_probe, refId=conf_id_ref, **params)
+    return CalcRMS(prbMol=mol_probe, refMol=mol_ref, prbId=conf_id_probe, refId=conf_id_ref, **params)
```
cwognum, line 31: **Question:** Do we need to support both options? Could we actually use [`dm.conformers.rmsd`](https://docs.datamol.io/stable/api/datamol.conformers.html#datamol.conformers._conformers.rmsd), which seems to use yet another method, i.e. `rdMolAlign.AlignMol()` ? How do these 3 differ?

polaris/evaluate/metrics/docking_metrics.py
```diff
@@ -0,0 +1,134 @@
...
+    conf_id_ref: int = -1,
+    drop_stereo: bool = False,
+    heavy_only: bool = True,
+    kabsch: bool = False,
+    symmetrizeConjugatedTerminalGroups=True,
+    **params,
+) -> float:
+    """RMSD calculation for isomers."""
+    mol_probe = deepcopy(mol_probe)
+    mol_ref = deepcopy(mol_ref)  # copy mols because rdkit RMSD calculation aligns mols
```
cwognum, line 47: **Suggestion:** Use `dm.copy_mol`. It does exactly the same, but that way if we ever improve the method somehow on the datamol side, we benefit from those changes here. 

```suggestion
    mol_probe = dm.copy_mol(mol_probe)
    mol_ref = dm.copy_mol(mol_ref)  # copy mols because rdkit RMSD calculation aligns mols
```

polaris/evaluate/metrics/docking_metrics.py
```diff
@@ -0,0 +1,134 @@
...
+import numpy as np
+from rdkit.Chem.rdMolAlign import CalcRMS, GetBestRMS
+
+import datamol as dm
+
+
+## The following rmsd implementation are modified based on ppsebusters https://github.com/maabuu/posebusters/blob/main/posebusters/modules/rmsd.py
+
+
+def _call_rdkit_rmsd(mol_probe: dm.Mol, mol_ref: dm.Mol, conf_id_probe: int, conf_id_ref: int, **params):
```
cwognum, line 15: **Suggestion:**
```suggestion
def _try_compute_rmsd(mol_probe: dm.Mol, mol_ref: dm.Mol, conf_id_probe: int, conf_id_ref: int, **params):
```

polaris/evaluate/metrics/docking_metrics.py
```diff
@@ -0,0 +1,134 @@
...
+        assert ValueError(
+            f"The list of probing molecules and the list of reference molecules are different sizes. {len(y_pred)} != {len(y_true)} "
+        )
+
+    rmsds = np.array(
+        [robust_rmsd(mol_probe=mol_probe, mol_ref=mol_ref) for mol_probe, mol_ref in zip(y_pred, y_true)]
+    )
+
+    if ignore_nan:
+        rmsds = rmsds[~np.isnan(rmsds)]
```
cwognum, line 132: **Question:** Do we need this option? At a first glance, it seems to me you would never want to use this because it could give a distorted view of the results (e.g. all are NaN except for 1 - Coverage could be 100%, but is that meaningful?)

polaris/evaluate/metrics/docking_metrics.py
```diff
@@ -0,0 +1,134 @@
...
+
+    return np.nansum(rmsds <= 2) * 100 / len(rmsds)
```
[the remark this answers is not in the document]
cwognum: I would also remove the `* 100`. I feel like metrics are more commonly reported in the `[0, 1]` domain, e.g. think of accuracy. What do you think?

polaris/evaluate/metrics/docking_metrics.py
```diff
@@ -0,0 +1,134 @@
...
+
+def conformer_to_mol(mol, conf):
+    """conv"""
+    mol = dm.copy_mol(mol)
+    mol.RemoveAllConformers()
+    mol.AddConformer(conf, assignId=True)
+    return mol
+
+
+def rmsd_coverage(
```
cwognum, line 108: **Question:** Other examples of RMSD I've seen typically include some sort of alignment step. Is that not needed here?

cwognum (commented):
Thanks @zhu0619 !

I understand that you intentionally mimicked the PoseBusters implementation, but I had some questions.

polaris/evaluate/metrics/docking_metrics.py
zhu0619: Here we want to compute the rmsd of two specific conformers in docking position.
`dm.conformers.rmsd` computes the rmsd of ALL the conformers of a molecule.  

After docking, the target was aligned with the original target from the crystal structure. The ligand was then extracted and used to compute the RMSD relative to the ligand in the reference position. 
Here, the metrics take only the ligands as input, assuming the ligand has already been aligned to the relevant position. So we don't want to use `AlignMol` to align them again.

polaris/evaluate/metrics/docking_metrics.py
zhu0619: I get your point. 
My initial thought was to provide an option that allows users to debug issues when RMSD discrepancies arise due to poor docking results or failures in RMSD computation.
But now I think the debugging should be handled at the user end.  

I will remove the option.

tests/conftest.py
```diff
@@ -125,6 +126,28 @@ def test_dataset(test_data, test_org_owner) -> DatasetV1:
     return dataset
 
 
+@pytest.fixture(scope="function")
+def test_docking_dataset(test_data, test_org_owner) -> DatasetV1:
+    # docking related
+    mols = test_data["smiles"].apply(dm.to_mol)
+    test_data["conformer_ref"] = np.array(
+        [conformer_to_mol(mol, dm.conformers.generate(mol).GetConformer(0)) for mol in mols]
```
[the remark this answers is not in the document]
zhu0619: This was to get the different conformers to compute the rmsd. 
It seems I can also archive that by specify the `random_seed` in the `generate` function.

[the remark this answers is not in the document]
zhu0619: You caught me being lazy here by skipping zarr loading. 
I will add proper zarr array.

polaris/evaluate/metrics/docking_metrics.py
```diff
@@ -0,0 +1,134 @@
...
+
+def _rmsd(
+    mol_probe: dm.Mol, mol_ref: dm.Mol, conf_id_probe: int, conf_id_ref: int, kabsch: bool = False, **params
+):
+    if kabsch is True:
+        return GetBestRMS(prbMol=mol_probe, refMol=mol_ref, prbId=conf_id_probe, refId=conf_id_ref, **params)
+    return CalcRMS(prbMol=mol_probe, refMol=mol_ref, prbId=conf_id_probe, refId=conf_id_ref, **params)
+
+
+def robust_rmsd(
```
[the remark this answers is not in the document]
zhu0619: I was being conservative here and try to keep close to the implementation from posebuster. I think the robustness came from the steps how they treat and fix the Mol objects when rmsd calculation fails.

Maybe again, this should be handled by user's end. And we assume the metric will apply on "ready" and "flawless" molecule inputs.  In this case the metrics can be simplified a lot. 

@Fransu86 What's your comments on the rmsd metrics from posebuster?

polaris/evaluate/metrics/docking_metrics.py
zhu0619: This was answered in earler comments.

polaris/evaluate/metrics/docking_metrics.py
```diff
@@ -0,0 +1,67 @@
...
+    if len(y_pred) != len(y_true):
+        assert ValueError(
+            f"The list of probing molecules and the list of reference molecules are different sizes. {len(y_pred)} != {len(y_true)} "
+        )
+
+    rmsds = np.array(
+        [_rmsd(mol_probe=mol_probe, mol_ref=mol_ref) for mol_probe, mol_ref in zip(y_pred, y_true)]
+    )
+
+    return np.nansum(rmsds <= max_rsmd) / len(rmsds)
```
cwognum, line 67: **Question:** Should we log a warning if there are any NaNs to explain how that may influence the results?

polaris/evaluate/metrics/docking_metrics.py
```diff
@@ -0,0 +1,67 @@
...
+
+import numpy as np
+from rdkit.Chem.rdMolAlign import CalcRMS
+import datamol as dm
+
+
+## The following rmsd implementation are modified based on ppsebusters https://github.com/maabuu/posebusters/blob/main/posebusters/modules/rmsd.py
+
+
+def _rmsd(mol_probe: dm.Mol, mol_ref: dm.Mol) -> float:
```
cwognum, line 13: **Question:** Are we confident the user will know to align the molecule beforehand? I am sure that for the cheminformatics community working on Docking this is common practice, but we're targeting ML scientists here who may not have that background. If it's as easy as calling a single method and it doesn't change an already aligned prediction, maybe we should just add it in? Or could we somehow check for alignment and raise an error if not aligned and ask the user to do it?

polaris/evaluate/metrics/docking_metrics.py
```diff
@@ -0,0 +1,67 @@
...
+
+from typing import Union, List
+
+import numpy as np
+from rdkit.Chem.rdMolAlign import CalcRMS
+import datamol as dm
+
+
+## The following rmsd implementation are modified based on ppsebusters https://github.com/maabuu/posebusters/blob/main/posebusters/modules/rmsd.py
+
```
cwognum, line 11: **Question:** Is this still accurate? Maybe we remove it? Or at least explain the difference to the original implementation?

polaris/evaluate/metrics/docking_metrics.py
```diff
@@ -0,0 +1,67 @@
...
+    # calculate RMSD
+    try:
+        return CalcRMS(
+            prbMol=mol_probe, refMol=mol_ref, symmetrizeConjugatedTerminalGroups=True, prbId=-1, refId=-1
+        )
+
+    except RuntimeError:
+        pass
+    except ValueError:
+        pass
```
cwognum, line 43: **Suggestion:** Could we add a comment explaining in which cases a user has provided a valid prediction to the benchmark, but the RMSD computation still fails? 

Something like: 
```suggestion
    # This can happen if ...
    except RuntimeError:
        pass
    # This can happen if ...
    except ValueError:
        pass
```

tests/test_evaluate.py
```diff
@@ -185,3 +185,13 @@ def test_metric_y_types(
...
+
+
+def test_metrics_docking(test_docking_benchmark: SingleTaskBenchmarkSpecification, caffeine, ibuprofen):
+    _, test = test_docking_benchmark.get_train_test_split()
+
+    predictions = np.array([caffeine, ibuprofen])
+
+    result = test_docking_benchmark.evaluate(y_pred=predictions)
+    for metric in test_docking_benchmark.metrics:
+        assert metric in result.results.Metric.tolist()
```
cwognum, line 197: **Suggestion:** Add sanity checks for extremes.

For example, with: 
- `predictions = np.array([caffeine, ibuprofen])`, coverage should be 1.0. 
- `predictions = np.array([ibuprofen, caffeine])`, coverage should be 0.0
  - or should this actually fail? Not the same number of atoms after all?

cwognum (commented):
Thank you @Fransu86 and @zhu0619 ! I love how much this has simplified! 

Couple more questions... As always!

polaris/evaluate/metrics/docking_metrics.py
zhu0619: The alignment needs to be done on the target protein with the docked ligand.  I don't think we should include that here.  Whoever decide to work on docking related task in AI community are supposed to know that.

Nevertheless, we can clarify that in the documentation.

polaris/evaluate/metrics/docking_metrics.py
```diff
@@ -0,0 +1,60 @@
...
+       The RMSD is calculated with first conformer of predicted molecule and only consider heavy atoms for RMSD calculation.
+       It is assumed that the predicted binding conformers are extracted from the docking output, where the receptor (protein) coordinates have been aligned with the original crystal structure.
+
+    Args:
+        mol_probe: Predicted molecule (docked ligand) with exactly one conformer.
+        mol_ref: Ground truth molecule (crystal ligand) with at least one conformer. If multiple conformers are
+            present, the lowest RMSD will be reported.
+
+    Returns:
+        PoseBusters results dictionary.
```
cwognum, line 22: **Fix**: Update the docs here!

cwognum (approved):
Looks great! Thank you for your patience! 🙏

PR #206, opened 2024-10-08
feat: deterministic Zarr checksum

### LINKED ISSUE #188: Generate a deterministic checksum for `DatasetV2` Zarr manifests
# Context

The `DatasetV2` class is the second generation of datasets on Polaris which are purpose-built to handle datasets of an unlimited size. This second generation heavily relies on the use of Zarr for chunking the dataset into more manageable pieces. 

With `DatasetV2`, a new mechanism was implemented for guaranteeing the validity of datasets. This mechanism involves creating a PyArrow table with the following schema and saving it to disk as a Parquet file:

| Path  |  Checksum | 
|---|---|
|  str  |  str | 

Once saved to disk, the md5 hash for the file is generated to produce the checksum for the _Zarr manifest_ of the dataset. This checksum is not used to check the validity of the dataset, it is only used to ensure the manifest file is intact when pulled in on the Polaris Hub. Once the integrity of the manifest is confirmed, the individual chunk checksums are used to guarantee the validity of each dataset chunk (and thus, the dataset). 

# Description

We currently use `os.scandir` to recurse throughout the Zarr archive and build the PyArrow table described above. Because `os.scandir` does not scan a directory deterministically, we produce different PyArrow tables on each walk of the same Zarr dataset. This ultimately produces a different Parquet file saved to disk and a different md5 hash for the same dataset.

We should devise a way to deterministically walk a Zarr archive such that the same Zarr manifest is produced for the same dataset. This method must utilize Python generators to prevent excess memory usage given the potential size of V2 datasets. 

# Acceptance Criteria

- Zarr manifest should include all files within the archive and their associated checksum
- The Zarr manifest checksum is the same when generated numerous times for the same dataset

# Links

- [Initial Zarr manifest implementation](https://github.com/polaris-hub/polaris/pull/185)

hmacdope:
## Changelogs

* Sort the directories of the dataset before they are walked by `recursively_build_manifest` to yield deterministic ordering of the Zarr manifest file and corresponding MD5 checksum. 
* Add test for recreating manifest file multiple times, checking its checksum

---

_Checklist:_

---

Fixes #188 

* This should (hopefully) take care of file ordering  messing up checksums for `zarr` manifests. 

* A specific case where you have encountered non-determinism may be helpful, as I struggled to generate a failing counterexample on `main`.

### DIFF
--- a/polaris/dataset/zarr/_manifest.py
+++ b/polaris/dataset/zarr/_manifest.py
@@ -25,6 +25,18 @@ def generate_zarr_manifest(zarr_root_path: str, output_dir: str):
     return zarr_manifest_path
 
 
+def deterministic_walk(root_path: str):
+    """
+    Recursively walk a directory and yield all files in a deterministic order.
+
+    Parameters:
+        root_path: The path to the root of the directory to walk
+    """
+
+    for dir_path, _, file_names in os.walk(root_path):
+        for file_name in sorted(file_names):
+            yield os.path.join(dir_path, file_name)
+
 def recursively_build_manifest(dir_path: str, writer: pq.ParquetWriter, zarr_root_path: str) -> str:
     """
     Recursive function that traverses a Zarr archive to build a V2 manifest file.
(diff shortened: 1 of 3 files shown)


polaris/dataset/zarr/_manifest.py
```diff
@@ -25,6 +25,20 @@ def generate_zarr_manifest(zarr_root_path: str, output_dir: str):
...
 
+def deterministic_walk(root_path: str):
+    """
+    Recursively walk a directory and yield all files in a deterministic order.
+
+    Parameters:
+        root_path: The path to the root of the directory to walk
+    """
+
+    for dir_path, dirs, file_names in os.walk(root_path):
```
Andrewq11, line 36: The main blocker here is that `file_names` will be held in-memory for each directory that is traversed. Depending on the size of the dataset and the chunking mechanism/pattern chosen by the dataset creator, it is possible to have a directory which contains enough files at a single level to exhaust one's system memory. 

We actually [explored this idea](https://github.com/polaris-hub/polaris/pull/186/commits/a968e0be7081dd6134776dcf41c7a80282d2c2f4#r1742914766) during our first implementation. Apologies, I probably should've explained this more clearly in the issue.

Andrewq11 (commented):
Thanks for taking the time to open this PR, @hmacdope. You're one of the first external contributors to Polaris and we appreciate your interest in the project. 

I've left a comment regarding the use of `os.walk`. While it would ensure deterministic chunk ordering, it could significantly increase memory usage by loading all file names into memory. Ideally, we’re looking for a solution that avoids this, which I understand can be tricky.

Let me know what you think and thanks again for the contribution.

polaris/dataset/zarr/_manifest.py
hmacdope: Ah no worries at all! I see why this would be problematic with all the `zarr` chunks. Feel free to close this PR as it won't address the issue. 

While researching for this issue I did come accross `zarr-checksum` that looks like it does what you need: https://github.com/dandi/zarr_checksum/tree/master and seems to have some of the `zarr` maintainers involved. If you think that looks like an acceptable solution then perhaps I could give that a try? Perhaps important question is whether the manifest file system is deployed (do we need to support two architectures?). Keen to hear your thoughts. 

I would love to keep contributing! If there are any issues I spy that I can help with. I also would love to contribute to XL competitions and benchmarks support (with my ASAPDiscovery hat on), so when the time comes I will try and help out.

Andrewq11: Building for the scale we hope to support has made for some interesting but tricky problems! It seems like Cas shared some additional context [here](https://github.com/polaris-hub/polaris/issues/188#issuecomment-2413631505) that I hope clarifies this checksum work a bit.

As for your other questions, the first version of the manifest file system is deployed and being used for V1 datasets which [leverage Zarr columns](https://polaris-hub.github.io/polaris/stable/tutorials/dataset_zarr.html). We're actively developing the second version of the system for V2 datasets, and it will diverge quite strongly in how it works relative to the current version. We plan to maintain both systems for the foreseeable future as deprecating the current system would make it challenging to ensure backward/forward compatibility.

Given this ongoing work, I'm going to close this PR, but we're looking forward to working together when you get the time! If you ever have any questions or need additional context, please feel free to reach out.

PR #208, opened 2024-10-24
chore: Update dataset manifest build

jstlaurent:
## Changelogs

- Update the way we build the Zarr archive manifest to use a single PyArrow table. This yields a reduction of ~90% in size of the resulting manifest, probably due to better compression. (A manifest of 3708 files went from 2.7MB to 119KB.)

---

_Checklist:_

- ~[ ] _Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR._~
- ~[ ] _Add tests to cover the fixed bug(s) or the newly introduced feature(s) (if appropriate)._~
- ~[ ] _Update the API documentation if a new function is added, or an existing one is deleted._~

---

### DIFF
--- a/polaris/dataset/zarr/_manifest.py
+++ b/polaris/dataset/zarr/_manifest.py
@@ -8,7 +8,8 @@ from pyarrow.parquet import write_table
 # PyArrow table schema for the V2 Zarr manifest file
 ZARR_MANIFEST_SCHEMA = schema([("path", string()), ("md5_checksum", string())])
 
-ROW_GROUP_SIZE = 128 * 1024 * 1024 # 128 MB
+ROW_GROUP_SIZE = 128 * 1024 * 1024  # 128 MB
+
 
 def generate_zarr_manifest(zarr_root_path: str, output_dir: str) -> str:
     """
--- a/polaris/dataset/zarr/_manifest.py
+++ b/polaris/dataset/zarr/_manifest.py
@@ -1,11 +1,12 @@
 import os
 from hashlib import md5
+from pathlib import Path
 
-import pyarrow as pa
-import pyarrow.parquet as pq
+from pyarrow import Table, schema, string
+from pyarrow.parquet import write_table
 
 # PyArrow table schema for the V2 Zarr manifest file
-ZARR_MANIFEST_SCHEMA = pa.schema([("path", pa.string()), ("md5_checksum", pa.string())])
+ZARR_MANIFEST_SCHEMA = schema([("path", string()), ("md5_checksum", string())])
 
 
 def generate_zarr_manifest(zarr_root_path: str, output_dir: str) -> str:
@@ -16,46 +17,37 @@ def generate_zarr_manifest(zarr_root_path: str, output_dir: str) -> str:
         zarr_root_path: The path to the root of a Zarr archive
         output_dir: The path to the directory which will hold the generated manifest
     """
-
     zarr_manifest_path = f"{output_dir}/zarr_manifest.parquet"
 
-    with pq.ParquetWriter(zarr_manifest_path, ZARR_MANIFEST_SCHEMA) as writer:
-        recursively_build_manifest(zarr_root_path, writer, zarr_root_path)
+    entries = manifest_entries(zarr_root_path, zarr_root_path)
+    manifest = Table.from_pylist(mapping=entries, schema=ZARR_MANIFEST_SCHEMA)
+    write_table(manifest, zarr_manifest_path)
 
     return zarr_manifest_path
 
 
-def recursively_build_manifest(dir_path: str, writer: pq.ParquetWriter, zarr_root_path: str) -> None:
+def manifest_entries(dir_path: str, root_path: str) -> list[dict[str, str]]:
     """
-    Recursive function that traverses a Zarr archive to build a V2 manifest file.
+    Recursive function that traverses a directory, returning entries consisting of every file's path and MD5 hash
 
     Parameters:
-        dir_path: The path to the current directory being processed in the archive
-        writer: Writer object for incrementally adding rows to the manifest Parquet file
-        zarr_root_path: The root path which triggered the first recursive call
+        dir_path: The path to the current directory being traversed
+        root_path: The root path from which to compute a relative path
     """
-
-    # Get iterator of items located in the directory at `dir_path`
+    entries = []
     with os.scandir(dir_path) as it:
-        #
-        # Loop through directory items in iterator
         for entry in it:
-            if entry.is_dir():
-                #
-                # If item is a directory, recurse into that directory
-                recursively_build_manifest(entry.path, writer, zarr_root_path)
-            elif entry.is_file():
-                #
-                # If item is a file, calculate its relative path and chunk checksum. Then, append that
-                # to the Zarr manifest parquet.
-                table = pa.Table.from_pydict(
+            if entry.is_file():
+                entries.append(
                     {
-                        "path": [os.path.relpath(entry.path, zarr_root_path)],
-                        "md5_checksum": [calculate_file_md5(entry.path)],
-                    },
-                    schema=ZARR_MANIFEST_SCHEMA,
+                        "path": str(Path(entry.path).relative_to(root_path)),
+                        "md5_checksum": calculate_file_md5(entry.path),
+                    }
                 )
-                writer.write_table(table)
+            elif entry.is_dir():
+                entries.extend(manifest_entries(entry.path, root_path))
+
+    return entries
 
 
 def calculate_file_md5(file_path: str) -> str:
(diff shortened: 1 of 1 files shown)


cwognum (approved):
Thanks, @jstlaurent.

If the data is saved as a single table, we assumed that would imply having to always load all of it into memory in its uncompressed form, which may not be possible for Zarr archives that have many files. 

However, this is a potential bottleneck we have not addressed in other parts of our infra. And after thinking about this some more, I don't think this is a limit we will run into anytime soon. For example, if we assume we save all 2.2M images of RXRX as separate files, then with some back-of-the-enveloppe style calculations we would have: 

```
>>> import sys
>>> from hashlib import md5
>>> sys.getsizeof(md5("/path/to/file".encode()).hexdigest())
81
>>> 81 * 2.2e6 * 2
356400000
```

So, ~0.3564 GB, uncompressed.

jstlaurent:
> Thanks, @jstlaurent.
> 
> If the data is saved as a single table, we assumed that would imply having to always load all of it into memory in its uncompressed form, which may not be possible for Zarr archives that have many files.
> 
> However, this is a potential bottleneck we have not addressed in other parts of our infra. And after thinking about this some more, I don't think this is a limit we will run into anytime soon. For example, if we assume we save all 2.2M images of RXRX as separate files, then with some back-of-the-enveloppe style calculations we would have:
> 
> ```
> >>> import sys
> >>> from hashlib import md5
> >>> sys.getsizeof(md5("/path/to/file".encode()).hexdigest())
> 81
> >>> 81 * 2.2e6 * 2
> 356400000
> ```
> 
> So, ~0.3564 GB, uncompressed.

Great comment. That sent me down a rabbit hole where I learned more about PyArrow and Parquet than I ever thought I would, to be honest.

The good news is that the very smart folks who built PyArrow and Parquet anticipated this problem. Inside a Parquet file, rows are grouped into what is aptly called [row groups](https://parquet.apache.org/docs/file-format/configurations/). When writing the Parquet file, using PyArrow, the [size of the row groups](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_table.html#pyarrow.parquet.write_table) can be specified.

The neat part is that when you want to read the file back, you can do so [by row group](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#pyarrow.parquet.ParquetFile.read_row_group). Only the data for that row group gets loaded into memory.

I also think using PyArrow's S3 file system might let it use range queries to pull only a subset of the file from the bucket. I'm less sure about that, because the doc is sparse there.

polaris/dataset/zarr/_manifest.py
```diff
@@ -1,12 +1,14 @@
...
-import pyarrow as pa
-import pyarrow.parquet as pq
+from pyarrow import Table, schema, string
+from pyarrow.parquet import write_table
 
 # PyArrow table schema for the V2 Zarr manifest file
-ZARR_MANIFEST_SCHEMA = pa.schema([("path", pa.string()), ("md5_checksum", pa.string())])
+ZARR_MANIFEST_SCHEMA = schema([("path", string()), ("md5_checksum", string())])
 
+ROW_GROUP_SIZE = 128 * 1024 * 1024 # 128 MB
```
mercuryseries, line 11: ruff format

```suggestion
ROW_GROUP_SIZE = 128 * 1024 * 1024  # 128 MB

```

mercuryseries (approved):
Thanks @jstlaurent! Beyond the performance improvements, was there any specific reason or issue that prompted this change?

cwognum:
> The good news is that the very smart folks who built PyArrow and Parquet anticipated this problem. Inside a Parquet file, rows are grouped into what is aptly called [row groups](https://parquet.apache.org/docs/file-format/configurations/). When writing the Parquet file, using PyArrow, the [size of the row groups](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_table.html#pyarrow.parquet.write_table) can be specified.
>
> The neat part is that when you want to read the file back, you can do so [by row group](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#pyarrow.parquet.ParquetFile.read_row_group). Only the data for that row group gets loaded into memory.

Indeed! I think @Andrewq11 and me reached the same conclusion. I forgot, however, why we didn't use the row group concept. I believe it's because we wanted to have a solution that can iteratively write, rather than all at once. Now the memory bottleneck is still a potential issue when creating the manifest locally.

But - in line with my earlier comments - I'm not too worried about that in the near future. 

(Famous last words)

Andrewq11 (commented):
Looks clean to me! Adding some thoughts to the comments already in this PR:

@cwognum 
> For example, if we assume we save all 2.2M images of RXRX as separate files, then with some back-of-the-enveloppe style calculations we would have: ...

Is this the "worst case scenario" in terms of the number of chunks we'll see in the near future? In your/our estmation, how likely will it be that we see datasets chunked more granularly and thus result in more files? 

In your example, my back-of-the-envelope calculations suggest that the `entries` variables will hold >700MB of memory prior to being written to disk as a parquet file:
```
import sys
from hashlib import md5

entries = []
number_of_files = 2200000  # 2.2 million
for i in range(number_of_files):
    random_path = "A/0.parquet" 
    random_checksum = md5("./README.md".encode()).hexdigest() 
    entries.append({"path": random_path, "md5_checksum": random_checksum})

total_size = sys.getsizeof(entries)  # shallow size of list
for item in entries:
    total_size += sys.getsizeof(item)  # shallow size of dict

    # Adding memory size of dict values b/c they are unique. The keys are the same and will likely
    # be handled more efficiently behind the scenes (e.g. storing a single copy).
    for value in item.values():
        total_size += sys.getsizeof(value)

print(total_size)  # >800MB
```
When I monitor my system memory usage with the above script, I'm getting about 730MB of usage.

I'm not against this approach because I have no benchmarking data to backup how many chunks we'll commonly see in a dataset, but I think that should be the deciding factor on whether we go with it. 

Disclaimer: I also haven't benchmarked the memory usage of the current approach, but I assume it is significantly less. 

@jstlaurent 
> The neat part is that when you want to read the file back, you can do so [by row group](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#pyarrow.parquet.ParquetFile.read_row_group). Only the data for that row group gets loaded into memory.

Another seemingly quite useful one is [`iter_batches`](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#pyarrow.parquet.ParquetFile.iter_batches). I don't think it relies on `row_groups` either so we could sidestep that if we wanted.

cwognum:
> Is this the "worst case scenario" in terms of the number of chunks we'll see in the near future?

I think that's a fair assumption! In theory, we have control over the chunk size and thus over the upper bound for the number of chunks. With images, however, if you want to use common codecs like PNG or JPEG2000, you'll need to use a chunk per image. I don't know of any other datasets that have significantly more images in our space, maybe JUMP CP (it's a bit opaque) but I would be surprised.

jstlaurent:
> Thanks @jstlaurent! Beyond the performance improvements, was there any specific reason or issue that prompted this change?

In the Inngest workflow that gets triggered on dataset creation, we make a copy of the manifest file, to add a _state_ column. We use that copy to keep track of the state of each chunk as the process goes through them to check they are present and the expected checksum matches.

When I was working on that part of the workflow, I noticed the copy of the file, with the additional column, occupied 120KB on disk, while the original manifest as 2.7MB. Spent a few hours trying to figure out what was being truncated, to realize that nothing was lost. Having one `Table` per entry means Parquet can't compress the file.

jstlaurent:
> When I monitor my system memory usage with the above script, I'm getting about 730MB of usage.
> 
> I'm not against this approach because I have no benchmarking data to backup how many chunks we'll commonly see in a dataset, but I think that should be the deciding factor on whether we go with it.
> 
> Disclaimer: I also haven't benchmarked the memory usage of the current approach, but I assume it is significantly less.

I'm actually less worried about memory usage on the client. The worst case scenario is someone running this on their MacBook with 16GB of memory and disk swap on a NVMe SSD. Compared to what our Modal functions will have available, that's bountiful.

There might also be a way to build the PyArrow table using [`from_batches`](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.from_batches), with a lazy generator hooked on the `os.scandir` iterator. I lack a good understanding of when PyArrow/Parquet will pull things in memory, so I don't know if that would actually make a difference on memory usage and/or impact compression on disk.

I created a fairly degenerate Zarr file with 33 million chunks, to have something to benchmark this kind of code. All chunks are fairly small, but the sheer number of them is actually causing problems with even basic shell commands like `ls`.

> Another seemingly quite useful one is [`iter_batches`](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#pyarrow.parquet.ParquetFile.iter_batches). I don't think it relies on `row_groups` either so we could sidestep that if we wanted.

That could be another option. I was thinking of using row groups to do fan-out in Modal, since that's partitioning that's built-into Parquet itself and not bytes offset we'd have to manage.

Either way, I want a way I can have control on how much data we pull in on the server-side. I like that PyArrow has multiple options for that.

PR #213, opened 2024-10-31
Fix: Exclude some computed properties during caching

cwognum:
## Changelogs

- Computing the zarr checksum manifest properties during caching would fail. We thus exclude these properties from the `model_dump()`


cwognum:
I'm going to skip review given that I would like to cut a new release of the client tonight before the announcement tomorrow.

PR #219, opened 2024-11-19
Chore: Add `imagecodecs` as an optional dependency

cwognum:
## Changelogs

- Adds `imagecodecs` as an optional dependency.
- Throw an error with an actionable message when a codec is missing.
- Update test cases to use `tmp_path` rather than `tmpdir` ([ref](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures)).

---

_Checklist:_

---

Out of curiosity: Are there any design patterns to handle optional dependencies in Python systematically? I did a quick Google search, but couldn't find much. Maybe @jstlaurent knows?

### DIFF
--- a/polaris/__init__.py
+++ b/polaris/__init__.py
@@ -4,7 +4,16 @@ import sys
 from loguru import logger
 
 from ._version import __version__
-from .loader import load_benchmark, load_dataset, load_competition
+from .loader import load_benchmark, load_competition, load_dataset
+
+try:
+    # Register imagecodecs if they are available.
+    from imagecodecs.numcodecs import register_codecs
+
+    register_codecs()
+except ImportError:
+    pass
+
 
 __all__ = ["load_dataset", "load_benchmark", "__version__", "load_competition"]
 
--- a/polaris/__init__.py
+++ b/polaris/__init__.py
@@ -6,15 +6,6 @@ from loguru import logger
 from ._version import __version__
 from .loader import load_benchmark, load_competition, load_dataset
 
-try:
-    # Register imagecodecs if they are available.
-    from imagecodecs.numcodecs import register_codecs
-
-    register_codecs()
-except ImportError:
-    pass
-
-
 __all__ = ["load_dataset", "load_benchmark", "__version__", "load_competition"]
 
 # Configure the default logging level
--- a/polaris/dataset/zarr/_utils.py
+++ b/polaris/dataset/zarr/_utils.py
@@ -40,5 +40,5 @@ def check_zarr_codecs(group: zarr.Group):
             raise error
 
         # Remove prefix and apostrophes
-        codec_id = error_message.removeprefix(prefix)[1:-1]
+        codec_id = error_message.removeprefix(prefix).strip("'")
         raise InvalidZarrCodec(codec_id)
--- a/polaris/dataset/zarr/_utils.py
+++ b/polaris/dataset/zarr/_utils.py
@@ -11,3 +11,22 @@ def load_zarr_group_to_memory(group: zarr.Group) -> dict:
         elif isinstance(item, zarr.Group):
             data[key] = load_zarr_group_to_memory(item)
     return data
+
+
+def check_zarr_codecs(group: zarr.Group):
+    """Check if all codecs in the Zarr group are registered."""
+    try:
+        for key, item in group.items():
+            if isinstance(item, zarr.Group):
+                check_zarr_codecs(item)
+    except ValueError as error:
+        # Zarr raises a generic ValueError if a codec is not registered.
+        # See also: https://github.com/zarr-developers/zarr-python/issues/2508
+        prefix = "codec not available: "
+        message = str(error)
+        codec_id = message.removeprefix(prefix)
+
+        if message.startswith(prefix):
+            raise RuntimeError(
+                f"This Zarr archive requires the {codec_id} codec. Install all optional codecs with 'pip install polaris-lib[codecs]'."
+            ) from error
--- a/polaris/dataset/zarr/_utils.py
+++ b/polaris/dataset/zarr/_utils.py
@@ -1,6 +1,16 @@
 import zarr
 import zarr.storage
 
+from polaris.utils.errors import InvalidZarrCodec
+
+try:
+    # Register imagecodecs if they are available.
+    from imagecodecs.numcodecs import register_codecs
+
+    register_codecs()
+except ImportError:
+    pass
+
 
 def load_zarr_group_to_memory(group: zarr.Group) -> dict:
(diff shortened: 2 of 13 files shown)


polaris/__init__.py
```diff
@@ -4,7 +4,16 @@
 from loguru import logger
 
 from ._version import __version__
-from .loader import load_benchmark, load_dataset, load_competition
+from .loader import load_benchmark, load_competition, load_dataset
```
mercuryseries, line 7: Do you aim to organize Python imports alphabetically whenever possible?

polaris/dataset/zarr/_utils.py
```diff
@@ -11,3 +11,22 @@ def load_zarr_group_to_memory(group: zarr.Group) -> dict:
...
+            if isinstance(item, zarr.Group):
+                check_zarr_codecs(item)
+    except ValueError as error:
+        # Zarr raises a generic ValueError if a codec is not registered.
+        # See also: https://github.com/zarr-developers/zarr-python/issues/2508
+        prefix = "codec not available: "
+        message = str(error)
+        codec_id = message.removeprefix(prefix)
+
+        if message.startswith(prefix):
```
mercuryseries, line 29: Minor: (for readability)

```suggestion
        if error_message.startswith(prefix):
```

polaris/dataset/zarr/_utils.py
```diff
@@ -11,3 +11,22 @@ def load_zarr_group_to_memory(group: zarr.Group) -> dict:
...
+            raise RuntimeError(
```
mercuryseries, line 30: Should we explicitly handle unexpected `ValueError` cases by re-raising them?

mercuryseries (approved):
Thanks, @cwognum! I’ve added a few minor suggestions—nothing major.

pyproject.toml
```diff
@@ -81,6 +81,9 @@ doc = [
     "nbconvert",
     "mike >=1.0.0"
 ]
+codecs = [
```
jstlaurent, line 84: **Comment**
FYI, the now standard way to define dependency groups has been codified in [PEP-735](https://peps.python.org/pep-0735/). It replaces `[project.optional-dependencies]`.

polaris/dataset/zarr/_utils.py
jstlaurent, line 29: **Question**
This will silently swallow any non-codec `ValueError`that gets raised. Is that ok?

polaris/dataset/zarr/_utils.py
jstlaurent, line 30: **Suggestion**
I'd suggest defining a specific exception class, rather than using a generic `RuntimeError`. This might allow code using the library to handle it or surface it in a specific way, without having to parse the error message.

polaris/dataset/zarr/_utils.py
```diff
@@ -11,3 +11,22 @@ def load_zarr_group_to_memory(group: zarr.Group) -> dict:
         elif isinstance(item, zarr.Group):
             data[key] = load_zarr_group_to_memory(item)
     return data
+
+
+def check_zarr_codecs(group: zarr.Group):
+    """Check if all codecs in the Zarr group are registered."""
+    try:
+        for key, item in group.items():
```
jstlaurent, line 19: **Question**
Is this where the exception gets raised, if the codec is not present?

Also,  does this require pulling the underlying data, or does this work purely on the metadata?

polaris/__init__.py
```diff
@@ -4,7 +4,16 @@
...
+
+try:
+    # Register imagecodecs if they are available.
+    from imagecodecs.numcodecs import register_codecs
+
+    register_codecs()
+except ImportError:
+    pass
```
jstlaurent, line 15: **Suggestion**
I'd put this in `polaris/dataset/zarr/_utils.py`, to be close the the `check_zarr_codecs` function. Otherwise, that's pretty much how you handle optional dependencies.

jstlaurent (approved):
Some comments, nothing major.

polaris/__init__.py
cwognum: Hahahaha I once installed a VS Code extension to do it for me. It does satisfy my OCD side! 😅

polaris/dataset/zarr/_utils.py
cwognum: No, it's not. Good catch!

polaris/dataset/zarr/_utils.py
cwognum: We should! Good catch!

cwognum: This is exactly what I suggested to the Zarr maintainers in https://github.com/zarr-developers/zarr-python/issues/2508, but funnily enough it didn't occur to me to follow my own advice.

polaris/dataset/zarr/_utils.py
cwognum: Yes, it raises the error as soon as you instantiate the array class that uses it. And no, this doesn't pull in any data.

tests/test_dataset.py
```diff
@@ -218,3 +222,37 @@ def _check_row_equality(d1, d2):
...
+
+def test_missing_codec_error(tmpdir):
+    """Check if the right error gets raised if a required codec is missing"""
+
+    class CustomCodec(Codec):
+        codec_id = "polaris_custom_test_codec"
+
+        def encode(self, buf):
+            buf = b"Random bytes"
+            return buf
```
mercuryseries, line 235: Nitpick:
```suggestion
            return b"Random bytes"
```

tests/test_dataset.py
```diff
@@ -218,3 +222,37 @@ def _check_row_equality(d1, d2):
 
     _check_row_equality(dataset[0], {"A": root["A"][0, :], "B": root["B"][0, :]})
     _check_row_equality(dataset[10], {"A": root["A"][10, :], "B": root["B"][10, :]})
+
+
+def test_missing_codec_error(tmpdir):
```
mercuryseries, line 227: **Minor:** I noticed that we’re currently mixing the use of `tmp_path` and `tmpdir`. While it doesn’t need to be addressed in this PR, it seems like `tmp_path` is nowadays [the preferred way](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures) over `tmpdir`.

mercuryseries (approved):

tests/test_dataset.py
cwognum: Good catch!

jstlaurent (approved):
👍

PR #223, opened 2024-11-25
Feature: Add the ability to compute metrics per group and then aggregate

cwognum:
## Changelogs

- Add support for a new metric type that computes a score per group and then aggregates the scores.
- Reworked the structure for metrics, allowing for configurable metrics.
- Fixed a minor bug in the `n_test_datapoints` method

---

_Checklist:_

---

[`RxRx3-Core`](https://www.rxrx.ai/rxrx3-core) calculates the metric first per treatment, before aggregating into a final score.

This caused some issues for our current Metric system. Up until this point, each metric in Polaris could be uniquely identified by a string label. We rely on this assumption throughout the code base (e.g. to serialize / deserialize metrics). In this PR, however, we're introducing a new grouped metric. This metric kind wraps another metric with additional logic and - importantly - is configurable through several variables. This implies that a simple string label can no longer uniquely identify a metric.

This raised the following questions: 
1. How should we serialize metrics moving forward?
    - How do we save metrics in the benchmark object? 
    - How do we reference those metrics in the results object?
2. How do we change the serialization without breaking backward's compatibility?


polaris/benchmark/_base.py
```diff
@@ -103,7 +103,7 @@ class BenchmarkSpecification(BaseArtifactModel, ChecksumMixin):
     target_cols: ColumnsType
     input_cols: ColumnsType
     split: SplitType
-    metrics: Union[str, Metric, list[str | Metric]]
+    metrics: str | Metric | GroupedMetric | list[str | Metric | GroupedMetric]
```
jstlaurent, line 106: **Comment**
This might be out of scope for the PR, but the type here should really be `set[Metric | GroupedMetric]`.


The `_validate_metrics` validator below should made into a **before** validator that normalizes the input.

polaris/benchmark/_base.py
```diff
@@ -147,9 +147,21 @@ def _validate_metrics(cls, v):
...
+            else:
+                unique_metrics.add(m.metric.name)
+
+            tmp.append(m)
+        v = tmp
+
+        if len(unique_metrics) != len(v):
             raise InvalidBenchmarkError("The task specifies duplicate metrics")
 
         if len(v) == 0:
```
jstlaurent, line 167: **Suggestion**
Something a little more compact:

```python
unique_metrics = set(Metric[m] if isinstance(m, str) else m for m in v)

if len(unique_metrics) != len(v):
    raise InvalidBenchmarkError("The task specifies duplicate metrics")

if len(unique_metrics) == 0:
    raise InvalidBenchmarkError("Specify at least one metric")

return unique_metrics
```

polaris/evaluate/_metric.py
```diff
@@ -116,23 +173,102 @@ def score(
...
+
+            try:
+                score = self.fn(**kwargs, **self.metric.value.kwargs)
+            except Exception as err:
+                if self.on_error == "ignore":
+                    continue
+                elif self.on_error == "raise":
+                    raise
+                elif self.default is not None:
+                    raise ValueError("Default value must be provided when on_error is 'default'") from err
```
jstlaurent, line 258: **Suggestion**
Implement a model validator so that this condition cannot happen:
```python
@model_validator(mode='after')
def check_default_required(self) -> Self:
    handling = self.on_error
    default = self.default
    if handling == 'default' and default is None:
        raise ValueError("Default value must be provided when on_error is 'default'")
    return self
```

polaris/evaluate/_metric.py
```diff
@@ -116,23 +173,102 @@ def score(
...
+class GroupedMetric(BaseModel):
+    """"""
+
+    metric: Metric
+
+    group_by: str
+    on_error: Literal["ignore", "raise", "default"] = "raise"
+    default: float | None = None
+    aggregation: Literal["mean", "median"] = "mean"
+
```
jstlaurent, line 204: **Suggestion**
Since this model is considered to be equivalent to the metric it wraps, for the intent of evaluating uniqueness, I'd suggest adding equal and hashcode implementations that will make that check easier:

```python
def __eq__(self, other: Self | Metric) -> bool:
    match other:
        case Metric():
            return other == self.metric
        case GroupedMetric():
            return self.metric == other.metric
        case _:
            return False

def __hash__(self) -> int:
    return hash(self.metric)
```

polaris/evaluate/_predictions.py
```diff
@@ -177,3 +179,69 @@ def _is_fully_specified(
...
+            return self
+
+        if test_set_subset is None:
+            test_set_subset = self.test_set_labels
+        if target_subset is None:
+            target_subset = self.target_labels
+
+        predictions = defaultdict(dict)
+        for test_set_name in self.predictions.keys():
+            if test_set_subset is not None and test_set_name not in test_set_subset:
```
jstlaurent, line 198: **Question**
Do you need this check? There's a guard above (`if test_set_subset is None:  test_set_subset = self.test_set_labels`) that should ensure it's always `not None`.

polaris/evaluate/_predictions.py
```diff
@@ -177,3 +179,69 @@ def _is_fully_specified(
...
+                continue
+
+            for target_name, target in self.predictions[test_set_name].items():
+                if target_subset is not None and target_name not in target_subset:
```
jstlaurent, line 202: **Question**
Likewise here, there's a guard above (`if target_subset is None:  target_subset = self.target_labels`) that should ensure this is always `not None`.

polaris/evaluate/_predictions.py
```diff
@@ -177,3 +179,69 @@ def _is_fully_specified(
...
+
+    def get_size(
+        self, test_set_subset: list[str] | None = None, target_subset: list[str] | None = None
+    ) -> int:
+        """Return the total number of predictions, allowing for filtering by test set and target"""
+        if test_set_subset is None and target_subset is None:
+            return sum(self.test_set_sizes.values()) * len(self.target_labels)
+        return len(self.get_subset(test_set_subset, target_subset))
+
+    def flatten(self) -> int:
```
jstlaurent, line 223: **Question**
Is the return value typing here accurate, given the doc?

jstlaurent (approved):
Very nice! I had a few questions and suggestions, but nothing blocking.

polaris/dataset/_subset.py
```diff
@@ -189,6 +191,38 @@ def as_array(self, data_type: Union[Literal["x"], Literal["y"], Literal["xy"]]):
...
+        for k in inputs:
+            df[k] = inputs[k]
+
+        return df
+
+    def filter_by_target(self, target_cols: List[str] | str) -> "Subset":
+        """Filter the subset to only include the specified target columns."""
+        target_cols = target_cols if isinstance(target_cols, list) else [target_cols]
+        copy = deepcopy(self)
+        copy.target_cols = target_cols
```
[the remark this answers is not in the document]
cwognum: Yes, it would have been. Good catch!

polaris/benchmark/_base.py
cwognum: Nice! For my own understanding: This does require [your other suggestion](https://github.com/polaris-hub/polaris/pull/223#discussion_r1858764073) to work, right?

polaris/evaluate/_predictions.py
cwognum: No, it's not. Good catch! Thanks!

polaris/benchmark/_base.py
jstlaurent: Yep.

polaris/evaluate/_metric.py
```diff
@@ -84,55 +164,241 @@ class Metric(Enum):
...
+        kind: The kind of metric, uses to discriminate different types of metrics in a discriminated union.
+        name: The name of the metric. The metrics for a specific benchmark need to be uniquely named.
+            In most cases, a sensible default will be set and the user doesn't need to worry about this.
+        config: For more complex metrics, this object should hold all parameters for the metric.
+    """
+
+    metric: BaseMetric
+    kind: Literal["default"] = "default"
+    name: str | None = None
+    config: BaseModel | None = None
```
jstlaurent, line 192: **Question**
I think this is where we should be using the discriminated union. For now, this should be:
```python
config: GroupedMetricConfig | None = None
```

Then the `kind` can actually be derived from what config model is defined. If there's nothing, then it's a default metric. If there a `GroupedMetricConfig`, then it's a grouped metric. If we have new types of metric in the future, then new configuration models will define what they are.

polaris/evaluate/_metric.py
```diff
@@ -84,55 +164,241 @@ class Metric(Enum):
...
-            if y_prob is None:
-                raise ValueError(f"{self} requires `y_prob` input. ")
-            pred = y_prob
+    # For any given benchmark, all of its metrics need to have a unique name.
+    # Within the scope of a given benchmark metric is thus uniquely identified by its name.
 
-        kwargs = {"y_true": y_true, self.y_type: pred}
-        return self.fn(**kwargs, **self.value.kwargs)
+    def __hash__(self) -> int:
+        return hash(self.name)
```
jstlaurent, line 271: **Question**
If the content of `config` matters to the behavior of the metric, then this is probably not accurate.

polaris/evaluate/_metric.py
```diff
@@ -84,55 +164,241 @@ class Metric(Enum):
...
 
-    def __call__(
-        self, y_true: np.ndarray, y_pred: np.ndarray | None = None, y_prob: np.ndarray | None = None
+    def __eq__(self, other) -> bool:
```
jstlaurent, line 273: **Comment**
See my question about `__hash__` above.

polaris/evaluate/_metric.py
```diff
@@ -84,55 +164,241 @@ class Metric(Enum):
...
+    Attributes:
+        kind: The kind of metric, uses to discriminate different types of metrics in a discriminated union.
+        config: The configuration for the grouped metric.
+    """
+
+    kind: Literal["grouped"] = "grouped"
+    config: GroupedMetricConfig
+
+    @model_validator(mode="after")
+    def check_default_required(self) -> Self:
```
jstlaurent, line 316: **Comment**
This should go into `GroupedMetricConfig`. That's where these attributes resides, and that's where they should be validated.

polaris/evaluate/_metric.py
```diff
@@ -84,55 +164,241 @@ class Metric(Enum):
...
+        """Check that a default value is provided when on_error is 'default'."""
+        handling = self.config.on_error
+        default = self.config.default
+        if handling == "default" and default is None:
+            raise ValueError("Default value must be provided when on_error is 'default'")
+        return self
+
+    def score(
```
jstlaurent, line 324: **Suggestion**
Move this into either the parent class or have it together with the config in one class.

- If into the parent class, have private `score` methods for each kind (default, grouped) of metrics.
- If into together with the config, have the parent class call it when config is present, with the default `solve` of no config.

polaris/evaluate/_metric.py
```diff
@@ -84,55 +164,241 @@ class Metric(Enum):
...
+A union of all supported Metric types in Polaris.
+"""
+
+AnnotatedMetricType = Annotated[MetricType, Field(..., discriminator="kind")]
+"""
+Annotates the MetricType with a discriminator field.
+"""
+
+
+def instantiate_metric(m: MetricType | str | dict):
```
jstlaurent, line 392: **Comment**
You don't need this anymore, or the union types above, if it's all in one `Metric` model.

polaris/benchmark/_base.py
```diff
@@ -103,8 +104,8 @@ class BenchmarkSpecification(BaseArtifactModel, ChecksumMixin):
     target_cols: ColumnsType
     input_cols: ColumnsType
     split: SplitType
-    metrics: Union[str, Metric, list[str | Metric]]
-    main_metric: str | Metric | None = None
+    metrics: set[MetricType]
```
jstlaurent, line 107: **Suggestion**
Let Pydantic work for you.

```suggestion
    metrics: set[MetricType] = Field(min_length=1)
```

polaris/benchmark/_base.py
```diff
@@ -136,35 +137,49 @@ def _validate_cols(cls, v, info: ValidationInfo):
...
+        unique_metrics = set(v)
 
-        if len(v) == 0:
+        if len(unique_metrics) != len(v):
+            raise InvalidBenchmarkError(
+                "The task specifies duplicate metrics. Make sure each metric has a unique name, e.g. Metric(name=...)"
+            )
+
+        if len(unique_metrics) == 0:
             raise InvalidBenchmarkError("Specify at least one metric")
```
jstlaurent, line 165: **Comment**
Not needed if you configure the field above.

polaris/benchmark/_base.py
```diff
@@ -136,35 +137,49 @@ def _validate_cols(cls, v, info: ValidationInfo):
...
+    def _validate_metrics(cls, v) -> set[MetricType]:
         """
         Verifies all specified metrics are either a Metric object or a valid metric name.
         Also verifies there are no duplicate metrics.
 
         If there are multiple test sets, it is assumed the same metrics are used across test sets.
         """
-        if not isinstance(v, list):
+        if isinstance(v, str):
             v = [v]
```
jstlaurent, line 150: **Suggestion**
Based on my suggestions for the `Metric`model:
```suggestion
            v = {'label': v}
```

polaris/benchmark/_base.py
```diff
@@ -136,35 +137,49 @@ def _validate_cols(cls, v, info: ValidationInfo):
...
+        if not isinstance(v, list):
+            v = list(v)
 
-        v = [m if isinstance(m, Metric) else Metric[m] for m in v]
+        for i, m in enumerate(v):
+            v[i] = instantiate_metric(m)
```
jstlaurent, line 155: **Suggestion**
Based on my suggestions for the `Metric`model:
```suggestion
        v = [m if isinstance(m, Metric) else Metric(**v) for m in v]
```

polaris/benchmark/_base.py
```diff
@@ -136,35 +137,49 @@ def _validate_cols(cls, v, info: ValidationInfo):
...
-    @field_validator("main_metric")
+    @field_validator("main_metric", mode="before")
+    @classmethod
     def _validate_main_metric(cls, v):
         """Converts the main metric to a Metric object if it is a string."""
-        # lu: v can be None.
-        if v and not isinstance(v, Metric):
-            v = Metric[v]
+        if v and not isinstance(v, MetricType):
+            v = instantiate_metric(v)
```
jstlaurent, line 174: **Suggestion**

```suggestion
        if v and isinstance(v, str):
            v = Metric(label=v)
```

polaris/benchmark/_base.py
```diff
@@ -262,22 +277,10 @@ def _validate_target_types(cls, v, info: ValidationInfo):
...
-    @field_serializer("metrics", "main_metric")
-    def _serialize_metrics(self, v):
-        """Return the string identifier so we can serialize the object"""
-        if isinstance(v, Metric):
-            return v.name
-        return [m.name for m in v]
+    @field_serializer("metrics")
+    def _serialize_metrics(self, v: set[MetricType]) -> list[dict]:
+        """Convert the set to a list"""
+        return [m.model_dump() for m in v]
```
jstlaurent, line 283: **Suggestion**
Return the models as a list, without dumping them, so that it plays nice with `include` and `exclude` option used when serializing this one:
```suggestion
        return list(v)
```

polaris/benchmark/_base.py
```diff
@@ -302,8 +305,12 @@ def _compute_checksum(self):
...
         for c in sorted(self.input_cols):
             hash_fn.update(c.encode("utf-8"))
-        for m in sorted(self.metrics, key=lambda k: k.name):
-            hash_fn.update(m.name.encode("utf-8"))
+        # We distinguish between Metrics and GroupedMetrics for backwards compatibility.
+        for name in sorted([m.name for m in self.metrics if isinstance(m, Metric)]):
+            hash_fn.update(name.encode("utf-8"))
+        for d in sorted([m.model_dump() for m in self.metrics if isinstance(m, GroupedMetric)]):
+            s = json.dumps(d, sort_keys="True")
+            hash_fn.update(s.encode("utf-8"))
```
jstlaurent, line 313: **Suggestion**
Encode the hash of the metric instance, rather than one of it's attribute. Then implement `__hash__` in the `Metric` class so that it returns a relevant and unique hash string.
```suggestion
        for metric_hash in sorted([hash(m) for m in self.metrics]):
            hash_fn.update(metric_hash.encode("utf-8"))
```

polaris/evaluate/_results.py
```diff
@@ -40,28 +41,24 @@ class ResultRecords(BaseModel):
 
     test_set: TestLabelType
     target_label: TargetLabelType
-    scores: dict[Metric, float]
+    scores: list[tuple[MetricType, float]]
```
jstlaurent, line 44: **Comment**
Not sure this is needed if you update the `Metric` class as I suggested.

polaris/evaluate/_results.py
```diff
@@ -40,28 +41,24 @@ class ResultRecords(BaseModel):
...
-            validated[metric] = score
+        validated = []
+        for metric, score in v:
+            metric = instantiate_metric(metric)
+            validated.append((metric, score))
         return validated
 
     @field_serializer("scores")
-    def serialize_scores(self, value: dict[Metric, float]) -> dict[str, float]:
+    def serialize_scores(self, value: list[tuple[MetricType, float]]) -> list[tuple[dict, float]]:
```
jstlaurent, line 59: **Comment**
I'm not sure you need this serializer at all.

jstlaurent (commented):
Left a bunch of comment. I think there's a way to have a nicer data structure, that has less impact on the existing code and is backwards-compatible with current benchmarks. That is, if I've understood the whole context for this change. I might be missing some info that makes my suggestions impractical. Let me know.

polaris/benchmark/_base.py
cwognum: This works, but to maintain backwards compatibility it does mean that the hash for a default metric has to be the label.

polaris/evaluate/_metric.py
cwognum: My thinking here was to shift the responsibility of defining a unique string ID per metric to the user. This was we could also show this unique ID in the UI for the Hub. Thinking about this again though, that actually seems like unnecessary friction and a worse UX.

polaris/benchmark/_base.py
cwognum: Actually, `hash()` needs to return an int, so this won't work.

polaris/evaluate/_results.py
cwognum: If it's a dict, the `Metric` is by default serialized to an inconvenient string that we can't directly deserialize, e.g. `"label='accuracy' config=None kind='default'"`. 

This raises a point I've run into before: It would be nice if we could have a unique (human-readable?) string identifier for each metric _within the context of a benchmark_, like we had before.

polaris/evaluate/_metric.py
```diff
@@ -84,55 +164,241 @@ class Metric(Enum):
...
+
+    Attributes:
+        metric: The actual callable that is at the core of the metric implementation.
+        kind: The kind of metric, uses to discriminate different types of metrics in a discriminated union.
+        name: The name of the metric. The metrics for a specific benchmark need to be uniquely named.
+            In most cases, a sensible default will be set and the user doesn't need to worry about this.
+        config: For more complex metrics, this object should hold all parameters for the metric.
+    """
+
+    metric: BaseMetric
```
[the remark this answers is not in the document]
cwognum: This cleans things up a lot! Thanks! 

Two points: 
- Because the `MetricInfo` should not be touched by the user, I changed this class to be frozen and made it a property of `Metric`, rather than it being part of the data model.
- I wasn't sure why you suggested changing to a `TypedDict`. To me it seems a Pydantic model has support for more features at a negligible cost. What am I missing here?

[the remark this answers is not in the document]
jstlaurent: >     * Because the `MetricInfo` should not be touched by the user, I changed this class to be frozen and made it a property of `Metric`, rather than it being part of the data model.

The individual fields can also be set to frozen, which gives you the same safety. You avoid having to have a bunch of computed properties just to handle another level of indirection.

>     * I wasn't sure why you suggested changing to a `TypedDict`. To me it seems a Pydantic model has support for more features at a negligible cost. What am I missing here?

It wasn't clear to me, at that point of the review, what this was meant to contain. If the idea is to store just a grab bag of arbitrary data, then is simpler to use a dict. If there's behavior attached to that data, then I agree a model makes more sense.

polaris/evaluate/_metric.py
```diff
@@ -34,105 +137,229 @@ class MetricInfo(BaseModel):
...
+            raise ValueError(f"Metric {self.label} not found in the available metrics.")
+        return metric_info
 
+    @computed_field
     @property
-    def y_type(self) -> bool:
-        """Whether the metric expects preditive probablities."""
-        return self.value.y_type
+    def kind(self) -> str:
+        return "default" if self.config is None else "grouped"
```
jstlaurent, line 292: **Suggestion**
This model doesn't need to know what other kind of metrics there are.

```suggestion
    def kind(self) -> MetricKind:
        return "default" if self.config is None else self.config._kind
```

polaris/evaluate/_metric.py
```diff
@@ -34,105 +137,229 @@ class MetricInfo(BaseModel):
...
+    """
+    The configuration for a GroupedMetric
+
+    Attributes:
+        group_by: The column to group by.
+        on_error: How to handle errors when computing the metric.
+        default: The default value to use when an error occurs.
+        aggregation: The aggregation method to use when combining the metric scores.
+    """
+
```
jstlaurent, line 197: **Suggestion**
If the config is in this model, then it should own all the config.

```suggestion
    _kind: MetricKind = PrivateAttr("grouped")
```

polaris/evaluate/_metric.py
```diff
@@ -15,15 +21,112 @@
...
+    Args:
+        y_true: The true target values of shape (n_samples,)
+        y_pred: The predicted target values of shape (n_samples,).
+        y_type: The type of predictions expected by the metric interface.
+    """
+    mask = mask_index(y_true)
+    return {"y_true": y_true[mask], y_type: y_pred[mask]}
+
+
```
jstlaurent, line 108: **Suggestion**

To go along with the rest.

```suggestion
MetricKind: TypeAlias = Literal["default", "grouped"]
```

polaris/evaluate/_metric.py
```diff
@@ -34,105 +137,229 @@ class MetricInfo(BaseModel):
...
 
     @property
-    def is_multitask(self) -> bool:
-        """Whether the metric expects a single set of predictions or a dict of predictions."""
-        return self.value.is_multitask
+    def info(self) -> MetricInfo:
+        """The metadata for the metric."""
+        metric_info = DEFAULT_METRICS.get(self.label)
+        if metric_info is None:
+            raise ValueError(f"Metric {self.label} not found in the available metrics.")
```
jstlaurent, line 286: **Comment**
This kind of validation should happen when the object is created, to prevent an invalid state from existing in the first place. Hence my initial suggestion to do this in a validator.

polaris/evaluate/_metric.py
```diff
@@ -34,105 +137,229 @@ class MetricInfo(BaseModel):
...
 
-        if self.y_type == "y_pred":
-            if y_pred is None:
-                raise ValueError(f"{self} requires `y_pred` input. ")
-            pred = y_pred
-        else:
-            if y_prob is None:
-                raise ValueError(f"{self} requires `y_prob` input. ")
-            pred = y_prob
+    def __hash__(self) -> str:
```
jstlaurent, line 348:
```suggestion
    def __hash__(self) -> int:
```

polaris/evaluate/_metric.py
```diff
@@ -34,105 +137,229 @@ class MetricInfo(BaseModel):
...
+            return adler32(self.label.encode())
 
-        kwargs = {"y_true": y_true, self.y_type: pred}
-        return self.fn(**kwargs, **self.value.kwargs)
+        # In case there is a config, we need to hash the config as well
+        config_dict = self.config.model_dump()
+        config_keys = sorted(config_dict.keys())
+        config_str = ",".join(f"{key}={config_dict[key]}" for key in config_keys)
+        hash_str = f"{self.label}({config_str})"
+        return adler32(hash_str.encode())
```
jstlaurent, line 358: **Suggestion**
Let Python work for you:
```python
        return hash(str(self))
```

polaris/evaluate/_metric.py
```diff
@@ -34,105 +137,229 @@ class MetricInfo(BaseModel):
...
-        self, y_true: np.ndarray, y_pred: np.ndarray | None = None, y_prob: np.ndarray | None = None
-    ) -> float:
-        """For convenience, make metrics callable"""
-        return self.score(y_true, y_pred, y_prob)
+    def __eq__(self, other) -> bool:
+        match other:
+            case Metric():
+                return hash(other) == hash(self)
+            case _:
+                return False
```
jstlaurent, line 365: **Suggestion**
To solve the hashing and benchmark checksum roblem:
```suggestion
                return False
                
    def __str__(self) -> str:
        return self.label + (json.dumps(self.config.model_dump(mode='json'), sort_keys=True) if self.config else '')
```

polaris/benchmark/_base.py
```diff
@@ -302,8 +306,11 @@ def _compute_checksum(self):
             hash_fn.update(c.encode("utf-8"))
         for c in sorted(self.input_cols):
             hash_fn.update(c.encode("utf-8"))
-        for m in sorted(self.metrics, key=lambda k: k.name):
-            hash_fn.update(m.name.encode("utf-8"))
+        # We distinguish between default and non-default metrics for the sake of backwards compatibility.
+        for label in sorted([m.label for m in self.metrics if m.kind == "default"]):
+            hash_fn.update(label.encode("utf-8"))
+        for hash in sorted(hash(m) for m in self.metrics if m.kind != "default"):
+            hash_fn.update(str(hash).encode("utf-8"))
```
jstlaurent, line 313: **Suggestion**
If you've defined `__str__`in `Metric`:
```suggestion
        for m in sorted(str(m) for m in self.metrics):
            hash_fn.update(m.encode("utf-8"))
```

jstlaurent (commented):
I think you can push this a little further, with even less coupling.

jstlaurent (approved):
Alright! 🚢 it!

cwognum:
@jstlaurent I decided to add back a `name` property to the Metric which uniquely identifies the metric _within the context of a benchmark_. This drastically simplified changes on the Hub side because I didn't have to work with the `Metric` `model_dump`. In the vast majority of cases, this property is set automatically and people don't have to worry about it.

jstlaurent (approved):
👍

PR #229, opened 2024-12-03
chore: Optimize Zarr store copy operation to Hub storage

jstlaurent:
## Changelogs

- Implement a custom Zarr store copy operation in the `S3Store`

---

_Checklist:_

---

A recent dataset added to the Hub sported a Zarr archive of modest size, but with a large number of files (194.056 MiB, 5067 files). The large number of files and the overhead of copying them to the Hub storage resulted in a transfer time of over 30 minutes. For comparison, using [Rclone](https://rclone.org/) to transfer the same Zarr archive to the same storage backend required 7 minutes.

The root cause of the client's slow upload lies in the implementation of the [`zarr.copy_store`](https://zarr.readthedocs.io/en/stable/api/convenience.html#zarr.convenience.copy_store) we use in our `StorageSession` class. This implementation makes no assumption about the nature of the source or destination stores, so it uses the most basic API supported by a Zarr store, [`MutableMapping`](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableMapping). It sequentially iterates through every key in the source store, checks their presence in the destination store, and then adds them to that store.

This PR updates our `S3Store` implementation with a custom copy method, that can leverage more efficient S3 operations, as well as parallel and concurrent operations, to improve throughput. The above mentioned dataset can now be copied to the Hub storage in ~2 minutes.

### DIFF
--- a/polaris/dataset/_base.py
+++ b/polaris/dataset/_base.py
@@ -1,5 +1,6 @@
 import abc
 import json
+from os import PathLike
 from pathlib import Path
 from typing import Any, MutableMapping
 from uuid import uuid4
@@ -173,15 +174,14 @@ class BaseDataset(BaseArtifactModel, abc.ABC):
 
         saved_on_hub = self.zarr_root_path.startswith(StorageSession.polaris_protocol)
 
-        if self._warn_about_remote_zarr:
-            if saved_on_hub:
-                # TODO (cwognum): The user now has no easy way of knowing whether the dataset is "small enough".
-                logger.warning(
-                    f"You're loading data from a remote location. "
-                    f"If the dataset is small enough, consider caching the dataset first "
-                    f"using {self.__class__.__name__}.cache() for more performant data access."
-                )
-                self._warn_about_remote_zarr = False
+        if self._warn_about_remote_zarr and saved_on_hub:
+            # TODO (cwognum): The user now has no easy way of knowing whether the dataset is "small enough".
+            logger.warning(
+                f"You're loading data from a remote location. "
+                f"If the dataset is small enough, consider caching the dataset first "
+                f"using {self.__class__.__name__}.cache() for more performant data access."
+            )
+            self._warn_about_remote_zarr = False
 
         try:
             if saved_on_hub:
@@ -310,12 +310,7 @@ class BaseDataset(BaseArtifactModel, abc.ABC):
         raise NotImplementedError
 
     @abc.abstractmethod
-    def to_json(
-        self,
-        destination: str,
-        if_exists: ZarrConflictResolution = "replace",
-        load_zarr_from_new_location: bool = False,
-    ) -> str:
+    def to_json(self, destination: str, if_exists: ZarrConflictResolution = "replace") -> str:
         """
         Save the dataset to a destination directory as a JSON file.
 
@@ -323,8 +318,6 @@ class BaseDataset(BaseArtifactModel, abc.ABC):
             destination: The _directory_ to save the associated data to.
             if_exists: Action for handling existing files in the Zarr archive. Options are 'raise' to throw
                 an error, 'replace' to overwrite, or 'skip' to proceed without altering the existing files.
-            load_zarr_from_new_location: Whether to update the current instance to load data from the location
-                the data is saved to. Only relevant for Zarr-datasets.
 
         Returns:
             The path to the JSON file.
@@ -362,3 +355,35 @@ class BaseDataset(BaseArtifactModel, abc.ABC):
 
     def __str__(self):
         return self.__repr__()
+
+    def _cache_zarr(self, destination: str | PathLike, if_exists: ZarrConflictResolution):
+        """
+        Caches the Zarr archive to a local directory.
+        """
+        from polaris.hub.storage import S3Store
+
+        if not self.uses_zarr:
+            return
+
+        current_zarr_root = Path(self.zarr_root_path)
+        destination_zarr_root = Path(destination) / current_zarr_root.name
+
+        # Copy over Zarr data to the destination
+        self._warn_about_remote_zarr = False
+
+        logger.info(f"Copying Zarr archive to {destination_zarr_root}. This may take a while.")
+        destination_store = zarr.open(str(destination_zarr_root), "w")
+        source_store = self.zarr_root.store.store
+
+        if isinstance(source_store, S3Store):
+            source_store.copy_to_destination(destination_store.store, if_exists, logger.info)
+        else:
+            zarr.copy_store(
+                source=self.zarr_root.store.store,
+                dest=destination_store.store,
+                log=logger.info,
+                if_exists=if_exists,
+            )
+        self.zarr_root_path = str(destination_zarr_root)
+        self._zarr_root = None
+        self._zarr_data = None
--- a/polaris/dataset/_base.py
+++ b/polaris/dataset/_base.py
@@ -1,7 +1,7 @@
 import abc
 import json
 from os import PathLike
-from pathlib import Path
+from pathlib import Path, PurePath
 from typing import Any, MutableMapping
 from uuid import uuid4
 
@@ -126,7 +126,7 @@ class BaseDataset(BaseArtifactModel, abc.ABC):
         Set the default cache dir if none and make sure it exists
         """
         if self._cache_dir is None:
-            self._cache_dir = str(Path(DEFAULT_CACHE_DIR) / _CACHE_SUBDIR / str(uuid4()))
+            self._cache_dir = str(PurePath(DEFAULT_CACHE_DIR) / _CACHE_SUBDIR / str(uuid4()))
         fs, path = fsspec.url_to_fs(self._cache_dir)
         fs.mkdirs(path, exist_ok=True)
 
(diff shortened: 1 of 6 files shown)


polaris/hub/storage.py
```diff
@@ -125,6 +129,101 @@ def _multipart_upload(self, key: str, value: bytes) -> None:
...
+
+        source_store_version = getattr(source, "_store_version", 2)
+        if source_store_version != self._store_version:
+            raise ValueError("zarr stores must share the same protocol version")
+
+        # The executor will allocate min(nbr_CPU + 4, 32) threads
+        with ThreadPoolExecutor() as executor:
+            total_copied = total_skipped = total_bytes_copied = 0
+
+            all_source_keys = sorted(source.keys())
```
cwognum, line 201: **Question:** Probably overkill for now, but are we worried that this may not fit in memory?

polaris/hub/storage.py
```diff
@@ -125,6 +129,101 @@ def _multipart_upload(self, key: str, value: bytes) -> None:
...
+
+            # Batch the keys, otherwise we end up with too many files open at the same time
+            # If we had Python 3.12, that would be itertools.batched, but alas...
+            batch_iter = iter(all_source_keys)
+            while batch := tuple(islice(batch_iter, self._batch_size)):
```
cwognum, line 206: **Compliment:** Nice! Had not seen `:=` before! Very succinct.

polaris/hub/storage.py
```diff
@@ -125,6 +129,101 @@ def _multipart_upload(self, key: str, value: bytes) -> None:
...
+                    else:
+                        raise
+            return copied, skipped, bytes_copied
+
+        source_store_version = getattr(source, "_store_version", 2)
+        if source_store_version != self._store_version:
+            raise ValueError("zarr stores must share the same protocol version")
+
+        # The executor will allocate min(nbr_CPU + 4, 32) threads
+        with ThreadPoolExecutor() as executor:
```
cwognum, line 198: **Question:** Don't want to scope creep too much here, but is there a simple way to add a progress bar of some sort? There is the "this may take a while" spinner, but that's annoyingly minimal for large datasets.

cwognum: I guess this can be done through the log function actually.

polaris/hub/client.py
cwognum, line 1: **Comment**: The other use of `copy_store()` is in the `cache()` method.

(side note: Maybe it would be better to have a `download` method in the client to centralize related logic)

Could we speed that up as well? Loading from the Hub will never be as fast as loading a dataset from local, and e.g. RxRx3-core is small enough to be stored locally on disk.

cwognum (commented):
Love it! I don't get exposed to the async part of Python enough. Thanks @jstlaurent !

polaris/hub/storage.py
jstlaurent: The `log(f"Copied {total_copied} ({total_bytes_copied} bytes), skipped {total_skipped}, of {len(all_source_keys)} keys")` line from below gets printed every batch. I've passed in `logger.info` as the log function, and it shows up in the console as the upload takes place.

We could display a nicer message if we wanted, with a percentage of progress and some formatting to the bytes number.

polaris/hub/storage.py
jstlaurent: That's a good question. I took this directly from the `copy_store` implementation, but we have to specific need for the keys to be sorted.

It ultimately depends on how `keys` is implemented in the source store. In our case, it's a `DirectoryStore`, so that method is a generator. Let's see what I can do...

polaris/hub/storage.py
cwognum: That works for now! Thanks!

jstlaurent:
@cwognum: Also implemented some optimizations when downloading a Zarr from Hub storage. Let me know what you think.

polaris/hub/storage.py
```diff
@@ -125,6 +129,149 @@ def _multipart_upload(self, key: str, value: bytes) -> None:
...
+    ## Custom methods
+
+    def copy_to_destination(
+        self, destination: Store, if_exists: ZarrConflictResolution = "replace", log: Callable = lambda: None
+    ) -> tuple[int, int, int]:
+        """
+        Copy the content of this store to the destination store.
+
+        We leverage the internal knowledge of this store to make the operation more efficient than `zarr.copy_store`:
+            - Parallel, concurrent `getitems` operations using a thread pool
```
cwognum, line 141: **Question:** I'm not seeing the parallelism here...? What am I missing? This code feels very sequential to me.

cwognum (approved):
Thanks again @jstlaurent! 🙏

polaris/dataset/_base.py
```diff
@@ -362,3 +355,35 @@ def __repr__(self):
...
+    def _cache_zarr(self, destination: str | PathLike, if_exists: ZarrConflictResolution):
+        """
+        Caches the Zarr archive to a local directory.
+        """
+        from polaris.hub.storage import S3Store
+
+        if not self.uses_zarr:
+            return
+
+        current_zarr_root = Path(self.zarr_root_path)
```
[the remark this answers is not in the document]
jstlaurent: It's true that here, `PurePath` would work because it's just stitching paths together. But according to the doc [`Path`](https://docs.python.org/3/library/pathlib.html#pathlib.Path), when instantiated, will be either a `PosixPath` or a `WindowsPath`, according to the platform the user is running on.

So for platform-agnostic file operations, `Path` is appropriate object to use.

There's a few places where we used [`str.split`](https://docs.python.org/3.10/library/stdtypes.html#str.split), with an assumption that `/` would be the path separator. That breaks on Windows.

polaris/dataset/_dataset.py
```diff
@@ -331,21 +265,75 @@ def __eq__(self, other):
...
+
+        Returns:
+            The path to the JSON file.
+        """
+        destination = Path(destination)
+        destination.mkdir(exist_ok=True, parents=True)
+
+        # Make a copy to cache the data, and then serialize that
+        copy = self.model_copy()
+        copy.cache(destination, if_exists=if_exists)
```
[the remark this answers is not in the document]
jstlaurent: `cache` now updates the `zarr_root_path` of the model instance to point to the now locally cached Zarr archive. Making a defensive copy of the model lets me do whatever I want to it, without messing up the instance the caller is manipulating.

polaris/hub/storage.py
jstlaurent: `getitems` is used to fetch multiple items concurrently and in parallel.

polaris/dataset/_base.py
```diff
@@ -362,3 +355,35 @@ def __repr__(self):
...
+        destination_zarr_root = Path(destination) / current_zarr_root.name
+
+        # Copy over Zarr data to the destination
+        self._warn_about_remote_zarr = False
+
+        logger.info(f"Copying Zarr archive to {destination_zarr_root}. This may take a while.")
+        destination_store = zarr.open(str(destination_zarr_root), "w")
+        source_store = self.zarr_root.store.store
```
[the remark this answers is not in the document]
jstlaurent: I agree, and that's not a bad idea. But that should be another PR.

polaris/hub/storage.py
```diff
@@ -125,6 +129,149 @@ def _multipart_upload(self, key: str, value: bytes) -> None:
...
+
+        Args:
+            destination: destination store
+            if_exists: behavior if the destination key already exists
+            log: optional logging function
+        """
+
+        destination_store_version = getattr(destination, "_store_version", 2)
+        if destination_store_version != self._store_version:
+            raise ValueError("zarr stores must share the same protocol version")
```
mercuryseries, line 155: Maybe we could create a custom exception since it's duplicated.

polaris/hub/storage.py
```diff
@@ -125,6 +129,149 @@ def _multipart_upload(self, key: str, value: bytes) -> None:
...
+                    result_copied, result_skipped, result_bytes_copied = future.result()
+                    total_copied += result_copied
+                    total_skipped += result_skipped
+                    total_bytes_copied += result_bytes_copied
+
+                log(
+                    f"Copied {total_copied} ({total_bytes_copied / (1024**2):.2f} MiB), skipped {total_skipped}, of {number_source_keys} keys. {(total_copied + total_skipped) / number_source_keys * 100:.2f}% completed."
+                )
+
+            return total_copied, total_skipped, total_bytes_copied
```
mercuryseries, line 271: This is the kind of code that makes you think, "Maybe I could do this for a living and actually enjoy it." 😅

mercuryseries (approved):
Thanks @jstlaurent! Looks great 💅. I’ve added a few minor comments here and there.

polaris/dataset/_base.py
```diff
@@ -362,3 +355,35 @@ def __repr__(self):
...
+
+        if isinstance(source_store, S3Store):
+            source_store.copy_to_destination(destination_store.store, if_exists, logger.info)
```
jstlaurent, line 379:
```suggestion
            source_store.copy_to_destination(destination_store, if_exists, logger.info)
```

polaris/dataset/_base.py
```diff
@@ -362,3 +355,35 @@ def __repr__(self):
...
+        else:
+            zarr.copy_store(
+                source=self.zarr_root.store.store,
+                dest=destination_store.store,
```
mercuryseries, line 383:
```suggestion
                dest=destination_store,
```

polaris/hub/storage.py
jstlaurent: Let's revisit when Zarr V3 is officially out. This code will change then.

Issue #212, opened 2024-10-30
Standardize artifact_type typing in _get_v1_dataset, _upload_v1_dataset and _upload_benchmark


mercuryseries:
# Context

In the `_get_v1_dataset`, `_upload_v1_dataset`, and `_upload_benchmark` methods, the `artifact_type` parameter is typed as `ArtifactSubtype`, but in many instances, we pass a string instead. This inconsistency can lead to confusion or potential issues.

# Description

To improve code consistency and maintainability, the `artifact_type` parameter should be standardized to consistently use `ArtifactSubtype` in `_get_v1_dataset`, `_upload_v1_dataset`, and `_upload_benchmark` methods. This will involve refactoring any instances where a string is passed instead of `ArtifactSubtype`. Additionally, any dependencies, tests, or documentation related to these methods should be reviewed to ensure compatibility and stability.

# Acceptance Criteria

- The `artifact_type` parameter is consistently typed as `ArtifactSubtype` in `_get_v1_dataset`, `_upload_v1_dataset`, and `_upload_benchmark`.
- All instances where a string is passed as `artifact_type` are updated to `ArtifactSubtype`.
- Code and comments are updated to reflect this change.
- All related tests pass without regression.

# Links

- https://github.com/polaris-hub/polaris/pull/210#discussion_r1823186707

PR #230, opened 2024-12-09
Misc changes

mercuryseries:
## Changelogs

- Fixes https://github.com/polaris-hub/polaris/issues/212
- Replace all occurences of `tmpdir` with `tmp_path`
- Update `list_datasets` and `list_benchmarks` to include support for v2 artifacts.

---

_Checklist:_

---


polaris/hub/client.py
```diff
@@ -287,17 +288,42 @@ def list_datasets(self, limit: int = 100, offset: int = 0) -> list[str]:
...
+            v2_response = self._base_request_to_hub(
+                url="/v2/dataset", method="GET", params={"limit": limit + offset, "offset": 0}
             )
-            response_data = response.json()
-            dataset_list = [bm["artifactId"] for bm in response_data["data"]]
+            v2_data = v2_response.json().get("data", [])
+            v2_datasets = [dataset["artifactId"] for dataset in v2_data]
+
+            # Apply offset and limit to v2 datasets
+            v2_datasets_offset = v2_datasets[offset : offset + limit]
```
cwognum, line 303: **Suggestion:** This shouldn't be needed! The API endpoint will already limit and offset the results

polaris/hub/client.py
```diff
@@ -287,17 +288,42 @@ def list_datasets(self, limit: int = 100, offset: int = 0) -> list[str]:
...
+
+            # If v2 datasets satisfy the limit, return them
+            if len(v2_datasets_offset) >= limit:
```
cwognum, line 306: **Comment:** Minor, but `len()` should never be _larger than_ the limit.

polaris/hub/client.py
```diff
@@ -407,18 +433,44 @@ def list_benchmarks(self, limit: int = 100, offset: int = 0) -> list[str]:
...
-            benchmarks_list = [f"{HubOwner(**bm['owner'])}/{bm['name']}" for bm in response_data["data"]]
+            v2_data = v2_response.json().get("data", [])
+            v2_benchmarks = [f"{HubOwner(**benchmark['owner'])}/{benchmark['name']}" for benchmark in v2_data]
+
+            # Apply offset and limit to v2 benchmarks
+            v2_benchmarks_offset = v2_benchmarks[offset : offset + limit]
+
+            # If v2 benchmarks satisfy the limit, return them
+            if len(v2_benchmarks_offset) >= limit:
+                return v2_benchmarks_offset
```
cwognum, line 452: **Comment:** Same suggestions here!

cwognum (approved):
Thanks, @mercuryseries !

polaris/hub/client.py
mercuryseries: It was me preparing for problems that might never happen (like I do in real life for no reason).

polaris/hub/client.py
mercuryseries: I made some tweaks.

mercuryseries: Good catch! I made some tweaks.

polaris/hub/client.py
```diff
@@ -288,17 +289,39 @@ def list_datasets(self, limit: int = 100, offset: int = 0) -> list[str]:
...
+
+            # If v2 datasets satisfy the limit, return them
+            if len(v2_datasets) == limit:
+                return v2_datasets
+
+            # Step 2: Calculate the remaining limit and fetch v1 datasets
+            remaining_limit = max(0, limit - len(v2_datasets))
+
+            v1_datasets = []
+            if remaining_limit > 0:
```
cwognum, line 311: **Comment:** This will always be true, because we return early if `len(v2_datasets) == limit`

polaris/hub/client.py
```diff
@@ -288,17 +289,39 @@ def list_datasets(self, limit: int = 100, offset: int = 0) -> list[str]:
...
+                v1_json_response = self._base_request_to_hub(
+                    url="/v1/dataset",
+                    method="GET",
+                    params={"limit": limit, "offset": max(0, offset - v2_json_response["metadata"]["total"])},
```
cwognum, line 315: **Suggestion:**
```suggestion
                    params={"limit": remaining_limit, "offset": max(0, offset - v2_json_response["metadata"]["total"])},
```

polaris/hub/client.py
```diff
@@ -288,17 +289,39 @@ def list_datasets(self, limit: int = 100, offset: int = 0) -> list[str]:
...
+                ).json()
+                v1_data = v1_json_response["data"]
+                v1_datasets = [dataset["artifactId"] for dataset in v1_data]
+
+            # Combine the v2 and v1 datasets
+            combined_datasets = v2_datasets + v1_datasets
+
+            # Ensure the final combined list respects the limit
+            return combined_datasets[:limit]
```
cwognum, line 324: **Suggestion:** If you implement my earlier suggestion, you won't need this. 

```suggestion
            return combined_datasets
```

polaris/hub/client.py
```diff
@@ -408,18 +431,41 @@ def list_benchmarks(self, limit: int = 100, offset: int = 0) -> list[str]:
...
+                v1_data = v1_json_response["data"]
+                v1_benchmarks = [
+                    f"{HubOwner(**benchmark['owner'])}/{benchmark['name']}" for benchmark in v1_data
+                ]
+
+            # Combine the v2 and v1 benchmarks
+            combined_benchmarks = v2_benchmarks + v1_benchmarks
+
+            # Ensure the final combined list respects the limit
+            return combined_benchmarks[:limit]
```
cwognum, line 468: **Suggestion**: Same here! 

- You can drop the if statement. 
- You can drop the `[:limit]` slice. 
- You don't need to initialize the `v1_benchmarks` as an empty list.

polaris/hub/client.py
```diff
@@ -288,17 +289,39 @@ def list_datasets(self, limit: int = 100, offset: int = 0) -> list[str]:
...
+            v2_datasets = [dataset["artifactId"] for dataset in v2_data]
+
+            # If v2 datasets satisfy the limit, return them
+            if len(v2_datasets) == limit:
+                return v2_datasets
+
+            # Step 2: Calculate the remaining limit and fetch v1 datasets
+            remaining_limit = max(0, limit - len(v2_datasets))
+
+            v1_datasets = []
```
cwognum, line 310: **Comment:** This is not needed.

```suggestion
```

cwognum (approved):
Thanks, @mercuryseries ! 

Why the many string conversions? The test cases weren't failing, so I assume it's to be consistent with typing? If so, what do you think about instead extending the type hints to also accept a `Path` ? That seems to actually be the more appropriate type here anyways.

mercuryseries:
> Why the many string conversions? The test cases weren't failing, so I assume it's to be consistent with typing? If so, what do you think about instead extending the type hints to also accept a `Path` ? That seems to actually be the more appropriate type here anyways.

Yes, I agree that extending the type hints to accept `Path` would be more appropriate. I considered doing that during the refactoring, but decided to hold off because the `datamol.utils.fs.join` function, which is used in some places, currently accepts a variable number of `str`: https://docs.datamol.io/0.7.15/api/datamol.utils.html#datamol.utils.fs.join. While it does internally convert input paths to `str` if you pass them in, the type hints don't reflect that. So, to properly support `Path` inputs, we'd need to update the type annotations of `join` and possibly other related functions.

cwognum:
@jstlaurent has repeatedly told me to not (ab)use Datamol for its filesystem... Old habits die hard... The use of `datamol.utils.fs.join` should be considered legacy code. Ideally, we should replace this with methods such as `os.path.join` (for local paths) or `urllib.parse.urljoin` (for URLs). I'm okay with leaving that outside the scope of this PR though.

PR #231, opened 2024-12-10
Feature: Client support for competitions

Andrewq11:
## Changelogs

- Client methods for downloading a competition from the Hub and submitting predictions 
- New `CompetitionSpecification` class that inherits from the `DatasetV2` class
- New `CompetitionPredictions` class that inherits from the `BenchmarkPredictions` class for standardized prediction uploads
- Removal of old competition code (e.g. differing URLs based on the `ArtifactSubtype`)
- Removal of the `CompetitionResults` class since these results never reach the client currently
- New method for loading a competition that accepts a local Zarr path for cached dataset downloads

---

_Checklist:_

---

## Important Notes
This PR is ready in terms of features, but is currently missing:
- Documentation/tutorials for these changes
- Updated tests

I will leave this PR as a draft to gather comments while I work on these next.


polaris/competition/_base.py
```diff
@@ -0,0 +1,229 @@
...
+    that in combination with the [`DatasetV2`][polaris.experimental._dataset_v2.DatasetV2] class, allow
+    users to participate in competitions hosted on Polaris Hub.
+
+    Examples:
+        Basic API usage:
+        ```python
+        import polaris as po
+
+        # Load the benchmark from the Hub
+        competition = po.load_competition("polaris/hello-world-competition")
```
Andrewq11, line 33: What are our thoughts on a never-ending `hello-world` competition hosted by us? Other competition sites do this and it is really good for onboarding newcomers to the feature.

polaris/competition/_base.py
```diff
@@ -0,0 +1,229 @@
...
+    split: SplitType
+    metrics: Union[str, Metric, list[str | Metric]]
+    main_metric: str | Metric
+    readme: str
+    target_types: dict[str, Union[TargetType, str]] = Field(default_factory=dict, validate_default=True)
+    start_time: datetime
+    end_time: datetime
+    n_test_sets: int
+    n_test_datapoints: dict[str, int]
+    n_classes: dict[str, int]
```
cwognum, line 89: **Comment**: Let's not do this now, but I do think we'll want to refactor this to inherit from some sort of `BaseBenchmark` or `BenchmarkMetadata` class. There's a lot of code duplication between this class and the benchmark class.

Let's revisit that once the `BenchmarkV2` is ready!

polaris/competition/_base.py
```diff
@@ -0,0 +1,229 @@
...
+    For additional meta-data attributes, see the [`DatasetV2`][polaris.experimental._dataset_v2.DatasetV2] class.
+    """
+
+    _artifact_type = "competition"
+
+    target_cols: ColumnsType
+    input_cols: ColumnsType
+    split: SplitType
+    metrics: Union[str, Metric, list[str | Metric]]
+    main_metric: str | Metric
```
cwognum, line 82: **Comment** FYI - These have changed in #223 and will have to be updated.

polaris/competition/_base.py
```diff
@@ -0,0 +1,229 @@
...
+    def _serialize_metrics(self, v) -> list[str] | str:
+        """Return the string identifier so we can serialize the object"""
+        if isinstance(v, Metric):
+            return v.name
+        return [m.name for m in v]
+
+    @field_serializer("target_types")
+    def _serialize_target_types(self, v) -> dict[str, str]:
+        """Convert from enum to string to make sure it's serializable"""
+        return {k: v for k, v in self.target_types.items()}
```
cwognum, line 123: **Question:** Is this needed? You're not converting anything are you?

polaris/competition/_base.py
```diff
@@ -0,0 +1,229 @@
...
+
+    @field_serializer("start_time", "end_time")
+    def _serialize_times(self, v: datetime) -> str:
+        """Convert from datetime to string to make sure it's serializable"""
+        return v.isoformat()
```
cwognum, line 128: **Suggestion:** I believe you can drop this. Pydantic got you covered! See e.g. [here](https://docs.pydantic.dev/2.2/usage/serialization/#modelmodel_dump_json)

```py
from datetime import datetime
from pydantic import BaseModel

class FooModel(BaseModel):
    foo: datetime

m = FooModel(foo=datetime(2032, 6, 1, 12, 13, 14))
print(m.model_dump_json())
#> {"foo":"2032-06-01T12:13:14"}
```

I do think we need to then switch to using `model_dump_json`, rather than `model_dump`, but I actually think our current use of `model_dump` is not correct!

polaris/competition/_base.py
```diff
@@ -0,0 +1,229 @@
...
+    @property
+    def test_set_labels(self) -> list[str]:
+        """The labels of the test sets."""
+        return sorted(list(self.n_test_datapoints.keys()))
+
+    @computed_field
+    @property
+    def test_set_sizes(self) -> dict[str, int]:
+        """The sizes of the test sets."""
+        return {k: v for k, v in self.n_test_datapoints.items()}
```
cwognum, line 147: **Suggestion:**

```suggestion
        return {k: len(v) for k, v in self.split[1].items()}
```

polaris/competition/_base.py
```diff
@@ -0,0 +1,229 @@
...
+    def task_type(self) -> str:
+        """The high-level task type of the benchmark."""
+        v = TaskType.MULTI_TASK if len(self.target_cols) > 1 else TaskType.SINGLE_TASK
+        return v.value
+
+    @computed_field
+    @property
+    def test_set_labels(self) -> list[str]:
+        """The labels of the test sets."""
+        return sorted(list(self.n_test_datapoints.keys()))
```
cwognum, line 141: **Suggestion:**

```suggestion
        return sorted(list(self.split[1].keys()))
```

polaris/competition/_base.py
```diff
@@ -0,0 +1,229 @@
...
+        n_test_datapoints: The number of datapoints in each test set.
+        n_classes: The number of classes within target columns that define a classification task.
+    For additional meta-data attributes, see the [`DatasetV2`][polaris.experimental._dataset_v2.DatasetV2] class.
+    """
+
+    _artifact_type = "competition"
+
+    target_cols: ColumnsType
+    input_cols: ColumnsType
+    split: SplitType
```
cwognum, line 80: **Fix:** You'll need a serializer and validator for the split!

cwognum (approved):
Thanks @Andrewq11 ! 

I left some comments. Nothing major!

polaris/competition/_base.py
Andrewq11: Good points. Once we better understand how we will evaluate results, and if there is some possibility of auto-eval in our future, we could return to this idea.

For now, I'll update that artifact ID to a placeholder so users aren't hit with a 404 if they copy/paste it.

polaris/evaluate/_predictions.py
```diff
@@ -177,3 +188,35 @@ def _is_fully_specified(
...
+        owner: A slug-compatible name for the owner of the artifact. It is redeclared here to be required.
+        report_url: A URL to a report/paper/write-up which describes the methods used to generate the predictions.
+        github_url: A URL to a code repository which contains the code used to generate the predictions.
+        contributors: A list of Polaris usernames representing those who contributed to producing the predictions.
+    """
+
+    _artifact_type = "competition-prediction"
+
+    name: SlugCompatibleStringType
+    owner: HubOwner
```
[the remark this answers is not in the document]
Andrewq11: Could you see people using the `CompetitionPredictions` object for something other than uploading their predictions to the Hub? There isn't much more functionality within the class that users can leverage. 

I'm not against making the `name`, `owner` and `report_url` field optional on the client and mandatory on the server. In my view, it just depends on if we think there is a use case for that object pre-upload.

polaris/loader/load.py
```diff
@@ -94,3 +96,22 @@ def load_benchmark(path: str, verify_checksum: ChecksumStrategy = "verify_unless
...
+    Note: Dataset is automatically loaded
+        If no argument is passed for `zarr_root_path`, the underlying competition dataset will be automatically
+        fetched from the Polaris Hub. If it is passed, the dataset will be retrieved locally from the location
+        specified in the passed value.
+
+    """
+
+    # Load from the Hub
+    client = PolarisHubClient()
+    return client.get_competition(*slug.split("/"), zarr_root_path=zarr_root_path)
```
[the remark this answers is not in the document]
Andrewq11: This is true, but I was looking at it from a data loading perspective. We have some network overhead with V2 datasets, and I think users would appreciate being able to cache a dataset on first retrieval and then use that local cache for more efficient retrieval on subsequent script invocations.

What are your thoughts on that?

polaris/hub/client.py
```diff
@@ -797,114 +774,75 @@ def _upload_benchmark(
...
+        self, owner: str | HubOwner, name: str, zarr_root_path: str | None = None
     ) -> CompetitionSpecification:
         """Load a competition from the Polaris Hub.
 
         Args:
             owner: The owner of the competition. Can be either a user or organization from the Polaris Hub.
             name: The name of the competition.
-            verify_checksum: Whether to use the checksum to verify the integrity of the dataset.
+            zarr_root_path: An optional path that specifies the cache location of the competition's dataset
+                on your local filesystem.
```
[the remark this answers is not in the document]
Andrewq11: Ok, I think I see where some confusion is coming from. I test the competition flow via a simple python script that calls `.load_competition(...)`, and then I invoke this numerous times while creating predictions, tweaking things, etc.

Because I use a simple python script, it recreates the competition object each time, redirecting the `zarr_root_path` back to the `PolarisFS` one. 

Do you typically use a Jupyter notebook and run the `.cache()` call in a separate cell?

polaris/hub/client.py
```diff
@@ -797,114 +774,75 @@ def _upload_benchmark(
...
-
-        # TODO (jstlaurent): response["dataset"]["artifactId"] is the owner/name unique identifier,
-        #  but we'd need to change the signature of get_dataset to use it
-        response_data["dataset"] = self._get_v1_dataset(
-            response_data["dataset"]["owner"]["slug"],
-            response_data["dataset"]["name"],
-            ArtifactSubtype.COMPETITION,
-            verify_checksum=verify_checksum,
-        )
+        url = f"/v1/competition/{owner}/{name}"
```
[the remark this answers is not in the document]
Andrewq11: I can't remember the exact rationale right now, but my feeling is that it was incorrect. I don't have a compelling reason for using `v2` when there isn't a `v1` to break yet.

polaris/hub/client.py
```diff
@@ -797,114 +774,75 @@ def _upload_benchmark(
...
-            success_msg="Evaluated competition predictions.",
-            error_msg="Failed to evaluate competition predictions.",
+            start_msg="Submitting competition predictions...",
+            success_msg="Submitted competition predictions.",
+            error_msg="Failed to submit competition predictions.",
         ) as progress_indicator:
-            competition.owner = HubOwner.normalize(competition.owner)
-
+            #
+            # Prepare prediction payload for submission
```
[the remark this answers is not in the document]
Andrewq11: Lol I only use it when a comment is sandwiched between a function signature and line of code. I think it makes those comments a bit more readable, but happy to hold back if we veto it.

polaris/competition/_base.py
```diff
@@ -0,0 +1,229 @@
...
+        prediction_values = np.array([0.0 for x in test])
+        predictions = CompetitionPredictions(
+            name="first-prediction",
+            owner=HubOwner(slug="dummy-user"),
+            report_url="REPORT_URL",
+            target_labels=competition.target_cols,
+            test_set_labels=competition.test_set_labels,
+            test_set_sizes=competition.test_set_sizes,
+            predictions=prediction_values,
+        )
```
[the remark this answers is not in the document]
Andrewq11: I agree with you and that would be ideal, but what about all the other metadata which someone can apply to a prediction? E.g. `github_url`, `contributors`, `tags`, `user_attributes`, `description`. 

It could make the `submit_predictions` signature quite long if we added all of that. What do you think?

polaris/competition/_base.py
Andrewq11: Agreed 👍

polaris/competition/_base.py
Andrewq11: I believe you're merging this today? I'll update this branch once you do.

polaris/competition/_base.py
Andrewq11: This is a remnant of the first version where there was no split on the client side. These achieve the same result though, don't they?

polaris/competition/_base.py
Andrewq11: Same comment from me as here! https://github.com/polaris-hub/polaris/pull/231#discussion_r1880532372

polaris/competition/_base.py
Andrewq11: For my understanding, why are they needed?

Serializer: If we define each competition specification ourselves, and we choose to use a python list (as we usually do), why is a serializer needed?

Validator: If we define each competition ourselves, can't we assume that the split is valid on retrieval?

I'm not against adding them, I just assumed that we were ok with less validation because we define/upload the competitions on our side.

polaris/evaluate/_predictions.py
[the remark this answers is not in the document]
cwognum: I don't feel strongly either way here. I mostly comment because I do think it's important to be consistent if we can to simplify maintenance. 

Bit of a tangent, but I think the UX for validation ahead of upload can be improved. The Zod errors we now surface to the client are rather cryptic.

polaris/loader/load.py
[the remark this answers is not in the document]
cwognum: I see the value, but I don't think this is the right way to implement it. For the sake of simplicity and consistency (we don't support this for normal datasets either), I would leave this out for now.

polaris/hub/client.py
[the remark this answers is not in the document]
cwognum: > Do you typically use a Jupyter notebook and run the .cache() call in a separate cell?

Yes, indeed. I do think this is valuable, but I think implementing it properly would be a little more complex than this. I could see this easily be confusing to the user. 

Hence my recommendation to leave it out for the time being.

polaris/hub/client.py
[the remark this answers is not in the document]
cwognum: Makes sense to me too!

polaris/hub/client.py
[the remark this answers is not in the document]
cwognum: All good. Just thought it was funny! 😄

polaris/competition/_base.py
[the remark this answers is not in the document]
cwognum: Yeah, I agree, but I think it's the better option. 

We could also simplify the method signature with a single metadata object (rather than several parameters), but that just feels like an unnecessary level of indirection?

polaris/competition/_base.py
cwognum: Yes, merged now!

polaris/competition/_base.py
cwognum: Yes, but logically it makes more sense to consider the split as the ground truth, since that actually affects the benchmarking logic.

polaris/competition/_base.py
cwognum: Even if it's just us uploading, the data models are quite complex and the validators have helped me spot errors a few times already. Especially since we have the code for the validation already, I think it will make our lives easier!

polaris/evaluate/_predictions.py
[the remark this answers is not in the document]
Andrewq11: Agreed. The cryptic errors is one, and the inconsistent formatting/highlighting of error messages can be improved too. It would be really nice to create a package for updating this messaging in the new year.

polaris/loader/load.py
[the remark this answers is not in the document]
Andrewq11: I'll concede to that, I think there are classier ways that this can be done.

polaris/competition/_base.py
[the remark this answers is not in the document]
Andrewq11: Given that the primary way users will submit predictions is likely through that method, then I agree with having the longer signature. Our docs and tutorials will point to it too.

Andrewq11:
@cwognum going to mark this ready for review despite missing those couple docs and tutorial. Everything else should be ready 💪

polaris/competition/_base.py
```diff
@@ -0,0 +1,360 @@
...
+    HttpUrlString,
+    SlugCompatibleStringType,
+    PredictionsType,
+    HubUser,
+)
+
+ColumnsType: TypeAlias = str | list[str]
+
+
+class CompetitionSpecification(DatasetV2):
```
jstlaurent, line 36: **Comment**
Feels weird that `DatasetV2` is experimental, but this is not. Should we move `DatasetV2` to a "regular" package?

jstlaurent, line 36: **Question**
Would it make sense to have this inherit from `BenchmarkSpecification` (or even the new `BenchmarkV1Specification`)? It's basically a combination of dataset and benchmark.

polaris/hub/client.py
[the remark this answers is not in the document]
jstlaurent: Agree, this is much cleaner and easier to understand.

polaris/loader/load.py
```diff
@@ -94,3 +95,22 @@ def load_benchmark(path: str, verify_checksum: ChecksumStrategy = "verify_unless
         benchmark.verify_checksum()
 
     return benchmark
+
+
+def load_competition(slug: SlugCompatibleStringType):
```
jstlaurent, line 100: **Suggestion**
This is not correct. Unfortunately, Python typing can't do the string template types we have in TypeScript. So this needs to be a `str`.
```suggestion
def load_competition(artifactId: str):
```

polaris/hub/client.py
```diff
@@ -797,114 +775,72 @@ def _upload_benchmark(
...
                 f"View it here: {urljoin(self.settings.hub_url, url)}"
             )
             return response_data
 
     def get_competition(
-        self,
-        owner: str | HubOwner,
-        name: str,
-        verify_checksum: ChecksumStrategy = "verify_unless_zarr",
+        self, owner: str | HubOwner, name: SlugCompatibleStringType
```
jstlaurent, line 790: **Suggestion**
Start using the artifact identifier as is, since that's what users are going to be using.

```suggestion
        self, artifactId: str
```

jstlaurent (approved):
Overall, nice work. I did have some questions/suggestions. Since Andrew is out, I can take a stab at implementing them.

cwognum:
@jstlaurent If you have the time, it would be great if you could take a stab at this. I was planning to take over from Andrew, but there are some other things on my plate I would really like to wrap before EOY so we can hit the ground running in 2025.

polaris/competition/_base.py
cwognum: Yes! It may requires some shuffling of what's part of the base class and what's not, but I do think that is the better approach here. How do you feel about inheriting the V2 benchmark already?

cwognum: Yes, I think it's time. 

Would require some documentation, though. Happy to work on that! Especially if you could work on this PR.

polaris/benchmark/_split.py
```diff
@@ -0,0 +1,115 @@
...
+    split: SplitType
+
+    @model_validator(mode="after")
+    def _validate_split(self) -> Self:
+        """
+        Verifies that:
+          1) There are no empty test partitions
+          2) There is no duplicate indices in any of the sets
+          3) There is no overlap between the train and test set
+          4) No row exists in the test set where all labels are missing/empty
```
Andrewq11, line 37: Since we've decided not to check for empty test set rows due to potential dataset size/download constraints, let's remove this line.

polaris/benchmark/_split.py
```diff
@@ -0,0 +1,114 @@
...
+    @property
+    def n_train_datapoints(self) -> int:
+        """The size of the train set."""
+        return len(self.split[0])
+
+    @computed_field
+    @property
+    def test_set_labels(self) -> list[str]:
+        """The labels of the test sets."""
+        return sorted(list(self.split[1].keys()))
```
Andrewq11, line 105: Any reason for the `sorted` here?

polaris/experimental/_split_v2.py
```diff
@@ -0,0 +1,172 @@
...
+    def serialize(self) -> bytes:
+        return self.indices.serialize()
+
+    @staticmethod
+    def deserialize(index_set: bytes) -> "IndexSet":
+        return IndexSet(indices=BitMap.deserialize(index_set))
+
+
+class SplitV2(BaseModel):
+    training: IndexSet
```
Andrewq11, line 58: Definitely a nitpick, but we use `train` in our docs and other places when referring to the training set. Should we do the same here instead of `training`?

cwognum: It's a fair question, but given the timeline I prefer not to make any last-minute changes if it's not absolutely needed.

polaris/benchmark/_split.py
cwognum: Hm, not sure! Doesn't seem needed. I might have had a reason for this which I have since then forgotten... Maybe for the checksum?

polaris/benchmark/_task.py
```diff
@@ -0,0 +1,144 @@
...
+            v = [v]
+
+        def _convert(m: str | dict | Metric) -> Metric:
+            if isinstance(m, str):
+                return Metric(label=m)
+            if isinstance(m, dict):
+                return Metric(**m)
+            return m
+
+        v = [_convert(m) for m in v]
```
jstlaurent, line 83: **Suggestion**
The `if isinstance(v, str)` above will make sure there's never a `str`here. So this will do the same:
```python
v = [Metric(**m) if isinstance(m, dict) else m for m in v]
```

jstlaurent (approved):
Minor nitpick. Looks good. 🚢 it!

polaris/benchmark/_task.py
cwognum: I don't think that works. You can still have a list of strings!

PR #234, opened 2024-12-16
feat: Support for handling Benchmarks V2

jstlaurent:
## Changelogs

- Make `BenchmarkSpecification` into an abstract class, with `BenchmarkV1Specification` and `BenchmarkV2Specification` implementations. The abstract class defines a series of abstract methods and properties 
for the subclasses to implement.
- Implemented a new type of split for Benchmark V2, using the [Roaring Bitmap](https://roaringbitmap.org/) format to efficiently store extremely large sets of indicices.
- Improved the type definition in the Pydantic models, to reflect the post-validation types. Where we are flexible on the input type, validators with `mode="before"` ensure that we convert the input to a 
single internal representation.
- Added upload/dowload methods to the Hub client, to support interacting with the Hub storage. 

---

_Checklist:_

---

The core changes are ready for review. I need to add some tests and ensure the doc is updated as well.


polaris/benchmark/_base.py
```diff
@@ -100,47 +98,43 @@ class BenchmarkSpecification(BaseArtifactModel, ChecksumMixin):
     _artifact_type = "benchmark"
 
     # Public attributes
-    # Data
-    dataset: Union[DatasetV1, CompetitionDataset, str, dict[str, Any]]
-    target_cols: ColumnsType
-    input_cols: ColumnsType
-    split: SplitType
+    dataset: BaseArtifactModel = Field(exclude=True)
```
cwognum, line 101: **Suggestion:**

```suggestion
    dataset: BaseDataset = Field(exclude=True)
```

polaris/benchmark/_base.py
```diff
@@ -188,153 +180,54 @@ def _validate_main_metric_is_in_metrics(self) -> Self:
...
 
-        train_idx_list = split[0]
-        full_test_idx_list = list(chain.from_iterable(split[1].values()))
+    @field_serializer("metrics")
+    def _serialize_metrics(self, value: set[Metric]) -> list[Metric]:
+        """
+        Convert the set to a list. Since metrics are models and will be converted to dict,
+        they will not be hashable members of a set.
+        """
+        return list(value)
```
cwognum, line 200: **Comment:** This is outside the scope of this PR, but I believe we may not be using Pydantic to its full potential here. Pydantic differentiates two `mode`s for its [`model_dump`](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_dump), i.e. `json` and `python`. I wonder if this conversion could be done automatically if we specify the `json` mode? And more generally, the use of these two modes seems like something we can be more rigorous about.

polaris/experimental/_benchmark_v2.py
```diff
@@ -0,0 +1,247 @@
...
+        Accepts an initial sequence of ints, and turn it into a BitMap
+        """
+        if isinstance(v, BitMap):
+            return v
+        return BitMap(v)
+
+    @computed_field
+    @cached_property
+    def datapoints(self) -> int:
+        return len(self.indices)
```
cwognum, line 43: **Suggestion:** Minor, but I would rename to `n_datapoints`.

polaris/experimental/_benchmark_v2.py
```diff
@@ -0,0 +1,247 @@
...
+        # TODO: Until we support multi-test benchmarks
+        yield "test", self.test
+
+
+class BenchmarkV2Specification(BenchmarkSpecification):
+    _version: ClassVar[Literal[2]] = 2
+
+    dataset: DatasetV2 = Field(exclude=True)
+    split: SplitV2
+    n_classes: dict[ColumnName, int]
```
cwognum, line 142: **Question:** Since `n_classes` is in both the V1 and V2, could there be a way to define it in the base class? Or is that hard to do given that the implementations differ (i.e. `computed_property` vs. a regular field)

polaris/experimental/_benchmark_v2.py
```diff
@@ -0,0 +1,247 @@
...
+    ) -> dict[str, Subset]:
+        """
+        Construct the test set(s), given the split in the benchmark specification. Used
+        internally to construct the test set for client use and evaluation.
+        """
+        # TODO: We need a subset class that can handle very large index sets without copying or materializing all of them
+        return {
+            label: self._get_subset(index_set.indices, hide_targets, featurization_fn)
+            for label, index_set in self.split.test_items()
+        }
```
cwognum, line 223: **Comment:** So this works out of the box with the subset class? That's pretty cool! 😎

polaris/hub/client.py
```diff
@@ -801,14 +823,53 @@ def _upload_benchmark(
...
+    ):
+        with ProgressIndicator(
+            start_msg="Uploading artifact...",
+            success_msg="Uploaded artifact.",
+            error_msg="Failed to upload benchmark.",
+        ) as progress_indicator:
+            # Get the serialized data-model
+            # We exclude the dataset as we expect it to exist on the hub already.
+            benchmark.owner = HubOwner.normalize(owner or benchmark.owner)
+            benchmark_json = benchmark.model_dump(exclude={"dataset"}, exclude_none=True, by_alias=True)
```
cwognum, line 847: **Comment:** Isn't it excluded by default?

```suggestion
            benchmark_json = benchmark.model_dump(exclude_none=True, by_alias=True)
```

tests/test_benchmark.py
```diff
@@ -140,70 +140,71 @@ def test_benchmark_from_json(test_single_task_benchmark, tmp_path):
...
-@pytest.mark.parametrize("is_single_task", [True, False])
-def test_benchmark_checksum(is_single_task, test_single_task_benchmark, test_multi_task_benchmark):
+@pytest.mark.parametrize("fixture", ["test_single_task_benchmark", "test_multi_task_benchmark"])
+def test_benchmark_checksum(fixture, request):
     """Test whether the checksum captures our notion of equality."""
 
-    obj = test_single_task_benchmark if is_single_task else test_multi_task_benchmark
-    cls = SingleTaskBenchmarkSpecification if is_single_task else MultiTaskBenchmarkSpecification
-    metrics_list = list(obj.metrics)
+    benchmark = request.getfixturevalue(fixture)
```
cwognum, line 147: **Comment:** Nice! 💅

cwognum (approved):
Thanks @jstlaurent and @mercuryseries ! This is looking good! 🚀

polaris/hub/client.py
jstlaurent: Good catch!

polaris/experimental/_benchmark_v2.py
jstlaurent: It does. I do have concerns about how it would handle truly large index sets. I tested the rest with a test/training split of ~1 million indices each. I need to test the Subset class to see how well it handles it.

polaris/experimental/_benchmark_v2.py
jstlaurent: I'm using this to match the the Hub-side data structure.

polaris/experimental/_benchmark_v2.py
jstlaurent: That's technically possible, but it would add a lot of boilerplate.

polaris/benchmark/__init__.py
```diff
@@ -1,11 +1,9 @@
...
-from polaris.benchmark._definitions import (
-    MultiTaskBenchmarkSpecification,
-    SingleTaskBenchmarkSpecification,
-)
+from polaris.benchmark._base import BenchmarkSpecification, BenchmarkV1Specification
+from polaris.benchmark._definitions import MultiTaskBenchmarkSpecification, SingleTaskBenchmarkSpecification
 
 __all__ = [
     "BenchmarkSpecification",
+    "BenchmarkV1Specification",
```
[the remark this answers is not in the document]
jstlaurent: Good question. I'd leave it as it, because V1 benchmarks were exposed through `SingleTaskBenchmarkSpecification` and `MultiTaskBenchmarkSpecification`.

polaris/benchmark/_base.py
jstlaurent: Serialization works correctly with `mode="json"`, even with this omitted. The thing is, these are still Pydantic models, so we should make sure Python mode for `model_dump` also works. This feels like a good middle ground.

polaris/hub/client.py
```diff
@@ -695,18 +716,17 @@ def _upload_v2_dataset(
...
                 logger.info("Copying the dataset manifest file to the Hub.")
                 with open(dataset.zarr_manifest_path, "rb") as manifest_file:
-                    storage.set_manifest(manifest_file.read())
+                    storage.set_file("manifest", manifest_file.read())
 
                 # Step 3: Upload the Zarr archive
                 logger.info("Copying Zarr archive to the Hub. This may take a while.")
 
-                destination = storage.root_store
+                destination = storage.store("root")
```
mercuryseries, line 729: Nice refactoring 💅

mercuryseries (approved):
Looks good to me, @jstlaurent! I added some minor comments here and there.

```python
try:
    while True:
        print("Thank you! (Until Julien is tired)")
except KeyboardInterrupt:
    print("Gratitude interrupted, but it’s endless in spirit! :)")
```

tests/test_benchmark_v2.py
```diff
@@ -0,0 +1,152 @@
...
+        "dataset": test_dataset_v2,
+        "split": SplitV2(
+            training=IndexSet(indices=BitMap([0, 1])),
+            test=IndexSet(indices=BitMap([max_index, max_index + 1])),  # Invalid indices
+        ),
+        "target_cols": ["B"],
+        "input_cols": ["A"],
+        "metrics": ["accuracy", "r2"],
+        "main_metric": "accuracy",
+        "n_classes": {"target1": 2},
```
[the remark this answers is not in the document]
jstlaurent: Good catch, thank you!

PR #243, opened 2025-01-10
Add custom codecs for RDKit Molecules and Biotite AtomArrays

cwognum:
## Changelogs

- Add a custom object codec for RDKit Molecules and Biotite AtomArrays

---

_Checklist:_

- [ ] ~_Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR._~
- [ ] ~_Add tests to cover the fixed bug(s) or the newly introduced feature(s) (if appropriate)._~
- [X] _Update the API documentation if a new function is added, or an existing one is deleted._
- [X] _Write concise and explanatory changelogs above._
- [X] _If possible, assign one of the following labels to the PR: `feature`, `fix`, `chore`, `documentation` or `test` (or ask a maintainer to do it for you)._

---

This PR builds on the discussion in #241. 

The use of custom `numcodecs`-compatible object codecs for use in Zarr seems like a natural interface point to document and implement conversion from and to drug discovery specific format. 

You would simply use this like: 

```py
import zarr
import datamol as dm
from polaris.dataset.zarr._codecs import RDKitMolCodec

root = zarr.open("test.zarr", mode="w")
root.empty("molecules", shape=100, dtype=object, object_codec=RDKitMolCodec(), chunks=(5,))
root["molecules"][0] = dm.to_mol("C1=CC=CC=C1")
```

PR #244, opened 2025-01-13
chore: Improve data loading performance in S3 store

jstlaurent:
## Changelogs

- Add a LRU cache to the `S3Store` class, to cache fetched objects
- Add chunk-based iterators, to improve performance when traversing dataset columns

---

_Checklist:_

---

This builds on the work Cas did in `feat/optimize-dataloader`. I increased the number of cached objects in the S3 store, since most of them are small anyway. I also added a Generator-based chunking iterator, that I think is simpler than the class-based approach. It should reduce the amount of decompression that takes place when iterating through a dataset row-by-row.

### DIFF
--- a/polaris/hub/storage.py
+++ b/polaris/hub/storage.py
@@ -2,6 +2,7 @@ import re
 from base64 import b64encode
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from contextlib import contextmanager
+from functools import lru_cache
 from hashlib import md5
 from itertools import islice
 from pathlib import PurePath
@@ -128,6 +129,14 @@ class S3Store(Store):
             self.s3_client.complete_multipart_upload(
                 Bucket=self.bucket_name, Key=full_key, UploadId=upload_id, MultipartUpload={"Parts": parts}
             )
+
+    @lru_cache()
+    def _get_object_body(self, full_key: str) -> bytes:
+        """
+        Basic caching for the object body, to avoid multiple reads on remote bucket.
+        """
+        response = self.s3_client.get_object(Bucket=self.bucket_name, Key=full_key)
+        return response["Body"].read()
 
     ## Custom methods
 
@@ -351,8 +360,7 @@ class S3Store(Store):
         with handle_s3_errors():
             try:
                 full_key = self._full_key(key)
-                response = self.s3_client.get_object(Bucket=self.bucket_name, Key=full_key)
-                return response["Body"].read()
+                return self._get_object_body(full_key=full_key)
             except self.s3_client.exceptions.NoSuchKey:
                 raise KeyError(key)
 
@@ -429,6 +437,12 @@ class S3Store(Store):
             page_iterator = paginator.paginate(Bucket=self.bucket_name, Prefix=self.prefix)
 
             return sum((page["KeyCount"] for page in page_iterator))
+
+    def __hash__(self):
+        """
+        Custom hash function, to enable lru_cache decorator on methods
+        """
+        return hash((self.bucket_name, self.prefix, self.s3_client))
 
 
 class StorageTokenAuth:
--- a/polaris/dataset/_base.py
+++ b/polaris/dataset/_base.py
@@ -2,7 +2,7 @@ import abc
 import json
 from os import PathLike
 from pathlib import Path, PurePath
-from typing import Any, MutableMapping
+from typing import Any, Iterable, MutableMapping
 from uuid import uuid4
 
 import fsspec
@@ -232,7 +232,7 @@ class BaseDataset(BaseArtifactModel, abc.ABC):
 
     @property
     @abc.abstractmethod
-    def rows(self) -> list[str | int]:
+    def rows(self) -> Iterable[str | int]:
         """Return all row indices for the dataset"""
         raise NotImplementedError
 
--- a/polaris/dataset/_dataset_v2.py
+++ b/polaris/dataset/_dataset_v2.py
@@ -2,7 +2,7 @@ import json
 import re
 from os import PathLike
 from pathlib import Path
-from typing import Any, ClassVar, Literal
+from typing import Any, ClassVar, Iterable, Literal
 
 import fsspec
 import numpy as np
@@ -82,19 +82,19 @@ class DatasetV2(BaseDataset):
     def n_rows(self) -> int:
         """Return all row indices for the dataset"""
         example = self.zarr_root[self.columns[0]]
-        if isinstance(example, zarr.Group):
-            return len(example[_INDEX_ARRAY_KEY])
-        return len(example)
+        match example:
+            case zarr.Group():
+                return len(example[_INDEX_ARRAY_KEY])
+            case _:
+                return len(example)
 
     @property
-    def rows(self) -> np.ndarray[int]:
-        """Return all row indices for the dataset
-
-        Warning: Memory consumption
-            This feature is added for completeness' sake, but it should be noted that large datasets could consume a lot of memory.
-            E.g. storing a billion indices with np.in64 would consume 8GB of memory. Use with caution.
+    def rows(self) -> Iterable[int]:
         """
-        return np.arange(len(self), dtype=int)
+        Return all row indices for the dataset
+        This feature is added for completeness' sake, but does not provide any performance benefits.
+        """
+        return range(self.n_rows)
 
     @property
     def columns(self) -> list[str]:


polaris/dataset/_iterator.py
```diff
@@ -0,0 +1,222 @@
...
+import numpy as np
+from typing_extensions import Self, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from polaris.dataset import DatasetV2
+
+SequenceContent: TypeAlias = Any
+
+
+def chunk_iterator(
```
cwognum, line 13: **Suggestion**: It would be nice to use this in the `Subset` class as well, as this is the actual class people would use for data access.

This poses two challenges: 
- The Subset needs to skip certain indices based on its `indices` field.
- Data transformations and data loading are strongly coupled in the Subset class.

polaris/dataset/_iterator.py
```diff
@@ -0,0 +1,222 @@
...
+    """
+    rng = np.random.default_rng(random_state)
+    chunk_indices = list(range(0, total_size, chunk_size))
+
+    if shuffle_chunks:
+        rng.shuffle(chunk_indices)
+
+    for start_idx in chunk_indices:
+        end_idx = min(start_idx + chunk_size, total_size)
+        chunk = content[start_idx:end_idx]
```
cwognum, line 45: **Comment**: This lacks the application of the Adapter, if any.

Two thoughts here:

- In the dataset classes, we should decouple the data loading from the data transformations.
- However, with https://github.com/polaris-hub/polaris/pull/243, I actually think we should deprecate the Adapter. We can achieve the same effect more seamlessly through the custom object codecs. In that case, Zarr handles the data transformations.

cwognum, line 45: **Comment**: This will also break when the column is a Zarr Group rather than a Zarr Array.

Referring to https://github.com/polaris-hub/polaris/pull/243 again, I think we no longer need groups. We can simplify our code (and our lives), if we assume the Zarr archive is not nested (i.e. the root solely has arrays as children). Building on our discussion from last week, I can't think of a case in which we would need granular access into a Zarr archive beyond that first level, meaning we could always use a custom object codec.

polaris/dataset/_iterator.py
```diff
@@ -0,0 +1,222 @@
...
+
+        if shuffle_within_chunk:
+            rng.shuffle(chunk)
+
+        for element in chunk:
+            yield element
+
+
+class ChunkedIterator:
```
cwognum, line 54: **Comment:** Your intention was to delete this and the `CachedChunkIterator`, I assume?

One reason I decided to have both is that the `ChunkedIterator` would make for a good interface point with ML-frameworks, e.g. the `sampler` in the Pytorch Lightning [DataLoader](https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader).

polaris/dataset/_iterator.py
```diff
@@ -0,0 +1,222 @@
...
+        random_state: The random seed to use for shuffling.
+
+    Yields:
+        Any: The next element in the content, shuffled as indicated.
+    """
+    rng = np.random.default_rng(random_state)
+    chunk_indices = list(range(0, total_size, chunk_size))
+
+    if shuffle_chunks:
+        rng.shuffle(chunk_indices)
```
cwognum, line 41: **Comment:** I had a reason not to use `shuffle`, but I forgot the precise reason why. I believe that for a dataset of the size of BELKA, shuffling the chunk rather than shuffling the indices (i.e. `rng.permutation`) was notably slower.

cwognum (changes requested):
I explored this a little further today (but also in a new branch, given the merge conflicts) and I was just about to open a PR for [`feat/optimize-data-access`](https://github.com/polaris-hub/polaris/tree/feat/optimize-data-access). You beat me to it! 😄 

We should merge in the caching layer for the `S3Store` as soon as possible, but the custom iterators need some further thinking. I've left some comments throughout the PR, but this is raising some higher level thoughts for me: 

1. Given #243, I think we could deprecate the `Adapter` and `DatasetFactory`as well as constrain the structure of the Zarr archive to not have any subgroups. That would likely be a breaking change, but I think it's worth it.
2. At several places, we mix data loading with data transformations, which makes it hard to load data from a cached chunk.

On a bit of a tangent, I found the following performance penalties for each _chunk_ access: 
- Fetching the data from the cloud bucket.
- Copying the data to a NumPy Array. 
- Decompressing the data. 

As we examined, the decompression added minimal overhead, but the data copies added notable overhead. Wonder if this has been sped up in Zarr V3. See also: https://github.com/zarr-developers/zarr-python/issues/1395

polaris/hub/storage.py
```diff
@@ -129,6 +130,14 @@ def _multipart_upload(self, key: str, value: bytes) -> None:
                 Bucket=self.bucket_name, Key=full_key, UploadId=upload_id, MultipartUpload={"Parts": parts}
             )
 
+    @lru_cache()
```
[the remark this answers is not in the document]
jstlaurent: There's a default of 128 entries. Setting it to 0 disables the eviction and lets the cache grow indefinitely.

polaris/dataset/_iterator.py
jstlaurent: The Subset class probably doesn't work for DatasetV2 as it stands. It makes too many assumptions that it's fine to load everything in memory.

jstlaurent: I can give a shot at creating a variant for V2.

polaris/hub/storage.py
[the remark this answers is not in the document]
cwognum: I was reading the wrong docs! My bad!

polaris/dataset/_iterator.py
jstlaurent: I voluntarily stripped this function down to accepting a `Sequence` type, which would work with a Zarr `Array`, but not a Zarr `Group`. Groups are `MutableMapping`s, which are not `Sequence`s. Granted, that wouldn't stop someone from passing one in, but there's only so much one can do.

cwognum: The `iter_rows` method could actually pass in a group!

jstlaurent:
> We should merge in the caching layer for the `S3Store` as soon as possible, but the custom iterators need some further thinking. I've left some comments throughout the PR, but this is raising some higher level thoughts for me:
> 
>     1. Given [Add custom codecs for RDKit Molecules and Biotite AtomArrays #243](https://github.com/polaris-hub/polaris/pull/243), I think we could deprecate the `Adapter` and `DatasetFactory`as well as constrain the structure of the Zarr archive to not have any subgroups. That would likely be a breaking change, but I think it's worth it.
> 
>     2. At several places, we mix data loading with data transformations, which makes it hard to load data from a cached chunk. 

I don't know if we specifically need to constrain the structure, but we do mix loading and transformations in a very ad-hoc way. That makes things more complicated. I do think leveraging object codecs to make the Zarr layer handle transformations lets us build a simpler access layer.

> On a bit of a tangent, I found the following performance penalties for each _chunk_ access:
> 
>     * Fetching the data from the cloud bucket.
> 
>     * Copying the data to a NumPy Array.
> 
>     * Decompressing the data.
> 
> 
> As we examined, the decompression added minimal overhead, but the data copies added notable overhead. Wonder if this has been sped up in Zarr V3. See also: [zarr-developers/zarr-python#1395](https://github.com/zarr-developers/zarr-python/issues/1395)

We can work on the fetching part, but the other two are inside Zarr-Python and out of our control. 

I stumbled upon [this repo](https://github.com/developmentseed/obstore) that might be an interesting way to speed up the S3 access layer.

There might be a way to leverage [Rust-based codec pipeline](https://github.com/ilan-gold/zarrs-python) in Zarr-Python. Not as good as pure Rust, but the [benchmarks](https://github.com/LDeakin/zarr_benchmarks) show some interesting performance improvements. Probably the best we can do

polaris/dataset/_iterator.py
jstlaurent: I checked and we have six datasets that use adapters. Maybe a use case for the new versioning system?

polaris/dataset/_iterator.py
jstlaurent: Not a bad idea. We should probably be more intentional about it, however.

cwognum (approved):
Thanks, @jstlaurent !

PR #245, opened 2025-01-14
chore: Add a CI job to run tests in Conda, to detect dependency failure earlier

jstlaurent:
## Changelogs

- Add a GitHub Actions job to run the test suite under a Conda environment
- Update the dependency list by removing unused ones

---

_Checklist:_

---

Since we've been tripped by some dependendencies that don't have all platforms supported in Conda, adding a CI flow that will run the tests from a clean Conda environment should raise these issues earlier.

### DIFF
--- a/.github/workflows/code-check.yml
+++ b/.github/workflows/code-check.yml
@@ -2,8 +2,8 @@ name: code-check
 
 on:
   push:
-    branches: ["main"]
-    tags: ["*"]
+    branches: [ "main" ]
+    tags: [ "*" ]
   pull_request:
     branches:
       - "*"
@@ -18,17 +18,17 @@ jobs:
       - name: Checkout the code
         uses: actions/checkout@v4
 
-      - name: Set up Python
-        uses: actions/setup-python@v4
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
         with:
-          python-version: "3.12"
+          enable-cache: true
+          cache-dependency-glob: ./pyproject.toml
 
-      - name: Install ruff
-        run: |
-          pip install ruff
+      - name: Install the project
+        run: uv sync --group dev
 
       - name: Lint
-        run: ruff check .
+        run: uv run ruff check
 
       - name: Format
-        run: ruff format . --check
+        run: uv run ruff format . --check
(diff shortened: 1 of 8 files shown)


.github/workflows/test.yml
```diff
@@ -12,12 +12,12 @@ on:
     - cron: "0 4 * * MON"
 
 jobs:
-  test:
+  test-uv:
```
cwognum, line 15: **Suggestion**: Minor, but can we change the name?

```
name: PyPi os=${{ matrix.os }} - python=${{ matrix.python-version }}

**Review by cwognum — APPROVED** [review]
Thanks! 🙏 

Random thought: As the CICD continues to grow, is there value in changing to a different service (e.g. I think I've seen @Andrewq11 use CodeFresh). I have never actually worked with such services, so I'm not sure it's worth it.

**Review by Andrewq11 — APPROVED** [review]
This is awesome, thanks @jstlaurent.

It would've been nice to do this test on conda-forge's CI infrastructure. That was the original thought, but apparently we cannot reliably get access to non-Linux runners for testing there. 

No issue though, this will be great for avoiding any further upstream dependency hiccups 🚀

File: .github/workflows/code-check.yml [hunk 2]
```diff
@@ -18,17 +18,17 @@ jobs:
       - name: Checkout the code
         uses: actions/checkout@v4
 
-      - name: Set up Python
-        uses: actions/setup-python@v4
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
```
mercuryseries, line 22:
```suggestion
        uses: astral-sh/setup-uv@v5
```

.github/workflows/test.yml
Andrewq11: > is there value in changing to a different service (e.g. I think I've seen @Andrewq11 use CodeFresh)
I feel as though GitHub Actions works well for our use cases and we don't have a real need/blocker which suggests we should take on the burden of onboarding/managing a new infra tool right now. 

CodeFresh is nice, but I don't see any extra value it could provide us on Polaris.

mercuryseries (approved):
Thanks, @jstlaurent!

PR #249, opened 2025-01-17
fix: Prevent external AWS configuration from breaking our boto3 config

jstlaurent:
## Changelogs

- Explicitly set the region when instantiating a boto3 client, to prevent it from picking one up from some external AWS config

---

_Checklist:_

- [ ]~ _Was this PR discussed in an issue? It is recommended to first discuss a new feature into a GitHub issue before opening a PR._~

---

Boto3 can automatically pick up external configuration for AWS, from various places in the environment: env variables, config files in user directory, etc. This recently caused a problem where it picked up a 
region from a user that was not compatible with the Cloudflare R2 bucket we use. Explicitly setting the region when the client is created will avoid this issue in the future.

### DIFF
--- a/polaris/hub/storage.py
+++ b/polaris/hub/storage.py
@@ -83,6 +83,7 @@ class S3Store(Store):
             aws_secret_access_key=secret_key,
             aws_session_token=token,
             endpoint_url=endpoint_url,
+            region_name="auto",
         )
         self.bucket_name = bucket_name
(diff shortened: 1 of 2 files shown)


cwognum (approved):
Thanks, @jstlaurent ! 🙏

mercuryseries (approved):
Thanks @jstlaurent !

Issue #251, opened 2025-01-23
Cannot po.load_dataset due to an unknown authorization error


VladVin:
### Polaris version

0.11.3

### Python Version

3.11

### Operating System

Mac

### Installation

pip install polaris-lib

### Description

During downloading a dataset of ASAP challenge on a freshly installed `polaris-lib` I faced an error when polaris tried to authenticate my request. This error isn't descriptive, and I have to go to the polaris's codebase to understand what happened.

### Steps to reproduce

Instructions to reproduced (just a copy from the Try it out [here](https://polarishub.io/datasets/asap-discovery/antiviral-admet-2025-sample)):
```python
import polaris as po

# Load the dataset from the Hub
dataset = po.load_dataset("asap-discovery/antiviral-admet-2025-sample")

# Get information on the dataset size
dataset.size()

# Load a datapoint in memory
dataset.get_data(
    row=dataset.rows[0],
    col=dataset.columns[0],
)

# Or, similarly:
dataset[dataset.rows[0], dataset.columns[0]]

# Get an entire data point
dataset[0]
```

### Additional output

The error:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 2
      1 # Load the dataset from the Hub
----> 2 dataset = po.load_dataset("asap-discovery/antiviral-admet-2025-sample")
      4 # Get information on the dataset size
      5 dataset.size()

File ~/miniconda3/envs/chem/lib/python3.11/site-packages/polaris/loader/load.py:37, in load_dataset(path, verify_checksum)
     34 if not is_file:
     35     # Load from the Hub
     36     with PolarisHubClient() as client:
---> 37         client.ensure_active_token()
     38         return client.get_dataset(*path.split("/"), verify_checksum=verify_checksum)
     40 # Load from local file

File ~/miniconda3/envs/chem/lib/python3.11/site-packages/polaris/hub/client.py:158, in PolarisHubClient.ensure_active_token(self, token)
    155         return True
    157 # Check if external token is still valid, or we're using password auth
--> 158 if not (self.has_user_password or self.external_client.ensure_active_token()):
    159     return False
    161 # If so, use it to get a new Hub token

File ~/miniconda3/envs/chem/lib/python3.11/site-packages/polaris/hub/external_client.py:87, in ExternalAuthClient.ensure_active_token(self, token)
     85     token = self.token
     86 try:
---> 87     return super().ensure_active_token(token) or False
[...]

PR #252, opened 2025-01-23
fix: Better handling when no cached token exists, with more decriptive error

jstlaurent:
## Changelogs

- Make sure the `ensure_active_token` methods will identify that there is no token
- Surface a better error message when it happens, so users can take action to remediate

---

_Checklist:_

---

Adresses #251

This bug fix will better handle cases where the OAuth token is `None`. This might result from a new user that has not logged in, tokens that were not cached, or cached credentials that have 
expired. When the client is used in the loader functions, and one such situation occurs, ane exception will be raised that includes a better, actionable error message for the user.

### DIFF
--- a/polaris/loader/load.py
+++ b/polaris/loader/load.py
@@ -7,7 +7,6 @@ from polaris.benchmark import MultiTaskBenchmarkSpecification, SingleTaskBenchma
 from polaris.dataset import DatasetV1, create_dataset_from_file
 from polaris.experimental._benchmark_v2 import BenchmarkV2Specification
 from polaris.hub.client import PolarisHubClient
-from polaris.utils.errors import PolarisUnauthorizedError
 from polaris.utils.types import ChecksumStrategy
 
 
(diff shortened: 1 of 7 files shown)


polaris/loader/load.py
```diff
@@ -34,7 +35,8 @@ def load_dataset(path: str, verify_checksum: ChecksumStrategy = "verify_unless_z
     if not is_file:
         # Load from the Hub
         with PolarisHubClient() as client:
-            client.ensure_active_token()
+            if not client.ensure_active_token():
+                raise PolarisUnauthorizedError()
```
cwognum, line 39: **Question:** Could we move this into the `__enter__()` method of the `PolarisHubClient`?

cwognum (approved):
Thanks, @jstlaurent !

polaris/loader/load.py
jstlaurent: I'd need to restructure how we do the login flow, because the CLI does this right now:
```python
with PolarisHubClient(settings=PolarisHubSettings(_env_file=client_env_file)) as client:
    client.login(auto_open_browser=auto_open_browser, overwrite=overwrite)
```

cwognum: I see! In that case I'm ok with leaving it as is for now!

jstlaurent: I do agree it would make for a nicer user experience. Let me see what I can do.

mercuryseries (approved):
Thanks, @jstlaurent!

jstlaurent:
Moved the check into the context manager's `_enter_` method, so it's more systematic.

cwognum (approved):

Issue #255, opened 2025-01-25
Exception when fetching a dataset


PatWalters:
### Polaris version

0.11.4

### Python Version

3.11

### Operating System

Linux

### Installation

uv pip install polaris-lib

### Description

```python
import polaris as po

benchmark = po.load_benchmark("polaris/hello-world-benchmark")
```
```
⠸ Fetching dataset.....
2025-01-25 15:09:11.549 | INFO     | polaris._artifact:_validate_version:90 - The version of Polaris that was used to create the artifact (0.0.0) is different from the currently installed version of Polaris (0.11.4).
2025-01-25 15:09:11.550 | INFO     | polaris.mixins._checksum:verify_checksum:65 - To verify the checksum, we need to recompute it. This can be slow for large datasets.
✔ SUCCESS: Fetched dataset.
2025-01-25 15:09:11.622 | INFO     | polaris._artifact:_validate_version:90 - The version of Polaris that was used to create the artifact (0.0.0) is different from the currently installed version of Polaris (0.11.4).
2025-01-25 15:09:11.624 | WARNING  | polaris.mixins._checksum:verify_checksum:58 - No checksum to verify against. Specify either the md5sum parameter or store the checksum in the dataset.md5sum attribute.
✔ SUCCESS: Fetched benchmark.
Error in callback <function Halo.__init__.<locals>.clean_up at 0x7b9d474dfd80> (for post_run_cell), with arguments args (<ExecutionResult object at 7b9dd6b2bf10, execution_count=2 error_before_exec=None error_in_exec=None info=<ExecutionInfo object at 7b9d4a182150, raw_cell="benchmark = po.load_benchmark("polaris/hello-world.." store_history=True silent=False shell_futures=True cell_id=52c65924-b9fc-46a8-af17-11ec3772e016> result=None>,),kwargs {}:
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
TypeError: Halo.__init__.<locals>.clean_up() takes 0 positional arguments but 1 was given
[...]

PR #256, opened 2025-01-27
fix: Replace Halo with rich for CLI spinner and progress tracking

jstlaurent:
## Changelogs

- Use [rich](https://github.com/Textualize/rich) to track progress, instead of Halo
- Use rich for logging, instead of Loguru, to harmonize CLI output
- Scope logging configuration to the library
- Remove Loguru and Halo as dependencies

Adresses issue #255

---

_Checklist:_

---

When used in iPython, Halo raises an exception. The bug has been patched in their repo secven months ago, but no new release has been created since. The last release is more than four years old. 

In order to fix this, I elected to use [rich](https://github.com/Textualize/rich), a CLI formatting library that has progress trackers. It is up-to-date and well maintained. I replaced the 
`ProgressIndicator` class with a context manager function that wraps a `Progress` object from `rich`. In addition to a spinner, we also get progress bars. By using a ContextVar, we can nest 
multiple `track_progress` context manager, to track multiple tasks.

In order not to break the visual output in the CLI, I also used the logging handler provided by `rich`, replacing the Loguru library we had been using. I also took the opportunity to scope the 
logging configuration to the Polaris library logging only, to avoid changing any config the user might have set up in their code.


polaris/utils/context.py
```diff
@@ -1,31 +1,61 @@
...
+    BarColumn,
+    Progress,
+    SpinnerColumn,
+    TaskProgressColumn,
+    TextColumn,
+    TimeElapsedColumn,
+)
+
+# Singleton Progress instance to be used for all calls to `track_progress`
+progress_instance = ContextVar(
```
cwognum, line 15: **Comment:** Seems that the `ContextVar` has been around since 3.7, but I've never seen it before! 

Let me see if I understand its use here correctly: By defining this as a `ContextVar`, we get a different `rich.progress.Progress` instance for each time we invoke `track_progress` since this has been decorated with a `@contextmanager`.

Would replacing: 
```py
progress = progress_instance.get()
```

by:
```py
progress = Progress(
    SpinnerColumn(),
    TextColumn("[progress.description]{task.description}"),
    BarColumn(),
    TaskProgressColumn(),
    TimeElapsedColumn(),
)
```

be the same thing?

cwognum (approved):
Thanks @jstlaurent ! 

Could you show a screenshot of what it now looks like?

polaris/utils/context.py
jstlaurent: No, it's actually the exact opposite. 😉 

By using `ContextVar`, we get back the _same_ `Progress` instance every time we call `track_progress`.

cwognum: Ah, I should've inferred that from the comments. Nice way to implement a Singleton!

I was thrown off by the `@contextmanager` since that seemed to indicate to me that we are switching contexts! 😅

cwognum:
Thanks for the screenshot @jstlaurent ! Looks great to me!

PR #260, opened 2025-01-31
chore: Update skip condition for integration tests

jstlaurent:
## Changelogs

- Update the skip check to skip if either required settings are missing
---

_Checklist:_

---

The previous check would pass if at least one of username or password was set, and the test would fail.

### DIFF
--- a/polaris/hub/settings.py
+++ b/polaris/hub/settings.py
@@ -34,7 +34,9 @@ class PolarisHubSettings(BaseSettings):
     """
 
     # Configuration of the pydantic model
-    model_config = SettingsConfigDict(env_file=".env", env_prefix="POLARIS_", extra="ignore")
+    model_config = SettingsConfigDict(
+        env_file=".env", env_prefix="POLARIS_", extra="ignore", env_ignore_empty=True
+    )
 
     # Hub settings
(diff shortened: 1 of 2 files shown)


cwognum (approved):
Thanks @jstlaurent ! I'm surprised that this has been working locally but now fails in the CICD. Is this needed because `os.getenv()` returns an empty string rather than `None` for PRs opened by external contributors? To predict our CICD secrets, presumably?

jstlaurent:
> Thanks @jstlaurent ! I'm surprised that this has been working locally but now fails in the CICD. Is this needed because `os.getenv()` returns an empty string rather than `None` for PRs opened by external contributors? To predict our CICD secrets, presumably?

Yep. When a PR comes from a fork, the secrets are present as environment variables, but they contain an empty string. Why GitHub chose this, I honestly don't understand... When using `os.getenv()`, the variable exists, so the `None` default value does not get returned. And since we test for `is None`, the test passes.

I updated the Pydantic settings class to `env_ignore_empty=True`, so empty values will use the default of `None` set in the class for these environment variables.

If we need to expose secrets to external contributors, we can create deployment environments and invite folks there. That way, we can control who gets access. I don't think we need this right now, but it's an option for the future.

PR #265, opened 2025-02-03
feat: Support for Model artifacts

mercuryseries:
## Changelogs

- Add a new `Model` class to represent a model artifact.
- Add a new `load_model` utility function to load a model.
- 
---

_Checklist:_

---

## TODO

- Add some tests


polaris/model/__init__.py
```diff
@@ -0,0 +1,20 @@
+from polaris._artifact import BaseArtifactModel
+from polaris.utils.types import HttpUrlString
+from polaris.utils.types import AccessType, HubOwner
+
+
+class Model(BaseArtifactModel):
```
cwognum, line 6: **Suggestion:** Could we add a docstring here with some details about this class? And could we add this class to our `mkdocs` documentation site?

polaris/evaluate/_results.py
cwognum, line 1: **Suggestion:** I would move the `model` attribute to the `ResultsMetadata` class.

cwognum, line 1: **Comment:** We want to treat the `model` field differently for V1 and V2 results. For V1 results, we don't want to support the `model` field at all, whereas for V2 results we want to replace the `github_url` and `paper_url` by the new `model` attribute. We will need to create a V1 and V2 implementation of results on the client side to support this.

polaris/evaluate/_results.py
```diff
@@ -151,6 +152,7 @@ class BenchmarkResults(EvaluationResult):
     benchmark_artifact_id: str | None = Field(None)
     benchmark_name: SlugCompatibleStringType | None = Field(None, deprecated=True)
     benchmark_owner: HubOwner | None = Field(None, deprecated=True)
+    model: Model | None = Field(None)
```
cwognum, line 155: **Suggestion:** Exclude the Model from the Pydantic dump of this class, similar to how this is done for Datasets in the Benchmark classes. 

```suggestion
    model: Model | None = Field(None, exclude=True)
```

polaris/hub/client.py
cwognum, line 1: **Question**: When retrieving a result from the Hub, should we automatically load the associated Model from the Hub?

cwognum (approved):
Thanks @mercuryseries ! 🙏

cwognum:
@mercuryseries For the failing test case, looks like you were just unlucky! We should set the random seed for that test case so that doesn't happen again.

polaris/hub/client.py
```diff
@@ -889,3 +890,50 @@ def submit_competition_predictions(
                 json=prediction_payload,
             )
             return response
+
+    def get_model(self, artifact_id: str) -> Model:
+        url = f"/v1/model/{artifact_id}"
```
jstlaurent, line 895: **Suggestion**
I'm open to push back on this, but I'm going to suggest that we add new endpoints as `v2` in the API. I think that's going to be less confusing when an artifact is going to be associated with other V2 artifacts.

polaris/model/__init__.py
jstlaurent: Good idea.

polaris/hub/client.py
jstlaurent: Since the proposed `Model` is pretty light, we could return it whole, inline with the results. Pydantic will parse it back to a `Model` instance when loading a `Result`.

Of course, if the `Model` evolves into something a lot more complex, then that might not longer be the best way to go. I think for now, it should work.

polaris/evaluate/_results.py
jstlaurent: That would also impact `CompetitionPredictions`.

jstlaurent (commented):
Short and sweet. Very nice. 😄 

I do agree with Cas' comment, about making this only available on Results V2, for Benchmark V2s.

cwognum (approved):
LGTM! Thanks @mercuryseries ! Nice job on splitting it in a V1 and V2.

polaris/benchmark/_base.py
```diff
@@ -35,6 +35,9 @@
     TargetType,
 )
 
+# Type variable for the return type of evaluate
+BenchmarkResultsType = TypeVar("BenchmarkResultsType")
```
Andrewq11, line 39: Nice addition with this! I'd suggest:

```suggestion
BenchmarkResultsType = TypeVar("BenchmarkResultsType", BenchmarkResultsV1, BenchmarkResultsV2)
```

So we can get some extra type safety and bind `BenchmarkResultsType` to either `BenchmarkResultsV1` or `BenchmarkResultsV2`.

polaris/evaluate/_metadata.py
```diff
@@ -32,3 +33,27 @@ def _repr_html_(self) -> str:
...
+
+    Attributes:
+        model: The model that was used to generate these results.
+        contributors: The users that are credited for these results.
+
+    For additional meta-data attributes, see the base classes.
+    """
+
+    # Additional meta-data
+    model: Model | None = Field(None, exclude=True)
```
Andrewq11, line 49: Would having information about the model used to generate a result be useful when serializing and instance of this class? For example, to keep it slim, could we implement a custom serializer for the `model` field here which just returned `model_owner/model_slug`?

polaris/hub/client.py
```diff
@@ -889,3 +890,50 @@ def submit_competition_predictions(
                 json=prediction_payload,
             )
             return response
+
+    def get_model(self, artifact_id: str) -> Model:
+        url = f"/v2/model/{artifact_id}"
```
Andrewq11, line 895: Is `v2` the correct choice given that we haven't had a `v1` for model artifacts yet? I've been guilty of flip-flopping on this in the past, so I just want to raise the question and confirm.

polaris/hub/client.py
```diff
@@ -889,3 +890,50 @@ def submit_competition_predictions(
...
+        owner: HubOwner | str | None = None,
+    ):
+        """Upload a model to the Polaris Hub.
+
+        Info: Owner
+            You have to manually specify the owner in the model data model. Because the owner could
+            be a user or an organization, we cannot automatically infer this from just the logged-in user.
+
+        Note: Required meta-data
+            The Polaris client and hub maintain different requirements as to which meta-data is required.
```
Andrewq11, line 914: Sorry for the nitpick: We usually capitalize 'hub' in our tutorials. I'd suggest the same for here!

polaris/hub/client.py
```diff
@@ -889,3 +890,50 @@ def submit_competition_predictions(
...
+            The requirements by the hub are stricter, so when uploading to the hub you might
+            get some errors on missing meta-data. Make sure to fill-in as much of the meta-data as possible
```
Andrewq11, line 916: Another nitpick (sorry): "meta-data" -> "metadata" to agree with your format in the `Model` class docstring.

uv.lock
```diff
@@ -451,7 +451,7 @@ name = "click"
 version = "8.1.8"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "colorama", marker = "sys_platform == 'win32'" },
+    { name = "colorama", marker = "platform_system == 'Windows'" },
```
Andrewq11, line 454: Out of curiosity, what required the change from `sys_platform` to `platform_system`?

Andrewq11 (approved):
Really nice @mercuryseries, thanks for this! 🚀 

Left a few minor comments but LGTM overall.

PR #266, opened 2025-02-06
[UNFINISHED] feat: Artifact Versioning

roselynh100:
## Changelogs

- `@model_validator` to accept a `slug` from the Hub (if passed), otherwise `slugify`s the artifact name
- Created function `build_urn` to construct an artifact urn
- Added version fields `artifact_version` and `artifact_changelog` to datasets
- `upload_dataset` accepts a flag `upload_as_new_version`, which defaults to False
  - If the user tries to upload a dataset with a name they've already used (without specifying `upload_as_new_version`), an error will be thrown

---

_Checklist:_

---


polaris/hub/client.py
```diff
@@ -672,12 +675,17 @@ def _upload_v1_dataset(
...
+                    "uploadAsNewVersion": upload_as_new_version,
                     **dataset_json,
                 },
                 timeout=timeout,
             )
 
-            with StorageSession(self, "write", dataset.urn) as storage:
+            inserted_dataset = response.json()
+
+            inserted_dataset_urn = build_urn("dataset", inserted_dataset["owner"]["slug"], inserted_dataset["slug"])
```
Andrewq11: *Note for team*: We need to do this because the slug is determined on the backend for new artifact versions.

polaris/_artifact.py
```diff
@@ -126,4 +126,4 @@ def to_json(self, path: str) -> None:
 
     @classmethod
     def urn_for(cls, owner: str | HubOwner, name: str) -> ArtifactUrn:
-        return f"urn:polaris:{cls._artifact_type}:{owner}:{slugify(name)}"
+        return build_urn(cls._artifact_type, owner, slugify(name))
```
Andrewq11, line 129: *Suggestion*: The URN will be incorrect once a new artifact version is uploaded (because the `name` and `slug` will differ). 

This actually might require a bit of changes in this class. We compute the slug property based on the name, which is no longer correct. I suggest we do this to allow for the slug to be returned by the Hub:

```
    _slug: SlugStringType | None = None # allow the slug to be provided by the Hub

    @field_validator("_slug")
    @classmethod
    def _validate_slug(cls, value: SlugStringType | None):
        # If slug not retrieved from Hub (e.g. created locally)
        if isinstance(value, None):
            # If name provided
            if cls.name:
                return slugify(cls.name)
            else:
                return None
        return value
```

We should then update line 129 to:
`return build_urn(cls._artifact_type, owner, cls.slug)`

Andrewq11 (commented):
Great work so far, @roselynh100 🚀 Left a couple comments but this is well on its way.

polaris/hub/client.py
```diff
@@ -672,12 +675,19 @@ def _upload_v1_dataset(
...
+                    "uploadAsNewVersion": upload_as_new_version,
                     **dataset_json,
                 },
                 timeout=timeout,
             )
 
-            with StorageSession(self, "write", dataset.urn) as storage:
+            inserted_dataset = response.json()
+
+            inserted_dataset_urn = build_urn(
```
Andrewq11, line 686: Just doing some more thinking on this. 

Instead of building the `urn` here, could you update the `dataset` object's `_slug`? Then, on line 690, you should be able to access `dataset.urn` as we did before, and it will create the correct `urn` automatically.

polaris/_artifact.py
```diff
@@ -53,11 +53,14 @@ class BaseArtifactModel(BaseModel):
...
+    slug: Annotated[Optional[SlugStringType], Field(validate_default=True)] = None
 
-    @computed_field
-    @property
-    def slug(self) -> SlugStringType | None:
-        return slugify(self.name) if self.name else None
+
+    @field_validator("slug")
+    def _validate_slug(cls, val: Optional[str], info) -> SlugStringType | None:
+        if val is None:
```
Andrewq11, line 61: Can we add a comment here briefly stating when this might be `None`? (e.g. when it is a new artifact created locally).

roselynh100:
These changes have been approved, but I'll close this PR for now and use `feat/artifact-versioning` as a feature branch. Will merge into main later!

PR #270, opened 2025-02-20
feat: Artifact Versioning - upload benchmark

roselynh100:
## Changelogs

- Added version fields `artifact_version` and `artifact_changelog` to benchmarks
- `upload_benchmark` accepts an optional parameter `parent_artifact_id` (owner/slug), which defaults to None
  - The benchmark will be uploaded as a subsequent version of the provided parent
  - This functionality is different from the previous `upload_as_new_version` param implemented in `feat/artifact-versioning`, and `upload_datasets` has been updated accordingly
- Modified dataset and benchmark `upload_to_hub` functions to also accept `parent_artifact_id`
- Added info about `parent_artifact_id` and uploading new versions of datasets and benchmarks to the documentation

---

_Checklist:_

---

### DIFF
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -638,6 +638,9 @@ class PolarisHubClient(OAuth2Client):
             )
 
             inserted_dataset = response.json()
+
+            # We modify the slug in the server
+            # Update dataset.slug here so dataset.urn is constructed correctly
             dataset.slug = inserted_dataset["slug"]
 
             with StorageSession(self, "write", dataset.urn) as storage:
@@ -703,6 +706,9 @@ class PolarisHubClient(OAuth2Client):
             )
 
             inserted_dataset = response.json()
+
+            # We modify the slug in the server
+            # Update dataset.slug here so dataset.urn is constructed correctly
             dataset.slug = inserted_dataset["slug"]
 
             with StorageSession(self, "write", dataset.urn) as storage:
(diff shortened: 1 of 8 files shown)


polaris/hub/client.py
```diff
@@ -796,6 +799,7 @@ def _upload_v1_benchmark(
             benchmark: The benchmark to upload.
             access: Grant public or private access to result
             owner: Which Hub user or organization owns the artifact. Takes precedence over `benchmark.owner`.
+            parent_artifact_id: The `owner/slug` of the parent benchmark, if uploading a new version of a benchmark.
```
Andrewq11, line 802: Since this is acting as a private method and this docstring exists in the calling `upload_benchmark` method, can we remove it from here? We can replace this docstring with "`Upload a V1 benchmark to the Polaris Hub.`" (similar to what we have for datasets). 

We could also add a similar, small docstring for `_upload_v2_benchmark`.

polaris/hub/client.py
```diff
@@ -830,13 +837,21 @@ def _upload_v2_benchmark(
...
+                json={
+                    "access": access,
+                    "datasetArtifactId": benchmark.dataset.artifact_id,
+                    "parentArtifactId": parent_artifact_id,
+                    **benchmark_json,
+                },
             )
 
+            uploaded_benchmark = response.json()
+            benchmark.slug = uploaded_benchmark["slug"]
```
Andrewq11, line 853: Can we add a small comment regarding why this line is important. If someone accidentally removes it, the underlying `benchmark.urn` will be wrong and it could be hard to identify why.

Andrewq11 (approved):
Very nice, @roselynh100 🚀 Thanks for all of this, very clean. Left just a couple comments.

PR #273, opened 2025-02-27
Draft: First pass at a BenchmarkV2 that can hold multiple splits.

cwognum:
This is currently a draft. Do not merge!

### DIFF
--- a/polaris/benchmark/_base.py
+++ b/polaris/benchmark/_base.py
@@ -156,7 +156,7 @@ class BenchmarkSpecification(PredictiveTaskSpecificationMixin, BaseArtifactModel
 @deprecated(
     "Use BenchmarkV2Specification instead. If you're loading this dataset from the Polaris Hub, you can ignore this warning."
 )
-class BenchmarkV1Specification(SplitSpecificationV1Mixin, ChecksumMixin):
+class BenchmarkV1Specification(SplitSpecificationV1Mixin, ChecksumMixin, BenchmarkSpecification):
     _version: ClassVar[Literal[1]] = 1
 
     dataset: DatasetV1 = Field(exclude=True)
--- a/polaris/benchmark/_base.py
+++ b/polaris/benchmark/_base.py
@@ -3,12 +3,11 @@ import json
 from hashlib import md5
 from itertools import chain
 from pathlib import Path
-from typing import Any, Callable, ClassVar, Literal, TypeVar, Generic
+from typing import Any, Callable, ClassVar, Literal
 
 import fsspec
 import numpy as np
 from pydantic import (
-    BaseModel,
     Field,
     computed_field,
     field_validator,
@@ -22,11 +21,10 @@ from polaris.benchmark._split import SplitSpecificationV1Mixin
 from polaris.benchmark._task import PredictiveTaskSpecificationMixin
 from polaris.dataset import DatasetV1, Subset
 from polaris.dataset._base import BaseDataset
-from polaris.evaluate import BenchmarkResultsV1, BenchmarkResultsV2
+from polaris.evaluate import BenchmarkResultsV1
 from polaris.evaluate.utils import evaluate_benchmark
 from polaris.hub.settings import PolarisHubSettings
 from polaris.mixins import ChecksumMixin
-from polaris.utils.dict2html import dict2html
 from polaris.utils.errors import InvalidBenchmarkError
 from polaris.utils.types import (
     AccessType,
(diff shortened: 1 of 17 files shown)


polaris/benchmark/_split_v2.py
```diff
@@ -89,88 +91,32 @@ def _validate_test_set(cls, v: IndexSet) -> IndexSet:
...
 
     @model_validator(mode="after")
     def validate_set_overlap(self) -> Self:
         """
         The training and test index sets do not overlap
         """
-        if self.training.intersect(self.test):
-            raise InvalidBenchmarkError("The predefined split specifies overlapping train and test sets")
+        if self.train.intersect(self.test):
+            raise InvalidSplitError("This split specifies overlapping training and test sets")
```
cwognum, line 103: **Nitpick**: Now that we can have multiple splits and now that we have a label, we should use it in the logging.

```suggestion
            raise InvalidSplitError(f"The {self.label} split specifies overlapping training and test sets")
```

polaris/benchmark/_benchmark_v2.py
```diff
@@ -87,70 +76,23 @@ def get_train_test_split(
...
-                If there are multiple test sets, these are returned in a dictionary and each test set has
-                an associated name. The targets of the test set can not be accessed.
-        """
-        train = self._get_subset(
-            self.split.training.indices, hide_targets=False, featurization_fn=featurization_fn
-        )
-        test = self._get_test_sets(hide_targets=True, featurization_fn=featurization_fn)
-        return train, test
-
-    def evaluate(
```
cwognum, line 100: **Fix:** `evaluate()` is now no longer defined for the Benchmark V2, I think.

polaris/predictions/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+    contributors: list[HubUser] = Field(default_factory=list)
+    model: Model | None = Field(None, exclude=True)
+    benchmark: BenchmarkV2Specification = Field(
+        description="The benchmark specification that these predictions were generated for."
+    )
+    split_label: str = Field(
+        description="The label of the split specified in the benchmark that these predictions were generated with."
+    )
+    predictions: DataFrame = Field(description="A DataFrame with the predictions for each target column.")
```
cwognum, line 30: **Question**: Is a `DataFrame` the best option here? I have two concerns: 

1. It needs to fully fit in memory, which may impose a significant memory bottleneck. 
2. It doesn't naturally support more complex data structures (e.g. an RDKit `Mol` object)

Zarr could be the format we suggest, but doesn't feel right either...

polaris/predictions/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+from pydantic import ConfigDict, Field, PrivateAttr
+from pydantic.alias_generators import to_camel
+
+from polaris._artifact import BaseArtifactModel
+from polaris.benchmark import BenchmarkV2Specification
+from polaris.model import Model
+from polaris.utils.types import HubUser
+
+
+class BenchmarkV2Prediction(BaseArtifactModel):
```
cwognum, line 13: **Question**: This is not used anywhere yet, right?

cwognum (commented):
Thanks @jstlaurent ! 

I left some comments as I was going through the changes, but I assume some of this is still work in progress.

Let me know if you have any specific questions that I missed. It wasn't entirely clear to me what to focus my feedback on.

polaris/benchmark/_base.py
```diff
@@ -213,23 +152,11 @@ def _repr_dict_(self) -> dict:
...
-        return self.__repr__()
-
 
 @deprecated(
     "Use BenchmarkV2Specification instead. If you're loading this dataset from the Polaris Hub, you can ignore this warning."
 )
-class BenchmarkV1Specification(
-    SplitSpecificationV1Mixin, ChecksumMixin, BenchmarkSpecification[BenchmarkResultsV1]
-):
+class BenchmarkV1Specification(SplitSpecificationV1Mixin, ChecksumMixin):
```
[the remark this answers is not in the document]
jstlaurent: Ah, that's a mistake when I removed the generics.

polaris/benchmark/_benchmark_v2.py
```diff
@@ -1,25 +1,26 @@
...
 
-class BenchmarkV2Specification(SplitSpecificationV2Mixin, BenchmarkSpecification[BenchmarkResultsV2]):
+class BenchmarkV2Specification(BenchmarkSpecification):
     _version: ClassVar[Literal[2]] = 2
 
     dataset: DatasetV2 = Field(exclude=True)
     n_classes: dict[ColumnName, int] = Field(default_factory=dict)
 
+    splits: list[SplitV2] = Field(
+        default_factory=list, description="The predefined splits for this benchmark."
```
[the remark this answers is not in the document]
jstlaurent: That's good to know. But this is more for clarity around the new API.

polaris/benchmark/_benchmark_v2.py
jstlaurent: No, because my proposal is that we move to uploading predictions directly. Trying to support results from the client is problematic.

polaris/benchmark/_split_v2.py
```diff
@@ -57,10 +57,12 @@ def deserialize(index_set: bytes) -> "IndexSet":
 
 
 class SplitV2(BaseModel):
-    training: IndexSet
-    test: IndexSet
+    label: str = Field(description="A label for this split")
+    method: str = Field(description="The method used to create this split")
```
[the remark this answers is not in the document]
jstlaurent: I thought about it, but then we constrain ourselves to one level of grouping. The discussion we had around support for repeated CV and multiple splits leads me to believe that will probably be a problem at some point, because the structure around defining all the permutations is far too fuzzy.

polaris/predictions/__init__.py
jstlaurent: Not yet, no. Bu this is what would be uploaded to the Hub.

```python
benchmark = po.load_benchmark("polaris/test-benchmark")
for train, test in benchmark.get_train_test_splits():
    my_model = Model(...)
    my_model.train(train)
    inference = my_model.infer(test)
    prediction = BenchmarkV2Prediction(benchmark=benchmark, model=model, predictions=inference, ...)
    prediction.upload_to_hub()
```

polaris/predictions/__init__.py
jstlaurent: Ideally, I'd like a format that would capture the values mapped to the target columns, so we don't need to implicitly map things together.

polaris/predictions/__init__.py
cwognum: **Question**: Should we stay closer to how we've done this for competitions and have a method on benchmark to submit the predictions? E.g. 

```py
benchmark.submit_predictions(prediction)
```

polaris/predictions/__init__.py
cwognum: The target column can be any dataset column and thus has the same flexibility as any Zarr array. 

We've already seen this play out as we've started to explore structural biology: Submissions for Runs 'N Poses would be the full Protein Structure (i.e. [`biotite.structure.AtomArray`](https://www.biotite-python.org/latest/apidoc/biotite.structure.AtomArray.html#biotite.structure.AtomArray)) and submissions for the Ligand Pose challenge is the 3D conformation of a molecule (i.e. [`rdkit.Chem.Mol`](https://github.com/rdkit/rdkit)).

Furthermore, these predictions can get large. Not as large as datasets, but large nonetheless. To take the example of structural biology again, you may have 1000s of structures you want to evaluate, each of which is a few megabytes in size. You only need one such structure in memory at a time, so the chunking of Zarr can come in handy here. 

I think using Zarr for predictions from a purely technical perspective makes sense, but I worry about the user experience.

polaris/predictions/__init__.py
jstlaurent: That's a good question. I initially started down that road, but walked it back to have an Prediction artifact. Not sure what's the best way for users.

jstlaurent:
Going to close this. It served as a nice starting point for further discussions.

PR #277, opened 2025-03-31
feat: Model Competitions - Model File Upload

roselynh100:
## Changelogs

- Updated `model` class to accept a `file_path` (optional). If `file_path` specified, also calculates new `size` and `md5sum` properties
  - `_validate_file` checks that 1) the file path exists, 2) the file is a `.onnx` file
  - File gets uploaded to r2 using `streaming_set_file()`, which calls the s3 method [`upload_fileobj`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/upload_fileobj.html)
- Created `InvalidModelError`, used when an invalid `file_path` is passed in to a Model

---

_Checklist:_

---

### DIFF
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -970,7 +970,7 @@ class PolarisHubClient(OAuth2Client):
             # If model file is specified, upload it to the Hub
             if model.file_path:
                 with StorageSession(self, "write", model.urn) as storage:
-                    with track_progress(description="Copying model file", total=1):
+                    with track_progress(description="Uploading model file", total=1):
                         with open(model.file_path, "rb") as file:
                             storage.streaming_set_file("root", file)
 
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -971,8 +971,8 @@ class PolarisHubClient(OAuth2Client):
             if model.file_path:
                 with StorageSession(self, "write", model.urn) as storage:
                     with track_progress(description="Copying model file", total=1):
-                        pass
-                        # storage.set_file("root", in_memory_onnx.getvalue())
+                        with open(model.file_path, "rb") as file:
+                            storage.streaming_set_file("root", file)
 
             # Inform the user about where to find their newly created artifact.
             model_url = urljoin(self.settings.hub_url, response.headers.get("Content-Location"))
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -971,7 +971,7 @@ class PolarisHubClient(OAuth2Client):
             if model.file_path:
                 with StorageSession(self, "write", model.urn) as storage:
                     with track_progress(description="Uploading model file", total=1):
-                        storage.streaming_set_file("root", model.file_path)
+                        storage.streaming_set_file("model", model.file_path)
 
             # Inform the user about where to find their newly created artifact.
             model_url = urljoin(self.settings.hub_url, response.headers.get("Content-Location"))
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -953,9 +953,10 @@ class PolarisHubClient(OAuth2Client):
                 in_memory_onnx = BytesIO()
 
                 onnx_model = onnx.load(model.file_path)
+                onnx.checker.check_model(onnx_model)
                 onnx.save_model(onnx_model, in_memory_onnx)
-                onnx_size = len(in_memory_onnx.getbuffer())
 
+                onnx_size = len(in_memory_onnx.getbuffer())
                 onnx_md5 = md5(in_memory_onnx.getbuffer()).hexdigest()
 
                 file_content = {
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -971,8 +971,7 @@ class PolarisHubClient(OAuth2Client):
             if model.file_path:
                 with StorageSession(self, "write", model.urn) as storage:
                     with track_progress(description="Uploading model file", total=1):
-                        with open(model.file_path, "rb") as file:
-                            storage.streaming_set_file("root", file)
+                        storage.streaming_set_file("root", model.file_path)
 
             # Inform the user about where to find their newly created artifact.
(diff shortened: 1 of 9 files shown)


polaris/hub/client.py
```diff
@@ -945,9 +946,41 @@ def upload_model(
...
+
             # Make a request to the Hub
             url = f"/v2/model/{model.artifact_id}"
-            response = self._base_request_to_hub(url=url, method="PUT", json={"access": access, **model_json})
+            response = self._base_request_to_hub(
+                url=url,
+                method="PUT",
+                json={
+                    "access": access,
+                    **({"fileContent": file_content} if file_content else {}),
```
Andrewq11, line 974: Would this breakdown if you just did `"fileContent": file_content`? I see that `file_content` is initialized to `None` whether there is a model path or not.

polaris/hub/client.py
```diff
@@ -945,9 +946,41 @@ def upload_model(
...
+                    **model_json,
+                },
+            )
+
+            # If model file is specified, upload it to the Hub
+            if model.file_path:
+                with StorageSession(self, "write", model.urn) as storage:
+                    with track_progress(description="Copying model file", total=1):
+                        storage.set_file("root", in_memory_onnx.getvalue())
```
Andrewq11, line 983: Let's chat about the need for loading the file in-memory here.

polaris/model/__init__.py
```diff
@@ -44,6 +50,22 @@ class Model(BaseArtifactModel):
     code_url: HttpUrlString | None = None
     report_url: HttpUrlString | None = None
 
+    file_path: str | None = None
+
+    @field_validator("file_path")
+    @classmethod
+    def _validate_file(cls, v: str | None) -> str | None:
+        if isinstance(v, str):
+            if not dmfs.is_file(v) or dmfs.get_extension(v) not in _SUPPORTED_MODEL_EXTENSIONS:
```
Andrewq11, line 59: Could we use Python's [`pathlib`](https://docs.python.org/3/library/pathlib.html) for this instead of `datamol`?

polaris/model/__init__.py
```diff
@@ -44,6 +50,22 @@ class Model(BaseArtifactModel):
...
+                raise InvalidModelError(f"{v} is not a valid .onnx file.")
+
+            try:
+                onnx_model = onnx.load(v)
+                onnx.checker.check_model(onnx_model)
+            except Exception:
+                raise InvalidModelError(f"{v} is not a valid .onnx file.")
```
Andrewq11, line 66: Let's remove this step (again, on me). We'll run the check server-side to avoid this workflow in the client potentially needing a lot of memory.

pyproject.toml
```diff
@@ -42,6 +42,7 @@ dependencies = [
     "httpx",
     "numcodecs[msgpack]>=0.13.1",
     "numpy <3",
+    "onnx",
```
Andrewq11, line 45: Given my above comments, can we remove this as a dep here and in the `env.yml`? Let's chat about it.

Andrewq11 (commented):
Thanks, @roselynh100! Very nice work. I suggested some things before that I'd like to walk back on. Apologies for this. Will explain more shortly.

polaris/hub/client.py
```diff
@@ -945,9 +945,34 @@ def upload_model(
...
+                    "access": access,
+                    "fileContent": file_content,
+                    **model_json,
+                },
+            )
+
+            # If model file is specified, upload it to the Hub
+            if model.file_path:
+                with StorageSession(self, "write", model.urn) as storage:
+                    with track_progress(description="Copying model file", total=1):
```
Andrewq11, line 973:
```suggestion
                    with track_progress(description="Uploading model file", total=1):
```

polaris/hub/client.py
```diff
@@ -945,9 +945,34 @@ def upload_model(
...
+                        with open(model.file_path, "rb") as file:
```
Andrewq11, line 974: Could you move this `open` call inside the `storage.streaming_set_file` method? Similar to this [example](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/upload_fileobj.html)!

polaris/hub/storage.py
```diff
@@ -583,6 +584,38 @@ def set_file(self, path: str, value: bytes | bytearray):
...
+        Set a file at the given path, without loading the file into memory.
+        """
+        if path not in self.paths.files:
+            raise NotImplementedError(f"{type(self.paths).__name__} only supports files {self.paths.files}.")
+
+        relative_path = self._relative_path(getattr(self.paths, path))
+        bucket_name = relative_path.parts[0]
+        key = '/'.join(relative_path.parts[1:])
+
+        match relative_path.suffix:
```
Andrewq11, line 598: Since we know that we'll only be using this for `.onnx` files (for now), can we remove the `.parquet` and `.roaring` cases? We can keep the default case just so we have this `match case` scaffold for later.

polaris/hub/client.py
```diff
@@ -945,9 +945,34 @@ def upload_model(
...
+                            storage.streaming_set_file("root", file)
```
Andrewq11, line 975: Is `"root"` needed here?

polaris/model/__init__.py
```diff
@@ -44,6 +50,31 @@ class Model(BaseArtifactModel):
...
+    @classmethod
+    def _validate_file(cls, v: str | None) -> str | None:
+        if isinstance(v, str):
+            if not Path(v).exists() or Path(v).suffix not in _SUPPORTED_MODEL_EXTENSIONS:
+                raise InvalidModelError(f"{v} is not a valid .onnx file.")
+        return v
+
+    @computed_field
+    @property
+    def size(self) -> int | None:
```
Andrewq11, line 65: Nice approach with these computed fields!

Andrewq11 (approved):
Nice work, @roselynh100! Looks good to me once we address a couple things.

PR #291, opened 2025-05-20
Refactor/public bucket access

j279li:
Refactors the Polaris Hub client and storage modules to simplify authentication handling and allow for public access of artifacts. Key changes include removing authentication for reads, and updating dataset and benchmark retrieval methods to get data path from the API response. 

### Authentication and Session Management Updates:
* Removedauthentication checks for retrieval methods in `PolarisHubClient` and added explicit token validation where necessary, such as in `upload_results` and `upload_model`.
* Introduced a `mode` parameter in `StorageSession` to control authentication requirements, making it optional for read operations but mandatory for write operations. 

### Dataset and Benchmark Handling Improvements:
* Simplified dataset and benchmark retrieval methods by replacing `_base_request_to_hub` with `request` and using metadata fields (`zarr_path`, `table_path`) for loading resources.

### Storage Refactor:
* Replaced custom S3 storage logic for public reads in `StorageSession` with `fsspec` for handling get file and store operations.

### Codebase Simplification:
* Consolidated logic for handling Zarr archives and storage paths across dataset-related methods.

### DIFF
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -342,11 +342,6 @@ class PolarisHubClient(OAuth2Client):
         root_url = response_data.get("root")
         extension_url = response_data.get("extension")
 
-        if root_url is not None:
-            root_url = root_url.strip('"')
-        if extension_url is not None:
-            extension_url = extension_url.strip('"')
-
         # Load the dataset table and optional Zarr archive
         with fsspec.open(root_url, mode="rb") as f:
             table = pd.read_parquet(f)
@@ -370,8 +365,6 @@ class PolarisHubClient(OAuth2Client):
 
         root_url = response_data.get("root")
 
-        if root_url is not None:
-            root_url = root_url.strip('"')
         # For v2 datasets, the zarr_path always exists
         dataset = DatasetV2(zarr_root_path=root_url, **response_data)
         return dataset
@@ -620,9 +613,6 @@ class PolarisHubClient(OAuth2Client):
         response_data = response.json()
 
         root_url = response_data.get("root")
-
-        if root_url is not None:
-            root_url = root_url.strip('"')
 
         return CompetitionSpecification(zarr_root_path=root_url, **response_data)
 
--- a/env.yml
+++ b/env.yml
@@ -17,6 +17,8 @@ dependencies:
   # Hub client
   - authlib
   - httpx
+  - requests
+  - aiohttp
 
   # Scientific
   - numpy < 3


j279li:
Tests right now are failing because the Hub hasn't been updated yet (as the client isn't sending authentication tokens anymore).

polaris/hub/storage.py
```diff
@@ -581,42 +587,21 @@ def set_file(self, path: str, value: bytes | bytearray):
...
-            path=relative_path.parent,
-            access_key=storage_data.key,
-            secret_key=storage_data.secret,
-            token=f"jwt/{self.token.access_token}",
-            endpoint_url=storage_data.endpoint,
-        )
-        return store[relative_path.name]
+        # The path is now a full URL, so we use it directly
+        with fsspec.open(path, mode="rb") as f:
+            return f.read()
```
Andrewq11, line 600: Could use your Zarr wisdom here @jstlaurent:

Seeing as we no longer need a storage session / credentials for retrieval, do we still have a use case for this `get_file` method in this class?

If not, what do you think about making this class specific to writes and then another class specific to retrieval which contains a simple `get_file` and `get_store` method?

polaris/dataset/_dataset.py
```diff
@@ -138,12 +137,11 @@ def load_zarr_root_from_hub(self):
...
         Loads a Zarr archive from the Hub.
         """
-        from polaris.hub.client import PolarisHubClient
         from polaris.hub.storage import StorageSession
+        import zarr
 
-        with PolarisHubClient() as client:
-            with StorageSession(client, "read", self.urn) as storage:
-                return zarr.open_consolidated(store=storage.store("extension"))
+        store = StorageSession.store(self)
```
Andrewq11, line 143: Related to my other comment about the `store` method, I would change this to:

```suggestion
        store = fsspec.get_mapper(self.zarr_root_path)
```

polaris/dataset/_dataset_v2.py
```diff
@@ -117,12 +117,11 @@ def load_zarr_root_from_hub(self):
...
         Loads a Zarr archive from the Hub.
         """
-        from polaris.hub.client import PolarisHubClient
         from polaris.hub.storage import StorageSession
+        import zarr
 
-        with PolarisHubClient() as client:
-            with StorageSession(client, "read", self.urn) as storage:
-                return zarr.open_consolidated(store=storage.store("root"))
+        store = StorageSession.store(self)
```
Andrewq11, line 123: Similar comment here!

polaris/hub/client.py
```diff
@@ -184,10 +179,10 @@ def fetch_token(self, **kwargs):
                 message=f"Could not obtain a token to access the Hub. Error was: {error.error} - {error.description}"
             ) from error
 
-    def _base_request_to_hub(self, url: str, method: str, **kwargs) -> Response:
-        """Utility function since most API methods follow the same pattern"""
+    def _base_request_to_hub(self, url: str, method: str, withhold_token: bool = False, **kwargs) -> Response:
```
Andrewq11, line 182: To be extra explicit about reading or writing, I would remove the default value of `False` from the `withhold_token` argument here.

polaris/hub/client.py
```diff
@@ -335,23 +333,19 @@ def _get_v1_dataset(
             A `Dataset` instance, if it exists.
         """
         url = f"/v1/dataset/{owner}/{slug}"
-        response = self._base_request_to_hub(url=url, method="GET")
+        response = self._base_request_to_hub(url=url, method="GET", withhold_token=True)
         response_data = response.json()
-
-        # Disregard the Zarr root in the response. We'll get it from the storage token instead.
         response_data.pop("zarrRootPath", None)
+        response_data.pop("zarr_root_path", None)
```
Andrewq11, line 339: Apologies if I missed this, but why is this needed now? I was under the assumption the data model returned from the Hub used the camel case `zarrRootPath` only.

polaris/hub/client.py
```diff
@@ -364,17 +358,15 @@ def _get_v1_dataset(
     def _get_v2_dataset(self, owner: str | HubOwner, slug: str) -> DatasetV2:
         """"""
         url = f"/v2/dataset/{owner}/{slug}"
-        response = self._base_request_to_hub(url=url, method="GET")
+        response = self._base_request_to_hub(url=url, method="GET", withhold_token=True)
         response_data = response.json()
-
-        # Disregard the Zarr root in the response. We'll get it from the storage token instead.
         response_data.pop("zarrRootPath", None)
+        response_data.pop("zarr_root_path", None)
```
Andrewq11, line 364: Same question here!

polaris/hub/client.py
```diff
@@ -364,17 +358,15 @@ def _get_v1_dataset(
...
 
-        # Load the Zarr archive
-        with StorageSession(self, "read", DatasetV2.urn_for(owner, slug)) as storage:
-            zarr_root_path = str(storage.paths.root)
-
-        dataset = DatasetV2(zarr_root_path=zarr_root_path, **response_data)
+        root_url = response_data.get("root")
+        print(f"Root URL: {root_url}")
```
Andrewq11, line 367:
```suggestion

```

polaris/hub/client.py
```diff
@@ -693,6 +691,8 @@ def upload_model(
             owner: Which Hub user or organization owns the artifact. Takes precedence over `model.owner`.
             parent_artifact_id: The `owner/slug` of the parent model, if uploading a new version of a model.
         """
+        if not self.ensure_active_token():
+            raise PolarisUnauthorizedError()
```
Andrewq11, line 695: Given that this is something which we'll probaby want to do on all uploads, I would suggest moving this check to the `_base_request_to_hub` method. We can run the check anytime `withhold_token` is set to `False`.

polaris/dataset/_base.py
```diff
@@ -171,15 +171,16 @@ def zarr_root(self) -> zarr.Group | None:
...
 
         if self._zarr_root is not None:
             return self._zarr_root
 
         if self.zarr_root_path is None:
             return None
 
-        saved_on_hub = self.zarr_root_path.startswith(StorageSession.polaris_protocol)
+        parsed = urlparse(self.zarr_root_path)
+        saved_on_hub = parsed.scheme == "https"
```
Andrewq11, line 183: We could get more specific here and ensure the root of the URL matches the path expected for our buckets. You could define a new settings variable in `polaris/hub/settings.py` that defines the expected root (e.g. `https://data.polarishub.io`).

Then, you could check here that the `self.zarr_root_path` starts with tha value defined at `settings.bucket_url` (or whatever we name it).

Andrewq11 (commented):
Thanks, @j279li! A lot of good work here. I've left some comments related to suggestions and clarifications that we should cover. 

Also, I've requested the wisdom of @jstlaurent on some storage action organization.

polaris/hub/client.py
j279li: Oh yes, you're right about that. Let me remove it.

polaris/hub/client.py
j279li: I'll fix it here too

polaris/dataset/_base.py
j279li: We could do that, but I'd also have to put the URL of the test bucket in that case. Do we want that in the client?

polaris/hub/client.py
j279li: Good idea

polaris/hub/storage.py
```diff
@@ -581,42 +587,21 @@ def set_file(self, path: str, value: bytes | bytearray):
...
-            secret_key=storage_data.secret,
-            token=f"jwt/{self.token.access_token}",
-            endpoint_url=storage_data.endpoint,
-        )
+    def store(self):
+        """
+        Return a fsspec mapper for the given path, using the provided path as a full URL.
+        """
+        path = self.zarr_root_path
+        return fsspec.get_mapper(path)
```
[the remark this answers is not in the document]
j279li: It does work for me, but upon inspection, I think it's because i'm calling it as StorageSession.store(self), which passes the dataset instance as the self argument to store. This is pretty sketchy, so I'll go with your suggestion.

polaris/dataset/_base.py
Andrewq11: We wouldn't! We could set the default value of the settings variable to the prod bucket URL, but for development, we could override it using the `.env` file. 

For example, if you named the settings variable `bucket_url`, we could add `POLARIS_BUCKET_URL` to the local `.env` file with a value of the dev url. That will override the default value set in the `settings.py` file. 

Here are the [Pydantic docs](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#dotenv-env-support) on it for further reading in case you are interested.

polaris/hub/storage.py
```diff
@@ -581,42 +587,21 @@ def set_file(self, path: str, value: bytes | bytearray):
             endpoint_url=storage_data.endpoint,
             content_type=content_type,
         )
-        store[relative_path.name] = value
+        # Use StorageSession with mode='write' for write operations
+        with StorageSession(self.hub_client, "write", self.resource, mode="write"):
+            store[relative_path.name] = value
```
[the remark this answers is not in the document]
j279li: I think it should work too yes.

polaris/dataset/_base.py
j279li: I see! That's great then, I'll change that up.

polaris/dataset/_base.py
```diff
@@ -171,15 +172,16 @@ def zarr_root(self) -> zarr.Group | None:
...
             return self._zarr_root
 
         if self.zarr_root_path is None:
             return None
 
-        saved_on_hub = self.zarr_root_path.startswith(StorageSession.polaris_protocol)
+        settings = PolarisHubSettings()
+        saved_on_hub = self.zarr_root_path and self.zarr_root_path.startswith(settings.bucket_url)
 
         if self._warn_about_remote_zarr and saved_on_hub:
```
jstlaurent, line 186: **Suggestion**
We'll probably end up with some circular import error here, because we import from a `polaris.hub`module, and these modules in turn import the various artifact classes. Ideally, the importing should only happen in one direction, which would be `polaris.hub -> everything else`.

The goal here is to determine if the path is local, to warn the user about remote files and their potentially slow access. From the `urlparse` import above, I'm guessing you tried that approach first, and it's probably a good one.

Another way is to leverage `fsspec`, which we have as a dependency. Using its [url_to_fs](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.core.url_to_fs) utility, we can do:

```suggestion
        fs, _ = fsspec.url_to_fs(self.zarr_root_path)
        remote = 'local' in fs.protocol
        
        if self._warn_about_remote_zarr and remote:
```

polaris/dataset/_dataset.py
```diff
@@ -138,12 +137,11 @@ def load_zarr_root_from_hub(self):
         """
         Loads a Zarr archive from the Hub.
         """
-        from polaris.hub.client import PolarisHubClient
         from polaris.hub.storage import StorageSession
+        import zarr
```
jstlaurent, line 141: **Suggestion**

Keep this at the top of the module. We moved `from polaris.hub.storage import StorageSession` because of circular import issues. Also, no need for the `StorageSession` here anymore.

```suggestion
```

polaris/hub/storage.py
jstlaurent: No, I don't think we do. In fact, I don't think we need anything to mediate reading files. Users know their way around `fsspec` for reading files. Same thing for Zarr, which will happily take a `fsspec` mapper as input.

I was also thinking that this would become a "write" manager, since that is the only time we'll need to manage credentials.

polaris/hub/storage.py
```diff
@@ -476,15 +477,19 @@ class StorageSession(OAuth2Client):
     """
     A context manager for managing a storage session, with token exchange and token refresh capabilities.
     Each session is associated with a specific scope and resource.
+    The `mode` parameter controls whether authentication is required:
+        - For 'read' mode, authentication is optional.
+        - For 'write' mode, authentication is always required.
```
jstlaurent, line 482: **Comment**
This a nice idea, but I don't think we need this class to mediate read operations. When we implement predictions/models upload, it will be relevant for write operations to that bucket. But for reading public resources, users can use `fsspec` as needed.

polaris/hub/client.py
```diff
@@ -335,23 +336,18 @@ def _get_v1_dataset(
...
         # Load the dataset table and optional Zarr archive
         with StorageSession(self, "read", Dataset.urn_for(owner, slug)) as storage:
-            table = pd.read_parquet(BytesIO(storage.get_file("root")))
-            zarr_root_path = storage.paths.extension
-
-            if zarr_root_path is not None:
-                # For V1 datasets, the Zarr Root is optional.
-                # It should be None if the dataset does not use pointer columns
-                zarr_root_path = str(zarr_root_path)
+            table = pd.read_parquet(BytesIO(storage.get_file(root_url)))
```
jstlaurent, line 348: **Suggestion**
I don't think we need the `StorageSesion` here anymore.

```suggestion
        table = pd.read_parquet(fsspec.open(root_url))
```

polaris/hub/client.py
```diff
@@ -184,10 +179,13 @@ def fetch_token(self, **kwargs):
                 message=f"Could not obtain a token to access the Hub. Error was: {error.error} - {error.description}"
             ) from error
 
-    def _base_request_to_hub(self, url: str, method: str, **kwargs) -> Response:
-        """Utility function since most API methods follow the same pattern"""
+    def _base_request_to_hub(self, url: str, method: str, withhold_token: bool, **kwargs) -> Response:
```
jstlaurent, line 182: **Question**
Do we need this new parameter? What happens if we don't check for this?

polaris/dataset/_base.py
jstlaurent: I actually have a comment in my review about winding this back to a simpler check. We want to warn the user about the potential size of remote datasets, so it's not so much about checking if it's on the Hub, as if it is remote. So using something like `fsspec` to check the protocol is enough.

Also, if we encode the bucket URL in the setting of the client, we're stuck if we want to change it in the future. The goal of having the Hub provide the full path to the client is that it can remain ignorant of these settings, letting us change them whenever we want.

polaris/hub/client.py
j279li: This should essentially not send the authentication token when making the request. I guess technically it's not entirely needed, since right now the get hub endpoints will just ignore the token (ie not authenticate) if one is present. I included because it seemed like good practice to not send it if it wasn't needed.

polaris/hub/storage.py
```diff
@@ -474,17 +475,18 @@ def set_token(self, token: dict[str, Any] | HubStorageOAuth2Token) -> None:
...
+    A context manager for managing a storage session for upload/write operations, with token exchange and token refresh capabilities.
     Each session is associated with a specific scope and resource.
     """
 
     polaris_protocol = "polarisfs"
 
     token_auth_class = StorageTokenAuth
 
-    def __init__(self, hub_client, scope: Scope, resource: ArtifactUrn):
+    def __init__(self, hub_client, scope: Scope, resource: ArtifactUrn, mode: str = "read"):
```
Andrewq11, line 486: Since we want to have this class become a write manager, we can remove the `mode` argument and its references.

polaris/hub/storage.py
```diff
@@ -581,42 +584,5 @@ def set_file(self, path: str, value: bytes | bytearray):
             endpoint_url=storage_data.endpoint,
             content_type=content_type,
         )
+        # Use StorageSession with mode='write' for write operations
```
Andrewq11, line 587:
```suggestion

```

polaris/hub/storage.py
```diff
@@ -18,6 +18,7 @@
 from zarr.context import Context
 from zarr.storage import Store
 from zarr.util import buffer_size
+import fsspec
```
Andrewq11, line 21: We can remove this since it is no longer used.
```suggestion
```

polaris/hub/client.py
Andrewq11, line 1: There's a few unused imports in this file which we can clean up!

polaris/hub/client.py
jstlaurent: Ok, let's leave it like that for now. Thanks!

PR #292, opened 2025-06-03
Feat/add prediction artifact and upload method

j279li:
## Summary of Changes

This PR introduces a new Prediction artifact, and an associated upload method 

### New Features
- **Prediction Artifact (`Predictions`)**
  - Added a new `Predictions` model to represent prediction artifacts as Zarr archives.
  - Includes validators to ensure Zarr structure consistency.
  - Provides convenient properties (`columns`, `dtypes`, `n_rows`, etc.) for quick data exploration.
  - Methods for setting/getting prediction values, creating Zarr archives from dicts, and uploading to the Polaris Hub.

- **Upload API (`upload_prediction`)**
  - Added a new `upload_prediction` method in `PolarisHubClient` for uploading prediction artifacts.
  - Handles metadata, manifest, and archive uploads with progress tracking.

---

### DIFF
--- a/polaris/evaluate/__init__.py
+++ b/polaris/evaluate/__init__.py
@@ -29,4 +29,5 @@ __all__ = [
     "evaluate_benchmark",
     "CompetitionPredictions",
     "BenchmarkPredictions",
+    "Predictions",
 ]
--- a/polaris/evaluate/_predictions.py
+++ b/polaris/evaluate/_predictions.py
@@ -1,29 +1,25 @@
-from collections import defaultdict
+import logging
+import re
+from typing import Any
 
 import numpy as np
-from pydantic import (
-    BaseModel,
-    ConfigDict,
-    Field,
-    TypeAdapter,
-    field_serializer,
-    field_validator,
-    model_validator,
-)
-from typing_extensions import Self
+import zarr
+from pydantic import model_validator
+from pydantic import TypeAdapter
 
-from polaris.evaluate import ResultsMetadataV1
-from polaris.utils.misc import convert_lists_to_arrays
-from polaris.utils.types import (
-    HttpUrlString,
-    HubOwner,
-    IncomingPredictionsType,
-    PredictionsType,
-    SlugCompatibleStringType,
-)
+from polaris._artifact import BaseArtifactModel
+from polaris.model import Model
+from polaris.utils.types import HubOwner
+from polaris.dataset.zarr._manifest import generate_zarr_manifest
+from polaris.utils.misc import calculate_file_md5
+from polaris.hub.client import PolarisHubClient
+
+logger = logging.getLogger(__name__)
+
+_INDEX_ARRAY_KEY = "__index__"
 
 
-class BenchmarkPredictions(BaseModel):
+class BenchmarkPredictions:
     """
     Base model to represent predictions in the Polaris code base.
 
@@ -37,15 +33,18 @@ class BenchmarkPredictions(BaseModel):
         test_set_sizes: The number of rows in each test set for the associated benchmark.
     """
 
-    predictions: PredictionsType
+    predictions: dict
     target_labels: list[str]
     test_set_labels: list[str]
     test_set_sizes: dict[str, int]
 
-    model_config = ConfigDict(arbitrary_types_allowed=True)
+    def __init__(self, predictions, target_labels, test_set_labels, test_set_sizes):
+        self.predictions = predictions
+        self.target_labels = target_labels
+        self.test_set_labels = test_set_labels
+        self.test_set_sizes = test_set_sizes
 
-    @field_serializer("predictions")
-    def _serialize_predictions(self, predictions: PredictionsType):
+    def _serialize_predictions(self, predictions):
         """
         Recursively converts all numpy values in the predictions dictionary to lists
         so they can be serialized.
@@ -59,16 +58,12 @@ class BenchmarkPredictions(BaseModel):
 
         return convert_to_list(predictions)
 
-    @field_validator("target_labels", "test_set_labels")
-    @classmethod
-    def _validate_labels(cls, v: list[str]) -> list[str]:
+    def _validate_labels(self, v: list[str]) -> list[str]:
         if len(set(v)) != len(v):
             raise ValueError("The predictions contain duplicate columns")
         return v
 
-    @model_validator(mode="before")
-    @classmethod
-    def _validate_predictions(cls, data: dict) -> dict:
+    def _validate_predictions(self, data: dict) -> dict:
         """Normalizes the predictions format to a standard representation we use internally"""
 
         # This model validator runs before any Pydantic internal validation.
@@ -76,20 +71,17 @@ class BenchmarkPredictions(BaseModel):
         # However, this implies that the fields can theoretically be any type.
 
         # Ensure the type of the incoming predictions is correct
-        validator = TypeAdapter(IncomingPredictionsType, config={"arbitrary_types_allowed": True})
-        predictions = validator.validate_python(data.get("predictions"))
+        predictions = data.get("predictions")
 
         # Ensure the type of the target_labels and test_set_labels is correct
-        validator = TypeAdapter(list[str])
-        target_labels = validator.validate_python(data.get("target_labels"))
-        test_set_labels = validator.validate_python(data.get("test_set_labels"))
+        target_labels = data.get("target_labels")
+        test_set_labels = data.get("test_set_labels")
 
         validator = TypeAdapter(dict[str, int])
         test_set_sizes = validator.validate_python(data.get("test_set_sizes"))
 
         # Normalize the predictions to a standard representation
-        predictions = convert_lists_to_arrays(predictions)
-        predictions = cls._normalize_predictions(predictions, target_labels, test_set_labels)
+        predictions = self._normalize_predictions(predictions, target_labels, test_set_labels)
 
         # Update class data with the normalized fields. Use of the `update()` method
         # is required to prevent overwriting class data when this class is inherited.
@@ -104,8 +96,7 @@ class BenchmarkPredictions(BaseModel):
 
         return data
 
-    @model_validator(mode="after")
-    def check_test_set_size(self) -> Self:
+    def check_test_set_size(self):
         """Verify that the size of all predictions"""
         for test_set_label, test_set in self.predictions.items():
             for target in test_set.values():
@@ -119,17 +110,14 @@ class BenchmarkPredictions(BaseModel):
                     )
         return self
 
-    @classmethod
-    def _normalize_predictions(
-        cls, predictions: IncomingPredictionsType, target_labels: list[str], test_set_labels: list[str]
-    ) -> PredictionsType:
+    def _normalize_predictions(self, predictions, target_labels, test_set_labels):
         """
         Normalizes  the predictions to a standard representation we use internally.
         This standard representation is a nested, two-level dictionary:
         `{test_set_name: {target_column: np.ndarray}}`
         """
         # (1) If the predictions are already fully specified, no need to do anything
-        if cls._is_fully_specified(predictions, target_labels, test_set_labels):
+        if self._is_fully_specified(predictions, target_labels, test_set_labels):
             return predictions
 
         # If not fully specified, we distinguish 4 cases based on the type of benchmark.
@@ -172,10 +160,7 @@ class BenchmarkPredictions(BaseModel):
 
         return predictions
 
-    @classmethod
-    def _is_fully_specified(
-        cls, predictions: IncomingPredictionsType, target_labels: list[str], test_set_labels: list[str]
-    ) -> bool:
+    def _is_fully_specified(self, predictions, target_labels, test_set_labels):
         """
         Check if the predictions are fully specified for the target columns and test set names.
         """
@@ -196,9 +181,7 @@ class BenchmarkPredictions(BaseModel):
 
         return True
 
-    def get_subset(
-        self, test_set_subset: list[str] | None = None, target_subset: list[str] | None = None
-    ) -> "BenchmarkPredictions":
+    def get_subset(self, test_set_subset: list[str] | None = None, target_subset: list[str] | None = None):
         """Return a subset of the original predictions"""
 
         if test_set_subset is None and target_subset is None:
@@ -209,7 +192,7 @@ class BenchmarkPredictions(BaseModel):
         if target_subset is None:
             target_subset = self.target_labels
 
-        predictions = defaultdict(dict)
+        predictions = {}
         for test_set_name in self.predictions.keys():
             if test_set_name not in test_set_subset:
                 continue
@@ -218,7 +201,7 @@ class BenchmarkPredictions(BaseModel):
                 if target_name not in target_subset:
                     continue
 
-                predictions[test_set_name][target_name] = target
+                predictions[test_set_name] = {target_name: target}
 
         test_set_sizes = {k: v for k, v in self.test_set_sizes.items() if k in predictions.keys()}
         return BenchmarkPredictions(
@@ -228,15 +211,13 @@ class BenchmarkPredictions(BaseModel):
             test_set_sizes=test_set_sizes,
         )
 
-    def get_size(
-        self, test_set_subset: list[str] | None = None, target_subset: list[str] | None = None
-    ) -> int:
+    def get_size(self, test_set_subset: list[str] | None = None, target_subset: list[str] | None = None):
         """Return the total number of predictions, allowing for filtering by test set and target"""
         if test_set_subset is None and target_subset is None:
             return sum(self.test_set_sizes.values()) * len(self.target_labels)
         return len(self.get_subset(test_set_subset, target_subset))
 
-    def flatten(self) -> np.ndarray:
+    def flatten(self):
         """Return the predictions as a single, flat numpy array"""
         if len(self.test_set_labels) != 1 and len(self.target_labels) != 1:
             raise ValueError(
@@ -244,12 +225,12 @@ class BenchmarkPredictions(BaseModel):
             )
         return self.predictions[self.test_set_labels[0]][self.target_labels[0]]
 
-    def __len__(self) -> int:
+    def __len__(self):
         """Return the total number of predictions"""
         return self.get_size()
 
 
-class CompetitionPredictions(BenchmarkPredictions, ResultsMetadataV1):
+class CompetitionPredictions(BenchmarkPredictions):
     """
     Predictions for competition benchmarks.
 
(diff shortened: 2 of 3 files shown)


polaris/evaluate/__init__.py
```diff
@@ -29,4 +29,5 @@
     "evaluate_benchmark",
     "CompetitionPredictions",
     "BenchmarkPredictions",
+    "Predictions",
```
jstlaurent, line 32: **Comment**
For this to be re-exported, you need to import it at the top of the file as well.

polaris/hub/client.py
```diff
@@ -714,3 +717,64 @@ def upload_model(
...
+        timeout: TimeoutTypes = (10, 200),
+        owner: HubOwner | str | None = None,
+        if_exists: ZarrConflictResolution = "replace",
+        parent_artifact_id: str | None = None,
+    ):
+        """
+        Upload a Predictions artifact (with Zarr archive) to the Polaris Hub.
+        """
+
+        logger = logging.getLogger(__name__)
```
jstlaurent, line 733: **Comment**
There's already a logger defined at the top of the module.

polaris/hub/client.py
```diff
@@ -714,3 +717,64 @@ def upload_model(
...
+                zmetadata_content = prediction.zarr_root.store.store[".zmetadata"]
+                destination[".zmetadata"] = zmetadata_content
+                # Copy the Zarr archive
+                destination.copy_from_source(
+                    prediction.zarr_root.store.store, if_exists=if_exists, log=logger.info
+                )
+
+        print(
+            f"[green]Your prediction has been successfully uploaded to the Hub. View it here: {prediction_url}"
+        )
```
jstlaurent, line 780: **Comment**
The `[green]` prefix will only work with `progress.log`. It won't be recognized as text formatting by `print`.

polaris/hub/client.py
```diff
@@ -714,3 +717,64 @@ def upload_model(
...
+            with track_progress(description="Copying Zarr archive", total=1) as (progress_zarr, task_zarr):
+                progress_zarr.log("[yellow]This may take a while.")
+                destination = storage.store("root")
+                # Consolidate Zarr metadata and upload .zmetadata
+                zarr.consolidate_metadata(prediction.zarr_root.store.store)
+                zmetadata_content = prediction.zarr_root.store.store[".zmetadata"]
+                destination[".zmetadata"] = zmetadata_content
+                # Copy the Zarr archive
+                destination.copy_from_source(
+                    prediction.zarr_root.store.store, if_exists=if_exists, log=logger.info
```
jstlaurent, line 775: **Suggestion**
Pass in the `log` from the progress tracker, to have the printout be nicer.
```suggestion
                    prediction.zarr_root.store.store, if_exists=if_exists, log=progress_zarr.log
```

polaris/hub/client.py
```diff
@@ -714,3 +717,64 @@ def upload_model(
...
+        url = f"/v2/prediction/{prediction.artifact_id}"
+        with track_progress(description="Uploading prediction metadata", total=1) as (progress, task):
+            response = self._base_request_to_hub(
+                url=url,
+                method="PUT",
+                json={
+                    "zarrManifestFileContent": {
+                        "md5Sum": prediction.zarr_manifest_md5sum,
+                    },
+                    "parentArtifactId": parent_artifact_id,
```
jstlaurent, line 749: **Change requested**
Predictions don't have parent artifacts.
```suggestion
```

polaris/hub/client.py
```diff
@@ -714,3 +717,64 @@ def upload_model(
...
+                    storage.set_file("manifest", manifest_file.read())
+
+            # Step 3: Upload Zarr archive
+            with track_progress(description="Copying Zarr archive", total=1) as (progress_zarr, task_zarr):
+                progress_zarr.log("[yellow]This may take a while.")
+                destination = storage.store("root")
+                # Consolidate Zarr metadata and upload .zmetadata
+                zarr.consolidate_metadata(prediction.zarr_root.store.store)
+                zmetadata_content = prediction.zarr_root.store.store[".zmetadata"]
+                destination[".zmetadata"] = zmetadata_content
```
jstlaurent, line 772: **Comment**
Not sure if we need this for predictions. @cwognum?

polaris/evaluate/_predictions.py
```diff
@@ -275,3 +286,169 @@ def __repr__(self):
...
+    Attributes:
+        benchmark_artifact_id: The artifact ID of the associated benchmark.
+        model: (Optional) The Model artifact used to generate these predictions.
+        zarr_root_path: Path to the Zarr archive containing the predictions.
+        column_types: (Optional) A dictionary mapping column names to expected types.
+    """
+
+    _artifact_type = "prediction"
+
+    benchmark_artifact_id: str
```
jstlaurent, line 304: **Comment**
This should be a `BenchmarkV2Specification` instance. This will allow this class to look up the target columns for the benchmark directly, and make sure that the predictions are valid. Both in terms of having predictions for all columns, but also to validate the type of each column.

polaris/evaluate/_predictions.py
```diff
@@ -275,3 +286,169 @@ def __repr__(self):
...
+
+    @property
+    def rows(self):
+        return range(self.n_rows)
+
+    @property
+    def zarr_manifest_path(self):
+        if self._zarr_manifest_path is None:
+            zarr_manifest_path = generate_zarr_manifest(
+                self.zarr_root_path, getattr(self, "_cache_dir", None)
```
jstlaurent, line 369: **Comment**
`_cache_dir` is never set on the class, so this will always be `None`.

polaris/evaluate/_predictions.py
```diff
@@ -275,3 +286,169 @@ def __repr__(self):
...
+
+    _artifact_type = "prediction"
+
+    benchmark_artifact_id: str
+    model: Model | None = None
+    zarr_root_path: str
+    _zarr_manifest_path: str | None = None
+    _zarr_manifest_md5sum: str | None = None
+    column_types: dict[str, type] = {}
+
```
jstlaurent, line 310: **Comment**
What you need in this class is to instantiate a Zarr file (probably as a Pydantic [Private attribute](https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.PrivateAttr)), in which you'll store the data the user sets.

That Zarr file can be initialized with a [root group](https://zarr.readthedocs.io/en/stable/user-guide/groups.html), and an [array](https://zarr.readthedocs.io/en/stable/user-guide/arrays.html) for each target column. These arrays can use codecs (from `polaris/dataset/zarr/codecs.py`) to handle more complex data types, matching the types defined on the underlying dataset of the benchmark.

polaris/evaluate/_predictions.py
```diff
@@ -275,3 +286,169 @@ def __repr__(self):
...
+    @model_validator(mode="after")
+    def _validate_predictions_zarr_structure(self):
```
jstlaurent, line 312: **Comment**
Since the Zarr archive won't be filled when the `Prediction` instance is created, this validation will not work. Also, it's not needed.

polaris/evaluate/_predictions.py
```diff
@@ -275,3 +286,169 @@ def __repr__(self):
...
+        return list(self.zarr_root.keys())
+
+    @property
+    def dtypes(self):
+        dtypes = {}
+        for arr in self.zarr_root.array_keys():
+            dtypes[arr] = self.zarr_root[arr].dtype
+        for group in self.zarr_root.group_keys():
+            dtypes[group] = object
+        return dtypes
```
jstlaurent, line 349: **Question**
This matches the API from the dataset, but I don't think we need it here.

polaris/evaluate/_predictions.py
```diff
@@ -275,3 +286,169 @@ def __repr__(self):
...
+    def zarr_manifest_md5sum(self, value: str):
+        if not re.fullmatch(r"^[a-f0-9]{32}$", value):
+            raise ValueError("The checksum should be the 32-character hexdigest of a 128 bit MD5 hash.")
+        self._zarr_manifest_md5sum = value
+
+    @property
+    def has_zarr_manifest_md5sum(self):
+        return self._zarr_manifest_md5sum is not None
+
+    def set_prediction_value(self, column: str, row: int, value: Any):
```
jstlaurent, line 391: **Comment**
I think this is broadly correct, but the users might be looking to set the values as arrays, rather than individually. @cwognum can probably let us know what API would be more useful.

jstlaurent (commented):
Left some comments about the model structure. We can chat about it some more if you'd like. 😄

polaris/evaluate/_predictions.py
j279li: That makes sense, I think we can also have both as well.

polaris/hub/client.py
cwognum: It speeds up access to cloud-based Zarr archives by reducing the number of files needed.

Given the smaller size of predictions, I don't think we need it. And even if it does become needed, it would just be us that pay the performance penalty. I would remove it, because it could complicate things a little by having to ensure the consolidated metadata and the actual Zarr archive stay in sync.

polaris/evaluate/_predictions.py
cwognum: > what about having both

I would prefer we stick with one. Even if it may not seem like it now, I could see this become one of those things where we end up maintaining two diverging code paths. We can always add it if users ask for it!

I think it's a safe assumption that predictions will always fit in memory and would thus provide an API that lets you set all of the predictions at once.

Ideally, setting them directly through the constructor would be nicest, I think:
```py
Predictions(
    predictions={
        'target_column_a': [0, 1, 2],
        'target_column_b': [3, 4, 5],
    }
)
```

For more complex objects, like `rdkit.Chem.Molecule` or `fastpdb.AtomArray`.

@jstlaurent I can remember us discussing this before and opting for dedicated `set_*()` methods, but my memory is failing me. Am I remembering that correctly?

jstlaurent: > For more complex objects, like `rdkit.Chem.Molecule` or `fastpdb.AtomArray`. @jstlaurent I can remember us discussing this before and opting for dedicated `set_*()` methods, but my memory is failing me. Am I remembering that correctly?

If the prediction Zarr arrays are initialized with the same codecs as the dataset columns, then the array elements passed in could be complex types and Zarr should encode them correctly.

We did have a discussion about how desirable it is to tightly couple the codecs classes defined in `polaris` with the Zarr archive definition, which we left up in the air.

j279li:
@jstlaurent I've encountered circular import issues with BenchmarkV2Specification, since the benchmark package already imports from the evaluate package. To work around this, I moved the Predictions class into its own package for now. Would that be alright, or do you have other ideas?

danielpeng1 (approved):
Nice job!

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,214 @@
...
+    def zarr_manifest_md5sum(self, value: str):
+        if not re.fullmatch(r"^[a-f0-9]{32}$", value):
+            raise ValueError("The checksum should be the 32-character hexdigest of a 128 bit MD5 hash.")
+        self._zarr_manifest_md5sum = value
+
+    @property
+    def has_zarr_manifest_md5sum(self):
+        return self._zarr_manifest_md5sum is not None
+
+    def upload_to_hub(
```
cwognum, line 197: **Suggestion:** Rather than including the upload method here, what do you think of having a `submit_predictions` method in the benchmark class? This is consistent with how we've done things for competitions. That way, we could automatically fill metadata such as the `benchmark` attribute in this class. [This is what we did for competitions](https://github.com/polaris-hub/polaris/blob/main/polaris/competition/__init__.py#L166)! 

The API would then look like:

```python
import polaris as po

# Load the benchmark from the Hub
benchmark = load_competition("dummy-user/dummy-name")

# Get the train and test data-loaders
train, test = benchmark.get_train_test_split()

# Work your magic to accurately predict the test set
prediction_values = np.array([0.0 for x in test])

# Submit your predictions
benchmark.submit_predictions(
    prediction_name="first-prediction",
    prediction_owner="dummy-user",
    report_url="REPORT_URL",
    predictions=prediction_values,
)
```

tests/test_dataset_v2.py
cwognum, line 1: **Suggestion:** Could we add some test casing for the predictions? At least make sure we test that the automatic conversion for object arrays works? It's non trivial to me that that works and having an associated test case would give me peace of mind. See e.g. [here](https://github.com/polaris-hub/polaris/blob/main/tests/test_benchmark_predictions.py).

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,214 @@
...
+    predictions: dict[str, list | np.ndarray | list[object]] = Field(exclude=True)
+
+    _zarr_root_path: str | None = PrivateAttr(None)
+    _zarr_manifest_path: str | None = PrivateAttr(None)
+    _zarr_manifest_md5sum: str | None = PrivateAttr(None)
+    _zarr_root: zarr.Group | None = PrivateAttr(None)
+    _cache_dir: str | None = PrivateAttr(None)
+
+    @model_validator(mode="after")
+    def _initialize_cache_and_zarr(self) -> Self:
```
cwognum, line 56: **Suggestion:** Rather than using a cache directory, I would create a temporary directory. I don't think we want this data to persist between sessions, and with a temp directory everything gets cleaned up automatically.

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,214 @@
...
+        """Set up the cache directory and initialize Zarr archive from predictions."""
+        # Set up cache dir if not already set
+        if self._cache_dir is None:
+            cache_root = Path(DEFAULT_CACHE_DIR) / _CACHE_SUBDIR
+            cache_root.mkdir(parents=True, exist_ok=True)
+            self._cache_dir = str(cache_root / str(uuid4()))
+            Path(self._cache_dir).mkdir(parents=True, exist_ok=True)
+        # Set zarr_root_path if not already set
+        if self._zarr_root_path is None:
+            self._zarr_root_path = str(Path(self._cache_dir) / "predictions.zarr")
```
cwognum, line 66: **Suggestion:** Since you have the `zarr_root_path` property already, I would move this code to there. That way, this method won't be needed and `_create_zarr_from_predictions` can become the model validator.

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,214 @@
...
+    def zarr_root_path(self) -> str:
+        return self._zarr_root_path
+
+    def _create_zarr_from_predictions(self):
+        """Create a Zarr archive from the predictions dictionary."""
+        # Ensure parent directory exists
+        os.makedirs(os.path.dirname(self.zarr_root_path), exist_ok=True)
+
+        store = zarr.DirectoryStore(self.zarr_root_path)
+        root = zarr.group(store=store)
```
cwognum, line 81: **Change requested:** Use the `zarr_root` property.

```suggestion
        root = self.zarr_root
```

You will have to change the `mode` with which you open the archive in that property to also allow for writing data.

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,214 @@
...
+    Attributes:
+        benchmark: The BenchmarkV2Specification instance this prediction is for.
+        model: (Optional) The Model artifact used to generate these predictions.
+        predictions: A dictionary mapping column names to prediction arrays/lists.
+    """
+
+    _artifact_type = "prediction"
+    benchmark: BenchmarkV2Specification
+    model: Model | None = Field(None, exclude=True)
+    predictions: dict[str, list | np.ndarray | list[object]] = Field(exclude=True)
```
cwognum, line 47: **Change requested:** Since we know the benchmark, we can do further verification and standardization, given us more confidence and the user a better experience. [We do so for Competition predictions as well](https://github.com/polaris-hub/polaris/blob/main/polaris/evaluate/_predictions.py#L71), but note that we need to tweak the code to support the `object` dtype.

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,214 @@
...
+
+        for col in target_cols:
+            if col not in self.predictions:
+                raise ValueError(f"Predictions for target column '{col}' not provided")
+
+            data = self.predictions[col]
+
+            # Determine the appropriate codec based on the dataset column annotation
+            codec_kwargs = {}
+            if hasattr(self.benchmark.dataset, "annotations") and col in self.benchmark.dataset.annotations:
```
cwognum, line 94: **Change requested:** `hasattr` is not needed here, since the v1 and v2 datasets inherit the `annotations` attribute from the `BaseDataset`.

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,214 @@
...
+                annotation = self.benchmark.dataset.annotations[col]
```
cwognum, line 95: **Change requested:** Rather than using the dataset annotations as the source of truth, we should use the dataset's Zarr archive to determine the dtype. Each prediction column should match the corresponding dataset column.

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,214 @@
...
+            self._zarr_root_path = str(Path(self._cache_dir) / "predictions.zarr")
+        # Create the Zarr archive with predictions at the specified zarr_root_path
+        self._create_zarr_from_predictions()
+        return self
+
+    @property
+    def zarr_root_path(self) -> str:
+        return self._zarr_root_path
+
+    def _create_zarr_from_predictions(self):
```
cwognum, line 75: **Change requested:** I think this method can be simplified a lot. It can copy the _configuration_ (dtype, codec, size, etc.) of each array from the benchmark's dataset's Zarr Root.

polaris/hub/client.py
```diff
@@ -714,3 +716,59 @@ def upload_model(
             progress.log(
                 f"[green]Your model has been successfully uploaded to the Hub. View it here: {model_url}"
             )
+
+    def upload_predictions(
```
cwognum, line 720: **Question:** This only works for BenchmarkV2, right? Do we check that somewhere, so we don't get orphaned predictions.

cwognum (changes requested):
Hi @j279li, sorry for not spotting any of this on my first review. I've been out of the loop on recent changes to Polaris, so I may also be missing some context here. 

We're making good progress, but I think we should take another look at how we've done things for competitions. Specifically, I would like us to revisit the user-facing API as well as the validation of the `Predictions` class.

Happy to find some chat to about this tomorrow!

polaris/hub/client.py
cwognum: Going through the Hub PR now, seems you've got it covered since you use the `retrieve` function of Benchmark V2 [here](https://github.com/polaris-hub/polaris-hub/blob/bee2b1dfcc32ca8ebf131b1adc4afba248f878e7/src/services/prediction-v2.ts#L52). That will throw a 404 if trying to use a V1 benchmark.

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,205 @@
...
+            self._zarr_root = zarr.group(store=store)
+        return self._zarr_root
+
+    @property
+    def zarr_root_path(self) -> str:
+        """Get the path to the Zarr archive root."""
+        if self._zarr_root_path is None:
+            # Create a temporary directory if not already set
+            if self._temp_dir is None:
+                self._temp_dir = tempfile.mkdtemp(prefix="polaris_predictions_")
```
cwognum, line 151: **Suggestion:** We could also clean up the directory. 

```py
def __del__(self) -> None:
    os.remove(self. _temp_dir)
```

cwognum (changes requested):
Thanks @j279li , we've made a lot of progress but there are still a couple of thinks I would like us to take a look at. Let me know if you have any questions!

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,205 @@
...
+        validator = TypeAdapter(IncomingPredictionsType, config={"arbitrary_types_allowed": True})
+        predictions = validator.validate_python(data["predictions"])
+
+        # Get benchmark and dataset info
+        benchmark = data["benchmark"]
+        dataset_root = benchmark.dataset.zarr_root
+
+        # Get expected size from test split
+        _, test = benchmark.get_train_test_split()
+        expected_size = len(test)
```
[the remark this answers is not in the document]
cwognum: We should indeed prepare for multiple test sets, even if the V2 benchmark doesn't support this yet. You can use the `test_set_sizes` property. 

However, as I shared in another comment, I think you can inherit most logic from the `BenchmarkPredictions` class.

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,205 @@
...
+from polaris.evaluate import ResultsMetadataV2
+from pydantic import TypeAdapter
+
+if TYPE_CHECKING:
+    from polaris.benchmark import BenchmarkV2Specification
+
+logger = logging.getLogger(__name__)
+
+
+class Predictions(ResultsMetadataV2):
```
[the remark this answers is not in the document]
Andrewq11: I'd agree with this.

polaris/prediction/_predictions_v2.py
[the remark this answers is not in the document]
Andrewq11: Ah yes, `BenchmarkPredictions`. Good catch.

cwognum (commented):
Sorry for the quick feedback, @j279li , but I saw the new commit and figured I would leave the feedback sooner rather than later.

polaris/benchmark/_benchmark_v2.py
```diff
@@ -154,3 +160,50 @@ def evaluate(
...
+            predictions=predictions,
+            contributors=contributors or [],
+            model=model,
+            description=description,
+            tags=tags or [],
+            user_attributes=user_attributes or {},
+        )
+
+        # Write predictions to Zarr archive before uploading
+        standardized_predictions.to_zarr()
```
cwognum, line 206: **Minor:** Should we move this into `client.submit_benchmark_predictions`, since it should happen whenever we upload predictions, not just when we call `submit_predictions`.

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,171 @@
...
+        _temp_dir (str | None): Temporary directory used for storing the Zarr archive.
+        predictions (dict): The predictions data, organized by test set and column.
+        contributors (list): List of contributors involved in generating the predictions.
+        model: The model used to generate the predictions.
+        description (str): Description of the predictions.
+        tags (list): Tags associated with the predictions.
+        user_attributes (dict): Additional user attributes for the predictions.
+    """
+
+    benchmark: BenchmarkV2Specification
```
cwognum, line 50: **Suggestion:** Rather than requiring the full `BenchmarkV2Specification`, I think it would be better to only require the properties we need. This makes testing easier and reduces overlap with `BenchmarkPredictions`, which already specifies most data you need from the benchmark (i.e. `target_labels`, `test_set_labels`, `test_set_sizes`). I believe the only other two things we need are the benchmark artifact ID and the dataset's Zarr root.

```suggestion
    dataset_zarr_root: zarr.Group 
    benchmark_artifact_id: str
```

This may also fix the circular import!

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,171 @@
...
+        _zarr_manifest_path (str | None): Path to the Zarr manifest file.
+        _zarr_manifest_md5sum (str | None): MD5 checksum of the Zarr manifest file.
+        _zarr_root (zarr.Group | None): The Zarr group object for the archive root.
+        _temp_dir (str | None): Temporary directory used for storing the Zarr archive.
+        predictions (dict): The predictions data, organized by test set and column.
+        contributors (list): List of contributors involved in generating the predictions.
+        model: The model used to generate the predictions.
+        description (str): Description of the predictions.
+        tags (list): Tags associated with the predictions.
+        user_attributes (dict): Additional user attributes for the predictions.
```
cwognum, line 47: **Minor:** Sharing this as an FYI, but feel free to leave things as is: We typically don't annotate private attributes (i.e. attributes starting with an underscore) in the docstring, since they're not user-facing. I typically also don't document the properties of the parent class, since it's extra effort to keep this docstring in sync with that model. What I rather do is something like: 
```
For additional metadata attributes, see the base classes.
```

(See e.g. [here](https://github.com/polaris-hub/polaris/blob/main/polaris/dataset/_dataset.py#L48C5-L48C62))

polaris/prediction/_predictions_v2.py
```diff
@@ -0,0 +1,171 @@
...
+                data = test_set_predictions[col]
+                template = dataset_root[col]
+                test_set_group.array(
+                    name=col,
+                    data=data,
+                    dtype=template.dtype,
+                    compressor=template.compressor,
+                    filters=template.filters,
+                    chunks=template.chunks,
+                    object_codec=getattr(template, "object_codec", None),
```
cwognum, line 93: **Minor**: Is the `getattr` needed here? Is `object_codec` not always specified?

tests/test_benchmark_predictions_v2.py
```diff
@@ -0,0 +1,157 @@
...
+        assert result.keys() == expected.keys()
+        for key in expected:
+            assert_deep_equal(result[key], expected[key])
+    elif isinstance(expected, np.ndarray):
+        assert np.array_equal(result, expected)
+    else:
+        assert result == expected
+
+
+@pytest.fixture(scope="function")
```
cwognum, line 26: **Suggestion:** Move all fixtures to `conftest.py`. for consistency. Without needing to import them, fixtures are made available as parameters for each test method.

tests/test_benchmark_predictions_v2.py
```diff
@@ -0,0 +1,157 @@
...
+    benchmark = BenchmarkV2Specification(
+        name="v2-benchmark-float-dtype",
+        dataset=dataset,
+        metrics=["mean_absolute_error"],
+        main_metric="mean_absolute_error",
+        split=split,
+        target_cols=["expt"],
+        input_cols=["smiles"],
+    )
+    return benchmark
```
cwognum, line 97: **Suggestion:** Could you use the existing [Dataset V2 fixture](https://github.com/polaris-hub/polaris/blob/main/tests/conftest.py#L129) as well? 

```suggestion
@pytest.fixture(scope="function")
def v2_benchmark_with_float_dtype(test_dataset_v2):
    train_indices = [0]
    test_indices = [1]
    split = SplitV2(training=IndexSet(indices=train_indices), test=IndexSet(indices=test_indices))
    benchmark = BenchmarkV2Specification(
        name="v2-benchmark-float-dtype",
        dataset=test_dataset_v2,
        metrics=["mean_absolute_error"],
        main_metric="mean_absolute_error",
        split=split,
        target_cols=["expt"],
        input_cols=["smiles"],
    )
    return benchmark
```

tests/test_benchmark_predictions_v2.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+@pytest.fixture(scope="function")
+def v2_benchmark_with_float_dtype(tmp_path, test_org_owner):
```
cwognum, line 64: **Suggestion:** This could be a general `test_benchmark_v2` fixture, since we don't have one yet. 

```suggestion
def test_benchmark_v2(test_dataset_v2):
```

tests/test_benchmark_predictions_v2.py
```diff
@@ -0,0 +1,157 @@
...
+        test_set_labels=["test"],
+        test_set_sizes={"test": 2},
+    )
+    assert isinstance(bp.predictions["test"]["expt"], np.ndarray)
+    assert bp.predictions["test"]["expt"].dtype == object
+    assert_deep_equal(bp.predictions, {"test": {"expt": np.array(mols, dtype=object)}})
+    # Check Zarr archive
+    zarr_path = bp.to_zarr()
+    assert zarr_path.exists()
+    root = zarr.open(str(zarr_path), mode="r")
```
cwognum, line 116: **Suggestion:** I would also check that the object codec is correctly set as expected, since that's the main thing the BenchmarkPredictionsV2 class does.

cwognum (approved):
I'm left with some minor comments, but this looks good to me! Thanks @j279li, great work ! 🚀

tests/test_benchmark_predictions_v2.py
cwognum: For `test_v2_fastpdb_object_dtype` too!

polaris/benchmark/_benchmark_v2.py
j279li: That's a good idea, will do

polaris/prediction/_predictions_v2.py
j279li: After further testing, it seems that zarr objects store codecs as `filters`, not as a separate attribute, so we should be able to remove `object_codec` entirely.

j279li: I'm on zarr v2 right now (2.18.7).

cwognum: Perfect! All good then.

PR #295, opened 2025-07-16
Refactor/Chore: Remove metrics and evaluation from BenchmarkV2Specification

j279li:
This pull request remove metrics from 'BenchmarkV2Specification' and removes client side evaluation. 

### Core Framework Updates:
* Refactored `PredictiveTaskSpecificationMixin` to remove metrics-related fields and added a new `task_type` property for high-level task categorization. (`polaris/benchmark/_task.py`) 
* Introduced `PredictiveTaskSpecificationMixinWithMetrics` to maintain backwards compatibility for benchmarks requiring metrics (V1) and updated imports and class definitions to use this mixin where applicable. (`polaris/benchmark/_base.py`, `polaris/competition/__init__.py`). 
* Refactored  `BenchmarkV2Specification` to inherit directly from `PredictiveTaskSpecificationMixin` and `BaseArtifactModel`, bypassing the middle `BenchmarkSpecification` abstraction. (`polaris/benchmark/_benchmark_v2.py`)

### Test Updates:
* Updated test cases to reflect the removal of metrics from BenchmarkV2. (`tests/conftest.py`, `tests/test_benchmark_v2.py`).

### DIFF
--- a/polaris/benchmark/__init__.py
+++ b/polaris/benchmark/__init__.py
@@ -1,19 +1,12 @@
-from polaris.benchmark._base import BenchmarkV1Specification, BenchmarkV1Specification as BenchmarkSpecification
+from polaris.benchmark._base import (
+    BenchmarkV1Specification,
+    BenchmarkV1Specification as BenchmarkSpecification,
+)
 from polaris.benchmark._benchmark_v2 import BenchmarkV2Specification
 from polaris.benchmark._definitions import MultiTaskBenchmarkSpecification, SingleTaskBenchmarkSpecification
 
 __all__ = [
     "BenchmarkSpecification",
-    "BenchmarkV1Specification",
-    "BenchmarkV2Specification",
-    "SingleTaskBenchmarkSpecification",
-    "MultiTaskBenchmarkSpecification",
-]
-
-from polaris.benchmark._benchmark_v2 import BenchmarkV2Specification
-from polaris.benchmark._definitions import MultiTaskBenchmarkSpecification, SingleTaskBenchmarkSpecification
-
-__all__ = [
     "BenchmarkV1Specification",
     "BenchmarkV2Specification",
     "SingleTaskBenchmarkSpecification",


polaris/benchmark/_task.py
```diff
@@ -59,6 +58,51 @@ def _parse_target_types(
...
+
+    @computed_field
+    @property
+    def task_type(self) -> str:
+        """The high-level task type of the benchmark."""
+        v = TaskType.MULTI_TASK if len(self.target_cols) > 1 else TaskType.SINGLE_TASK
+        return v.value
+
+
+class PredictiveTaskSpecificationMixinWithMetrics(PredictiveTaskSpecificationMixin):
```
cwognum, line 92: **Suggestion:** Rather than having a `PredictiveTaskSpecificationMixinWithMetrics`, I'd suggest to either: 

1. Create a new `MetricsSpecificationMixin` that holds all logic and attributes for the metrics. 
2. Move the all metric logic and attributes directly to the `BenchmarkSpecification`, since I believe that's the only class the uses it in the end. Both competitions and the V2 benchmarks don't, since they submit predictions.

I would prefer (2) since it simplifies the code. We're not having too many mixin classes that then only get used once. What do you think?

polaris/benchmark/_base.py
cwognum, line 1: **Suggestion**: Since the Benchmark V2 no longer inherits from this class, we can simplify this code. E.g. we should be able to merge `BenchmarkSpecification` and `BenchmarkV1Specification` into a single class.

cwognum (commented):
Thanks, @j279li ! Short and sweet. 

I think we can simplify this some more. Take a look and let me know what you think!

polaris/benchmark/_task.py
j279li: Yeah, I think that could work and make this a bit cleaner, certainly.

polaris/benchmark/_base.py
j279li: Sounds good!

tests/conftest.py
```diff
@@ -426,6 +424,7 @@ def v2_benchmark_with_rdkit_object_dtype(tmp_path, test_org_owner):
         dtype=object,
         object_codec=RDKitMolCodec(),
     )
+    root.array("smiles", data=np.array(["CCO", "CCC"]))  # Add smiles column
```
Andrewq11, line 427: Nice catch on these. Out of curiosity, were tests that inherited this fixture failing due to the absence of this column?

polaris/competition/__init__.py
```diff
@@ -22,7 +22,9 @@
 )
 
 
-class CompetitionSpecification(DatasetV2, PredictiveTaskSpecificationMixin, SplitSpecificationV1Mixin):
+class CompetitionSpecification(
+    DatasetV2, PredictiveTaskSpecificationMixinWithMetrics, SplitSpecificationV1Mixin
```
Andrewq11, line 26: I'm sure you'll get this in the next round of revisions, but just another heads up that competitions don't need access to the metrics.

@cwognum's suggestion about moving all metrics logic to the `BenchmarkV1Specification` would be quite nice!

tests/conftest.py
j279li: They weren't any that were failing, I think.

cwognum (approved):
Thanks @j279li ! 🚢 it!

tests/conftest.py
j279li: I'll double check.

Andrewq11 (commented):
Awesome work, @j279li! I've left just a couple comments we should resolve quickly.

PR #296, opened 2025-07-17
Feat: Add Support for multiple test-set benchmarks

j279li:
This pull request introduces support for multiple test sets in the `SplitV2` class and related functionality across the codebase. The changes enhance flexibility by allowing benchmarks to define more than one test set, while maintaining backward compatibility with the previous single-test-set format.

### Enhancements to `SplitV2` functionality:

* Modified the `SplitV2` class to replace the single `test` field with a `test_sets` dictionary, enabling support for multiple test sets. Added validators to ensure proper deserialization, validation, and backward compatibility for single-test-set benchmarks.

### Changes to benchmark specification and client:

* Updated the `BenchmarkV2Paths` class to allow dynamic fields for multiple test sets and added a `test_sets` property for structured access to test set URLs. Maintained backward compatibility with the single `test` field.
* Modified `_get_v2_benchmark` in `polaris/hub/client.py` to handle multiple test sets when parsing split data from the response. Structured test sets separately from the training set.

### Testing improvements:

* Added new test cases in `tests/test_benchmark_v2.py` to validate functionality for multiple test sets, including validation rules, backward compatibility, and split properties.
* Introduced a new pytest fixture `test_benchmark_v2_multiple_test_sets` in `tests/conftest.py` for creating benchmarks with multiple test sets.


tests/test_benchmark_v2.py
```diff
@@ -147,3 +147,109 @@ def test_benchmark_v2_n_classes_validation(test_dataset_v2):
...
+    train, test = benchmark.get_train_test_split()
+    assert len(train) == 5
+    assert isinstance(test, dict)
+    assert len(test) == 3
+    assert len(test["test_set_1"]) == 3
+    assert len(test["test_set_2"]) == 2
+    assert len(test["test_set_3"]) == 4
+
+    # Test backward compatibility - accessing .test property should work
+    assert split.test.datapoints > 0  # Should return first test set or combined
```
Andrewq11, line 193: Noting this as related to my previous comment about returning the first test set if `test` is accessed.

tests/conftest.py
```diff
@@ -410,6 +410,29 @@ def test_benchmark_v2(test_dataset_v2, test_org_owner):
     return benchmark
 
 
+@pytest.fixture(scope="function")
+def test_benchmark_v2_multiple_test_sets(test_dataset_v2, test_org_owner):
```
Andrewq11, line 414: I don't believe this fixture is being used currently, and it would be useful in the new test you've created with the same name. We should also update their names so that they do not conflict.

tests/test_benchmark_v2.py
```diff
@@ -147,3 +147,109 @@ def test_benchmark_v2_n_classes_validation(test_dataset_v2):
...
+    benchmark = BenchmarkV2Specification(
+        name="test-benchmark-multiple-test-sets",
+        owner=test_org_owner,
+        dataset=test_dataset_v2,
+        split=split,
+        target_cols=["A"],
+        input_cols=["B"],
+        metrics=["mean_absolute_error"],
+        main_metric="mean_absolute_error",
+    )
```
Andrewq11, line 174: I would remove this and import the pytest fixture you created in `conftest.py`.

polaris/benchmark/_split_v2.py
```diff
@@ -58,80 +58,110 @@ def deserialize(index_set: bytes) -> "IndexSet":
...
         if isinstance(v, bytes):
             return IndexSet.deserialize(v)
         return v
 
+    @field_validator("test_sets", mode="before")
+    @classmethod
+    def _parse_test_index_sets(cls, v: dict[str, bytes | IndexSet]) -> dict[str, IndexSet]:
+        """Parse test sets from bytes or IndexSet objects"""
+        if not isinstance(v, dict):
+            return {}
```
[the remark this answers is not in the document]
j279li: It was a fallback for an incorrect input, yes, I'll raise a more appropriate error.

cwognum (changes requested):
Nice work, @j279li ! 

The change I suggested to switch to multiple (train, test) pairs (rather than assuming a single train set) does increase the scope of this PR a bit (e.g. you'll have to change the data structures on the Hub's side too), but I think this is a change we would like to make sometime soon and doing it now that there are no V2 benchmarks yet will simplify our lives later on.

What do you think of that?

polaris/hub/oauth.py
```diff
@@ -95,8 +95,7 @@ class DatasetV2Paths(ArtifactPaths):
 
 class BenchmarkV2Paths(ArtifactPaths):
     training: AnyUrlString = Field(json_schema_extra={"file": True})
-    test: AnyUrlString = Field(json_schema_extra={"file": True})
-    test_2: int = 0
```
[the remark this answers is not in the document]
j279li: I think so, yeah..

polaris/benchmark/_split_v2.py
```diff
@@ -56,82 +56,140 @@ def deserialize(index_set: bytes) -> "IndexSet":
...
+    A single train-test split pair containing training and test index sets.
+
+    This intermediate class represents one train-test split, which allows
+    SplitV2 to support multiple such pairs for cross-validation scenarios.
+    """
+
     training: IndexSet
     test: IndexSet
 
-    @field_validator("training", "test", mode="before")
```
cwognum, line 63: **Comment:** Since the validation for `training` and `test` is the exact same, you don't need separate validators. You can specify both variables in the `@field_validator` decorator like we had before!

polaris/benchmark/_split_v2.py
```diff
@@ -56,82 +56,140 @@ def deserialize(index_set: bytes) -> "IndexSet":
...
+    def n_test_datapoints(self) -> int:
+        """The size of the test set."""
+        return self.test.datapoints
+
+    @property
+    def max_index(self) -> int:
+        """Maximum index across train and test sets"""
+        max_indices = []
+
+        # Only add max if the bitmap is not empty
```
cwognum, line 126: **Question:** Why is this needed?

polaris/benchmark/_split_v2.py
```diff
@@ -56,82 +56,140 @@ def deserialize(index_set: bytes) -> "IndexSet":
...
+        if len(self.training.indices) > 0:
+            max_indices.append(self.training.indices.max())
+        if len(self.test.indices) > 0:
+            max_indices.append(self.test.indices.max())
```
cwognum, line 130: **Suggestion:** This will never happen, given that a `@field_validator` checks that this is not the case. This also clearly signals we'll never take the `max()` of an empty iterable, which would throw an exception.

```suggestion
        max_indices.append(self.test.indices.max())
```

polaris/benchmark/_split_v2.py
```diff
@@ -56,82 +56,140 @@ def deserialize(index_set: bytes) -> "IndexSet":
...
+        if not v:
+            raise InvalidBenchmarkError("At least one split must be specified")
+
+        parsed_splits = {}
+        for label, split_data in v.items():
+            if isinstance(split_data, dict):
+                parsed_splits[label] = TrainTestIndices(**split_data)
+            else:
+                parsed_splits[label] = split_data
+        return parsed_splits
```
cwognum, line 162: **Suggestion:** I don't think any of this is needed. Pydantic will automatically throw an error if something is not a dictionary or specified. Additionally, I believe (but I'm less sure about this) it will automatically try convert dictionaries to the appropriate Pydantic class.

polaris/benchmark/_split_v2.py
```diff
@@ -56,82 +56,140 @@ def deserialize(index_set: bytes) -> "IndexSet":
...
-    def test_items(self) -> Generator[tuple[str, IndexSet], None, None]:
-        # TODO: Until we support multi-test benchmarks
-        yield "test", self.test
+    def split_items(self) -> Generator[tuple[str, TrainTestIndices], None, None]:
+        """Yield all splits with their labels"""
+        for label, split in self.splits.items():
+            yield label, split
 
 
 class SplitSpecificationV2Mixin(BaseModel):
```
cwognum, line 195: **Question:** Seeing how `SplitSpecificationV2Mixin` and `SplitV2` overlap quite a bit, could we merge them into one?

cwognum (commented):
Hi @j279li , I wasn't sure if this was ready for review yet, but I took a quick look through everything except for the test cases, to get you some feedback early. Hope it's helpful!

TL;DR - This looks great. We'll break backwards compatibility between V1 and V2 of the benchmark (e.g. changing return format of the `get_train_test_splits()`), but that's ok since (1) there's really no V2s yet, (2) we clearly mark this as a new version, and (3) we will likely be the only ones creating these benchmarks.

polaris/benchmark/_split_v2.py
j279li: max_index right? I think it was a to-do in the old splitv2 that hadn't been implemented for multi-test sets. I believe it's mainly there to make sure the indices actually exist. In `benchmarkspecificationv2`, we have :         
```
if self.split.max_index >= dataset_length:
            raise InvalidBenchmarkError("The predefined split contains invalid indices")
```

cwognum: >        # Only add max if the bitmap is not empty

I specifically meant ☝️ 

Why is the check for not empty needed? Wouldn't it just return 0?

cwognum: Clear, thanks!

polaris/benchmark/_split_v2.py
cwognum: Let me take a closer look again later, but can you speak to what the `SplitSpecificationV2Mixin` adds, with regards to `SplitV2`? From the diff during review, it felt like it was mostly passing information through.

j279li: Yeah, from my understanding, it's basically just a way of making the API a bit cleaner, a wrapper that lets us call `benchmark.n_splits` instead of `benchmark.split.n_splits`. I guess it isn't that necessary in the end. I could just remove it and add splitv2 directly as a field in the benchmark specification, which should be a-ok as well. 
`

danielpeng1 (approved):

cwognum (approved):
Looking good! 🚢  it!

polaris/hub/client.py
```diff
@@ -484,10 +484,22 @@ def _get_v2_benchmark(self, owner: str | HubOwner, slug: str) -> BenchmarkV2Spec
...
+        splits = {}
+
+        for split_label, split_urls in split_data.items():
+            # Each split should have 'training' and 'test' objects with filePath, datapoints, md5Checksum
+            split_indices = {}
+            for data_type, url_info in split_urls.items():
+                # Extract the actual URL from the filePath field
+                url = url_info["filePath"]
+                with fsspec.open(url, mode="rb") as f:
+                    split_indices[data_type] = f.read()
```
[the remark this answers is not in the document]
j279li: Yes, I will do that.

PR #298, opened 2025-08-06
Fix/prediction upload

j279li:
This pull request introduces improvements to how Zarr object codecs and chunking are handled for prediction outputs, as well as some minor serialization and documentation fixes. The main focus is on making object codec support explicit and robust, especially for custom codecs, and ensuring correct serialization behavior for non-JSON-serializable fields.

Enhancements to Zarr object codec and chunking support:

* Added a utility function `detect_object_codec_and_chunking` in `polaris/utils/zarr/codecs.py` to determine the correct object codec, filter list, and chunking compatibility from template filters.
* Made codec chunking support explicit by adding a `supports_chunking` attribute to `RDKitMolCodec` (set to `True`) and `AtomArrayCodec` (set to `False`).
* Improved handling of `None` values in `AtomArrayCodec.encode` to explicitly set missing values in packed arrays.

Serialization and model improvements:

* Updated `BenchmarkPredictionsV2` to exclude the non-serializable `dataset_zarr_root` attribute from JSON serialization using a Pydantic `Field` directive, and modified `__repr__` to exclude the `predictions` field as they are also stored in zarr.

Minor documentation cleanup:

* Removed an outdated docstring reference to `splits` in `BenchmarkV2Specification`.

### DIFF
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -766,7 +766,7 @@ class PolarisHubClient(OAuth2Client):
 
         # Set owner
         prediction.owner = HubOwner.normalize(owner or prediction.owner)
-        prediction_json = prediction.model_dump(by_alias=True, exclude_none=True)
+        prediction_json = prediction.model_dump(by_alias=True, exclude_none=True, exclude={"predictions"})
 
         # Step 1: Upload metadata to Hub
         with track_progress(description="Uploading prediction metadata", total=1) as (progress, task):
--- a/polaris/hub/client.py
+++ b/polaris/hub/client.py
@@ -766,7 +766,7 @@ class PolarisHubClient(OAuth2Client):
 
         # Set owner
         prediction.owner = HubOwner.normalize(owner or prediction.owner)
-        prediction_json = prediction.model_dump(by_alias=True, exclude_none=True, exclude={"predictions"})
+        prediction_json = prediction.model_dump(by_alias=True, exclude_none=True)
 
         # Step 1: Upload metadata to Hub
         with track_progress(description="Uploading prediction metadata", total=1) as (progress, task):
--- a/polaris/prediction/_predictions_v2.py
+++ b/polaris/prediction/_predictions_v2.py
@@ -9,10 +9,12 @@ import numpy as np
 import zarr
 from pydantic import (
     PrivateAttr,
+    Field,
     model_validator,
 )
 
 from polaris.utils.zarr._manifest import generate_zarr_manifest, calculate_file_md5
+from polaris.utils.zarr.codecs import detect_object_codec_and_chunking
 from polaris.evaluate import ResultsMetadataV2
 from polaris.evaluate._predictions import BenchmarkPredictions
 
@@ -33,7 +35,7 @@ class BenchmarkPredictionsV2(BenchmarkPredictions, ResultsMetadataV2):
     For additional metadata attributes, see the base classes.
     """
 
-    dataset_zarr_root: zarr.Group
+    dataset_zarr_root: zarr.Group = Field(exclude=True)  # Zarr Group cannot be JSON serialized
     benchmark_artifact_id: str
     _artifact_type = "prediction"
     _zarr_root_path: str | None = PrivateAttr(None)
@@ -70,13 +72,28 @@ class BenchmarkPredictionsV2(BenchmarkPredictions, ResultsMetadataV2):
             for col in self.target_labels:
                 data = test_set_predictions[col]
                 template = dataset_root[col]
+                
+                # Use utility function to detect codec and chunking compatibility
+                chunks = template.chunks
+                if template.dtype == object:
+                    object_codec, filters, chunks_compatible = detect_object_codec_and_chunking(
+                        template.filters
+                    )
+                    # Disable chunking if codec doesn't support it
+                    if not chunks_compatible:
+                        chunks = None
+                else:
+                    object_codec = None
+                    filters = list(template.filters) if template.filters else []
+                
                 test_set_group.array(
                     name=col,
                     data=data,
                     dtype=template.dtype,
                     compressor=template.compressor,
-                    filters=template.filters,
-                    chunks=template.chunks,
+                    filters=filters,
+                    chunks=chunks,
+                    object_codec=object_codec,
                     overwrite=True,
                 )
 
@@ -142,7 +159,7 @@ class BenchmarkPredictionsV2(BenchmarkPredictions, ResultsMetadataV2):
         return self._zarr_manifest_md5sum is not None
 
     def __repr__(self):
-        return self.model_dump_json(by_alias=True, indent=2)
+        return self.model_dump_json(by_alias=True, exclude={"predictions"}, indent=2)
 
     def __str__(self):
         return self.__repr__()
--- a/polaris/prediction/_predictions_v2.py
+++ b/polaris/prediction/_predictions_v2.py
@@ -32,8 +32,10 @@ logger = logging.getLogger(__name__)
 # Reserved metadata key for storing original Python type
 RESERVED_TYPE_KEY = "python_type"
 
+
 class ReservedTypes(str, Enum):
     """Reserved type identifiers for object data stored in Zarr arrays."""
+
     RDKIT_MOL = "rdkit.Chem.Mol"
     ATOM_ARRAY = "biotite.structure.AtomArray"
 
--- a/polaris/prediction/_predictions_v2.py
+++ b/polaris/prediction/_predictions_v2.py
@@ -85,12 +85,20 @@ class BenchmarkPredictionsV2(BenchmarkPredictions, ResultsMetadataV2):
                 # Handle object data conversion
                 if template.dtype == object:
                     sample = next((item for item in data if item is not None), None)
-                    
+
                     # Define object type handlers
                     if isinstance(sample, Chem.Mol):
-                        object_codec, final_data, filters = VLenBytes(), [convert_mol_to_bytes(item) for item in data], None
+                        object_codec, final_data, filters = (
+                            VLenBytes(),
+                            [convert_mol_to_bytes(item) for item in data],
+                            None,
+                        )
                     elif isinstance(sample, struc.AtomArray):
-                        object_codec, final_data, filters = MsgPack(), [convert_atomarray_to_dict(item) for item in data], None
+                        object_codec, final_data, filters = (
+                            MsgPack(),
+                            [convert_atomarray_to_dict(item) for item in data],
+                            None,
+                        )
                     else:
                         object_codec, final_data, filters = None, list(data), template.filters
 
@@ -109,7 +117,7 @@ class BenchmarkPredictionsV2(BenchmarkPredictions, ResultsMetadataV2):
                     # Non-object data uses original data and template filters
                     final_data = data
                     filters = template.filters
-                    
+
                     test_set_group.array(
                         name=col,
                         data=final_data,
--- a/polaris/prediction/_predictions_v2.py
+++ b/polaris/prediction/_predictions_v2.py
@@ -4,6 +4,7 @@ import re
 import shutil
 from pathlib import Path
 import tempfile
+from enum import StrEnum
 
 import numpy as np
 import zarr
@@ -28,6 +29,16 @@ from polaris.evaluate._predictions import BenchmarkPredictions
 
 logger = logging.getLogger(__name__)
 
+# Reserved metadata key for storing original Python type
+RESERVED_TYPE_KEY = "python_type"
+
+
+class ReservedTypes(StrEnum):
+    """Reserved type identifiers for object data stored in Zarr arrays."""
+
+    RDKIT_MOL = "rdkit.Chem.Mol"
+    ATOM_ARRAY = "biotite.structure.AtomArray"
+
 
 class BenchmarkPredictionsV2(BenchmarkPredictions, ResultsMetadataV2):
     """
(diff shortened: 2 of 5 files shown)


cwognum (commented):
In line with my comments in the Hub PR: 

Rather than having these workarounds, I think it makes more sense to ditch the custom codecs. We can use default codecs (i.e. `MsgPack` for AtomArrays and `VLenBytes` for RDKit mols), to ensure it remains a valid default Zarr archive that any machine can open, and then convert from these formats to the objects we need internally.

This would be a bigger change than you've signed up for, though, as it also requires non-trivial changes to the dataset class. Let's do the following: 

1. For datasets, we keep using the custom codecs
2. For predictions, we use default Zarr codecs and add conversion code on the client and service side.

Does that make sense?

polaris/utils/zarr/codecs.py
```diff
@@ -149,3 +150,76 @@ def decode(self, buf, out=None):
 
 register_codec(RDKitMolCodec)
 register_codec(AtomArrayCodec)
+
+
+def convert_atomarray_to_dict(atom_array):
```
cwognum: **Suggestion:** Could you add type-hints? 

```suggestion
def convert_atomarray_to_dict(atom_array: AtomArray | None) -> dict[str, list] | None:
```

polaris/utils/zarr/codecs.py
```diff
@@ -149,3 +150,76 @@ def decode(self, buf, out=None):
...
+        "element": atom_array.element,
+        "atom_id": atom_array.atom_id,
+        "b_factor": atom_array.b_factor,
+        "occupancy": atom_array.occupancy,
+        "charge": atom_array.charge,
+    }
+    return {k: v.tolist() for k, v in data.items()}
+
+
+def convert_dict_to_atomarray(data):
```
cwognum: **Suggestion:** Could you add type-hints? 

```suggestion
def convert_dict_to_atomarray(data: dict) -> AtomArray:
```

polaris/utils/zarr/codecs.py
```diff
@@ -149,3 +150,76 @@ def decode(self, buf, out=None):
...
+            occupancy=data["occupancy"][ind],
+            charge=data["charge"][ind],
+            atom_id=data["atom_id"][ind],
+        )
+        atom_array.append(atom)
+
+    return struc.array(atom_array)
+
+
+def convert_mol_to_bytes(mol):
```
cwognum: **Suggestion:** Could you add type-hints?

```suggestion
def convert_mol_to_bytes(mol: rdkit.Chem.Mol | None) -> bytes:
```

polaris/utils/zarr/codecs.py
```diff
@@ -149,3 +150,76 @@ def decode(self, buf, out=None):
...
+        return b""
+
+    if not isinstance(mol, Chem.Mol):
+        raise ValueError(f"Expected an RDKitMol, but got {type(mol)} instead.")
+
+    props = Chem.PropertyPickleOptions.AllProps
+    return mol.ToBinary(props)
+
+
+def convert_bytes_to_mol(mol_bytes):
```
cwognum: **Suggestion:** Could you add type hints?

```suggestion
def convert_bytes_to_mol(mol_bytes: bytes) -> rdkit.Chem.Mol:
```

polaris/utils/zarr/codecs.py
```diff
@@ -149,3 +150,76 @@ def decode(self, buf, out=None):
...
+        )
+        atom_array.append(atom)
+
+    return struc.array(atom_array)
+
+
+def convert_mol_to_bytes(mol):
+    """Convert RDKit Mol to bytes that can be stored with standard VLenBytes codec."""
+    if mol is None or (isinstance(mol, bytes) and len(mol) == 0):
+        return b""
```
cwognum: **Suggestion:** Let's not account for an empty byte string here! 

```suggestion
    if mol is None:
        return b""
```

polaris/utils/zarr/codecs.py
```diff
@@ -149,3 +150,76 @@ def decode(self, buf, out=None):
...
 register_codec(AtomArrayCodec)
+
+
+def convert_atomarray_to_dict(atom_array):
+    """Convert AtomArray to a dict that can be stored with standard MsgPack codec."""
+    if atom_array is None:
+        return None
+
+    if not isinstance(atom_array, struc.AtomArray):
+        raise ValueError(f"Expected an AtomArray, but got {type(atom_array)} instead")
```
cwognum: **Suggestion:** TypeError?

```suggestion
        raise TypeError(f"Expected an AtomArray, but got {type(atom_array)} instead")
```

polaris/utils/zarr/codecs.py
```diff
@@ -149,3 +150,76 @@ def decode(self, buf, out=None):
...
+    return struc.array(atom_array)
+
+
+def convert_mol_to_bytes(mol):
+    """Convert RDKit Mol to bytes that can be stored with standard VLenBytes codec."""
+    if mol is None or (isinstance(mol, bytes) and len(mol) == 0):
+        return b""
+
+    if not isinstance(mol, Chem.Mol):
+        raise ValueError(f"Expected an RDKitMol, but got {type(mol)} instead.")
```
cwognum: **Suggestion:** TypeError?

```suggestion
        raise TypeError(f"Expected an RDKitMol, but got {type(mol)} instead.")
```

polaris/hub/client.py
```diff
@@ -766,7 +766,7 @@ def submit_benchmark_predictions(
 
         # Set owner
         prediction.owner = HubOwner.normalize(owner or prediction.owner)
-        prediction_json = prediction.model_dump(by_alias=True, exclude_none=True)
+        prediction_json = prediction.model_dump(by_alias=True, exclude_none=True, exclude={"predictions"})
```
cwognum: **Suggestion:** Rather than defining it here as `exclude=...`, you can do the same as you did for `BenchmarkPredictionsV2. dataset_zarr_root` and exclude it through a `Field` object.

polaris/prediction/_predictions_v2.py
```diff
@@ -142,7 +241,7 @@ def has_zarr_manifest_md5sum(self):
         return self._zarr_manifest_md5sum is not None
 
     def __repr__(self):
-        return self.model_dump_json(by_alias=True, indent=2)
+        return self.model_dump_json(by_alias=True, exclude={"predictions"}, indent=2)
```
cwognum: **Suggestion:** If you follow my earlier suggestion about moving this to a `Field`, you won't have to repeat this here.

polaris/prediction/_predictions_v2.py
```diff
@@ -61,33 +71,122 @@ def to_zarr(self) -> Path:
...
+        return converted_predictions
+
     @property
     def zarr_root(self) -> zarr.Group:
         """Get the zarr Group object corresponding to the root, creating it if it doesn't exist."""
         if self._zarr_root is None:
             store = zarr.DirectoryStore(self.zarr_root_path)
-            self._zarr_root = zarr.group(store=store)
+            raw_root = zarr.group(store=store)
+            self._zarr_root = raw_root
```
cwognum: **Question:** Out of curiosity, why this change?

polaris/prediction/_predictions_v2.py
```diff
@@ -61,33 +71,122 @@ def to_zarr(self) -> Path:
...
+
+        Returns:
+            Full predictions dictionary with same structure as self.predictions,
+            but with object data converted back to original types (AtomArray, RDKit Mol)
+        """
+        converted_predictions = {}
+
+        for test_set_label in self.test_set_labels:
+            if test_set_label not in self.zarr_root:
+                continue
```
cwognum: **Question:** Do we expect this to happen? If not, we should use `root.require_group` like we did before.

polaris/prediction/_predictions_v2.py
```diff
@@ -61,33 +71,122 @@ def to_zarr(self) -> Path:
...
+
+            test_set_group = self.zarr_root[test_set_label]
+            converted_predictions[test_set_label] = {}
+
+            for target in self.target_labels:
+                if target not in test_set_group:
+                    continue
```
cwognum: **Question:** Do we expect this to happen? If not, we should use `root.require_group` like we did before.

polaris/prediction/_predictions_v2.py
```diff
@@ -61,33 +71,122 @@ def to_zarr(self) -> Path:
...
+                template = self.dataset_zarr_root[target]
+                if template.dtype == object:
+                    # Use filters to determine conversion (simple and reliable)
+                    filters = zarr_array.filters or []
+                    if any(isinstance(f, MsgPack) for f in filters):
+                        converter = convert_dict_to_atomarray
+                    elif any(isinstance(f, VLenBytes) for f in filters):
+                        converter = convert_bytes_to_mol
+                    else:
+                        converter = None
```
cwognum: **Comment:** This won't work. It will be many-to-one mapping from encoders to internal types. In other words, not every `MsgPack` array will be an `AtomArray` and not every `VLenBytes` will be a RDKit `Mol`.

This could get tricky... How about we store the converter function in the Zarr metadata for now? We can use a string label to identify it?

polaris/prediction/_predictions_v2.py
```diff
@@ -61,33 +71,122 @@ def to_zarr(self) -> Path:
...
+
+                    # Apply conversion
+                    if converter:
+                        converted = [converter(item) if item is not None else None for item in data]
```
cwognum: **Suggestion:** You don't need to check for `None` here. You already do so in the converter functions. 

```suggestion
                        converted = [converter(item) for item in data]
```

polaris/prediction/_predictions_v2.py
```diff
@@ -61,33 +71,122 @@ def to_zarr(self) -> Path:
...
                     compressor=template.compressor,
-                    filters=template.filters,
+                    filters=filters,
                     chunks=template.chunks,
                     overwrite=True,
                 )
 
         return Path(self.zarr_root_path)
 
+    def get_converted_predictions(self) -> dict:
```
cwognum: **Suggestion** When would this method be used? This class doesn't really read to Zarr, but only write to it, right? So the internal representation would always be in the "right" (i.e. converted) dtype?

polaris/prediction/_predictions_v2.py
```diff
@@ -61,33 +71,122 @@ def to_zarr(self) -> Path:
 
         This method should be called explicitly when ready to write predictions to disk.
         """
-        root = self.zarr_root
+        # Get raw zarr root for writing (not the converting wrapper)
+        store = zarr.DirectoryStore(self.zarr_root_path)
+        raw_root = zarr.group(store=store)
```
cwognum: **Question:** Why is this needed?

polaris/prediction/_predictions_v2.py
```diff
@@ -61,33 +71,122 @@ def to_zarr(self) -> Path:
...
+                            final_data = list(data)
+                    else:
+                        # All items are None
+                        codec = MsgPack()
+                        final_data = list(data)
+
+                    # Object data uses converted data and custom filters
+                    filters = [codec]
+                else:
+                    # Non-object data uses original data and template filters
```
cwognum, line 117: **Suggestion:** Anything that's _not_ a Mol or AtomArray should fall back to the dataset as a source of truth.

I would simplify this whole piece to something like: 
```py

data = test_set_predictions[col]
template = dataset_root[col]

sample = next((item for item in data if item is not None), None)

if isinstance(sample, rdkit.Chem.Mol): 
    codec = VLenBytes()
    data = convert_mol_to_bytes(data)
elif isinstance(sample, AtomArray): 
    codec = MsgPack()
    data = convert_mol_to_bytes(data)
else:
    filters = template.filters
```

polaris/prediction/_predictions_v2.py
```diff
@@ -61,33 +71,122 @@ def to_zarr(self) -> Path:
...
+                    final_data = data
+                    filters = template.filters
+
+                # Single array creation for both cases
                 test_set_group.array(
```
cwognum: **Suggestion:** If you agree with me, you'll have to set metadata here regarding the converter function. I would specify that as an `StrEnum`. Something like: 

```py
class ReservedTypeKey = "python_type"

class ReservedTypes(StrEnum):
    RDKIT_MOL = 'rdkit.Chem.Mol'
    ATOM_ARRAY = 'biotite.structure.AtomArray'

test_set_group.array(
    ...
    attributes={ReservedTypeKey: ReservedTypes.RDKIT_MOL}
)
```

Then when we load the data, we can do: 
```py
if array.attributes.get(ReservedTypeKey) == ReservedTypes.RDKIT_MOL:
     ...
```

cwognum (commented):
Hi @j279li , left some comments. Most are pretty minor! Take a look and let me know what you think.

polaris/prediction/_predictions_v2.py
j279li: My mistake, it was a leftover from a prior change/test.

polaris/prediction/_predictions_v2.py
j279li: Yes, that's true. It's there because I wanted to verify the structure of the objects and just make sure I could convert the codecs back properly. I could remove it and repurpose this for the evaluation service, as we will need reads there.

polaris/prediction/_predictions_v2.py
j279li: Great idea! Although since we have some tests that run on python 3.10, I just used `Enum` instead of `StrEnum`.

polaris/prediction/_predictions_v2.py
j279li: That was an oversight, leftover from a previous experimental change.

polaris/prediction/_predictions_v2.py
j279li: I just removed it for the sake of simplicity.

Issue #301, opened 2025-09-17
Add support for Python 3.13


FriedGil:
Python 3.13 has been released for almost a year. Trying to install polaris-lib (with uv and python version 3.13) leads to some cryptic errors, in my case:

`error: 'arrow/pch.h' file not found 23 | #include "arrow/pch.h" | ^~~~~~~~~~~~~ 1 error generated.`

Switching to Python 3.12 solves this issue, but it would be better for polaris to support Python 3.13.

JacksonBurns:
Bump. This would be great to have.

PR #302, opened 2025-09-17
fix: Add python 3.13 support

FriedGil:
Addresses #301 

## Changelogs
- Updated all version specifications to include 3.13
- Bumped pyarrow dependency to >=21.0.0 for 3.13 compatibility
- Verified all tests pass on Python   3.13


cwognum:
Hi @FriedGil , thank you for the contribution! This would be a nice change! 🚀 

It looks like the CICD is currently failing, unrelated to the changes you made. Could you try change the `0.0.0.dev1` to `0.0.0` in the [`pyproject.toml`](https://github.com/polaris-hub/polaris/blob/main/pyproject.toml#L124-L125)?

```
[tool.setuptools_scm]
fallback_version = "0.0.0"
```

FriedGil:
Yes. The tests passed locally but clearly something is awry. I'll do that and take another look.

cwognum:
Hi @FriedGil , I suspect you'll need to remove some of the upper bounds in the `env.yml`, same as you did for the `pyproject.toml`.

FriedGil:
I see. Is there a way for me to replicate the workflow locally so that you don't have to approve it each time?

cwognum:
Good question! It's because you're a first-time contributor. I've relaxed the rules for the time being, such that my approval should no longer be needed as long as you're not a new user on Github. I can't disable it altogether.

FriedGil:
I've looked into it further and fastpdb does not have have a working conda package for python 3.13 (it only has a python wheel).

A PR was made to change this but failed checks and has not been resolved in almost a year: https://github.com/conda-forge/fastpdb-feedstock/pull/3

JacksonBurns:
@cwognum @FriedGil `fastpdb` has since added support for Python 3.13 and 3.14 - could this PR be revived?

